Issue #10, opened 2020-04-09
Setup Gradle build


jgrandja:
We should setup the Gradle build to follow the conventions in [Spring Security](https://github.com/spring-projects/spring-security).

The gradle build should support the following modules:

`core` module with base package -> `org.springframework.security.oauth2.server.authorization`
`config` module with base package -> `org.springframework.security.config`
`samples` module

jbellmann:
Hello, I would like to contribute here, but I'm not sure to start because there are already 2 PRs (#23 and  #24 ) open that contradict somehow to what is required in this ticket, a multi-module setup.
What is recommended, wait until #23 or #24 are merged or creating a PR based on what is currently in master?

At the moment it looks like people want to help/contribute from outside but without clear structure.

@jgrandja 
@eleftherias

jgrandja:
Thanks for reaching out @jbellmann. We're going to hold off on the main project gradle build for now until we actually have framework code to build against. We're starting off with samples #19 and next week I'll have another epic setup for the `authorization_code` grant sample. At that point, we'll have some more issues that can be worked on and more structure to the project plan. It would be great if you can reach out next week after you review the new issues. Sounds good?

PR #30, opened 2020-04-18
Add Resource Server Sample

### LINKED ISSUE #4: Add Resource Server Sample
Related to

It would be good to have a Resource Server sample that integrates with the authorization server sample. It should be quite similar to [the Resource Server sample](https://github.com/spring-projects/spring-security/tree/master/samples/boot/oauth2resourceserver/src) in Spring Security.

Like other samples, it should go into the `samples` directory. For simplicity, classes ought to be placed into a package called `sample`.

Given a valid token from the authorization server, the following test should pass:

```java
this.mockMvc.perform(get("/")
        .header("Authorization", "Bearer " + token))
        .andExpect(status().isOk());
```

ketola:
Add Resource Server Sample
Fixes #4 

- I configured the resource server based on the instructions at https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2resourceserver
- The token validation obviously fails until the auth server is able to provide the token. And as an example, I left the issuer uri to point at https://accounts.google.com.
- I noticed there was a separate issue about configuring the gradle build, that's why I left the module as a standalone.


pivotal-issuemaster:
@ketola Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects-experimental/spring-authorization-server&pullRequestId=30)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects-experimental/spring-authorization-server&pullRequestId=30) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@ketola Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects-experimental/spring-authorization-server&pullRequestId=30)!

samples/resource-server-sample/bin/main/application.yaml
```diff
@@ -0,0 +1,8 @@
+# See https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2resourceserver
```
anzap, line 1: Files in `/bin` should not be committed. PR #31 adds this folder to .gitignore.

samples/resource-server-sample/build.gradle
```diff
@@ -0,0 +1,26 @@
...
+	implementation 'org.springframework.boot:spring-boot-starter-web'
+	implementation 'org.springframework.security:spring-security-oauth2-resource-server:5.3.1.RELEASE'
+	implementation 'org.springframework.security:spring-security-oauth2-jose:5.3.1.RELEASE'
+	implementation 'org.springframework.security:spring-security-config:5.3.1.RELEASE'
+	
+	testImplementation 'org.springframework.boot:spring-boot-starter-test'
+}
+
+test {
+	useJUnitPlatform()
```
anzap, line 25: Should we use JUnit4 or JUnit5 for testing?

ketola: I changed it to use junit 5 api, as that was also used in the parent

ketola:
> I see you added this under `samples` folder. Project does not seem to have gone the multimodule way as there is now a top-level src folder with a `sample` package. Not sure this should be cleaned up now or will be part of issue #10 .

I left the module as standalone on purpose as I saw that there was another issue for creating the gradle build.

samples/resource-server-sample/build.gradle
```diff
@@ -0,0 +1,28 @@
+plugins {
```
jgrandja, line 1: Please modify the build to follow the [helloworld](https://github.com/spring-projects-experimental/spring-authorization-server/blob/master/samples/helloworld/spring-authorization-server-samples-helloworld.gradle) sample, which uses the [spring-build-conventions](https://github.com/spring-gradle-plugins/spring-build-conventions) plugin (same as Spring Security).

The build file should be renamed to `spring-authorization-server-samples-oauth2resourceserver.gradle`

samples/resource-server-sample/src/main/java/sample/SampleResource.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package sample;
+
+import org.springframework.web.bind.annotation.GetMapping;
+import org.springframework.web.bind.annotation.RestController;
+
+@RestController
+public class SampleResource {
```
jgrandja, line 22: Can we rename this to something more meaningful? Maybe `ResourceController` and `resource()`?

samples/resource-server-sample/src/main/resources/application.yaml
```diff
@@ -0,0 +1,8 @@
+# See https://docs.spring.io/spring-security/site/docs/current/reference/html5/#oauth2resourceserver 
+# for more configuration options
+spring:
+  security:
+    oauth2:
+      resourceserver:
+        jwt:
+          issuer-uri:  https://accounts.google.com
```
jgrandja, line 8: Let's use the `jwk-set-uri` property instead, that way the resource server will at least startup.

We could also mock the jwkSet endpoint using `MockWebServer`. Take a look at `OAuth2ResourceServerConfigurerTests` in Spring Security for an example on how to set it up.

samples/resource-server-sample/build.gradle
ketola: Done.

samples/resource-server-sample/src/main/java/sample/SampleResource.java
ketola: Done.

samples/resource-server-sample/src/main/resources/application.yaml
ketola: With this one, I will need some more time as I have not used the `MockWebServer` before, I'll do some research, check the example and try to finish this later today.

ketola:
@jgrandja, I noticed you labeled this as 'duplicate', where has this work been duplicated?

jgrandja:
@ketola We mark PR's as duplicate if there is an associated issue, which there is in this case. However, we typically mark it when we merge. I'll remove the label until we merge to avoid confusion.

samples/resource-server-sample/src/main/resources/application.yaml
ketola: Ok, I checked out the sample from OAuth2ResourceServerConfigurerTests with the MockWebServer, but It started to feel like a lot's work to get the token validation mocked that way. Instead, I decided to mock the JwtDecoder. Let me know what you think.

jgrandja:
Thanks for the updates @ketola. This is now in master.

Issue #14, opened 2020-04-09
Consider adding .editorconfig


jgrandja:

rwinch:
Closing in favor of gh-32

PR #32, opened 2020-04-19
Add an .editorconfig copied from the Spring Security project

VonUniGE:
In order to solve issue #14, here is an .editorconfig copied from the main Spring Security project:
https://github.com/spring-projects/spring-security/blob/master/.editorconfig

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,19 @@
+root = true
+
+[*]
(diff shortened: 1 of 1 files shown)


rwinch:
Thanks for the Pull Request! This is now merged into master :smile:

Issue #43, opened 2020-04-20
Implement Authorization Model / Service


jgrandja:
An authorization server needs to maintain existing authorizations between a client and resource owner. For example, when a resource owner grants access to a client (to access its protected resources), the authorization server must persist certain data in order to validate the authorization grant flow until it completes.

For example, during an `authorization_code` grant flow, the authorization server must persist the following data:

- Certain Authorization Request parameters - `OAuth2AuthorizationRequest`
- The identifier of the client - `client_id`
- The identifier of the resource owner - current `Authentication.getName()`
- The authorization `code` parameter
- The granted access token - `OAuth2AccessToken`

The initial implementation should provide an in-memory implementation of `OAuth2AuthorizationService`, similar to `InMemoryOAuth2AuthorizedClientService` in the `spring-security-oauth2-client` module.

### Implementation Requirements

- `InMemoryOAuth2AuthorizationService` should store in a `Map`
- `OAuth2Authorization` should be immutable
- `OAuth2Authorization.attributes` should be used for storing data that is specific to an authorization grant, eg. authorization `code` parameter, `OAuth2AuthorizationRequest`, etc.
- javadoc `class` and `public` methods
- Unit tests

### Specification References

[4.1.  Authorization Code Grant](https://tools.ietf.org/html/rfc6749#section-4.1)

watsta:
Hi @jgrandja, I would be happy to work on this issue.

jgrandja:
Thank you @watsta. The issue is yours.

> Or do we have to examine the `OAuth2Authorization` object to find out the type of the token?

Yes, this is correct. The `OAuth2AccessToken` is a field in `OAuth2Authorization` and the `authorization_code` should be in `OAuth2Authorization.attributes`.

Issue #67, opened 2020-04-24
Implement Token Endpoint


jgrandja:
An authorization server provides a token endpoint, which the client uses to exchange an authorization grant for an access token.

The `OAuth2TokenEndpointFilter` should be implemented as a `Filter`. The  `DefaultAuthorizationCodeTokenResponseClient` in the `spring-security-oauth2-client` module is the client that calls the Token Endpoint.

### Implementation Requirements

- the `Filter` should process requests for the (default) path `/oauth2/token`
- `authorizationGrantConverter` should convert a valid `authorization_code` Access Token Request to `OAuth2AuthorizationCodeAuthenticationToken` #68
- the `OAuth2AuthorizationCodeAuthenticationToken` should be passed to the `AuthenticationManager`
- the `AuthenticationManager` should be composed of `OAuth2AuthorizationCodeAuthenticationProvider` #68
- `OAuth2AccessTokenAuthenticationToken` #68 should be returned from `AuthenticationManager`
- the `OAuth2AccessToken` should be updated in the _in-flight_ `OAuth2Authorization` #43
- javadoc `class` and `public` methods
- Unit tests

### Specification References

[3.1.  Token Endpoint](https://tools.ietf.org/html/rfc6749#section-3.2)
[4.1.  Authorization Code Grant](https://tools.ietf.org/html/rfc6749#section-4.1)
[4.1.3.  Access Token Request](https://tools.ietf.org/html/rfc6749#section-4.1.3)
[4.1.4.  Access Token Response](https://tools.ietf.org/html/rfc6749#section-4.1.4)

kratostaine:
@jgrandja May I pick this up?

jgrandja:
Thank you @kratostaine. The issue is yours.

jgrandja:
@kratostaine How are things coming along with the PR? Do you have any questions or do you need any help?

kratostaine:
@jgrandja Thank you for enquiring. I don't have any questions right now and am going as per the RFC 6749 spec. Will reach out whenever I have any doubts.

kratostaine:
> `authorizationGrantConverter` should convert a valid `authorization_code` Access Token Request to `OAuth2AuthorizationCodeAuthenticationToken`

@jgrandja Have a query regarding this. How should the `code` be validated as a valid auth code?

kratostaine:
> `OAuth2AccessTokenAuthenticationToken` #68 should be returned from `AuthenticationManager`

[...]

Issue #66, opened 2020-04-24
Implement Authorization Endpoint


jgrandja:
An authorization server provides an authorization endpoint, which it uses when interacting with the resource owner to obtain authorization for a client.

The `OAuth2AuthorizationEndpointFilter` should be implemented as a `Filter`. The  `OAuth2AuthorizationRequestRedirectFilter` in the `spring-security-oauth2-client` module is the client `Filter` that redirects to the Authorization Endpoint.

### Implementation Requirements

- the `Filter` should process requests for the (default) path `/oauth2/authorize`
- `authorizationRequestConverter` should convert a valid Authorization Request to `OAuth2AuthorizationRequest`
- the `RegisteredClientRepository` #40 should be used to validate the `client_id` parameter
- the `OAuth2AuthorizationService` #43 should be used to persist the _in-flight_ `OAuth2Authorization`
- the `codeGenerator` should be used to generate the authorization `code` parameter and it should also be stored in `OAuth2Authorization.attributes` for later validation in #67
- javadoc `class` and `public` methods
- Unit tests

### Specification References

[3.1.  Authorization Endpoint](https://tools.ietf.org/html/rfc6749#section-3.1)
[4.1.  Authorization Code Grant](https://tools.ietf.org/html/rfc6749#section-4.1)
[4.1.1.  Authorization Request](https://tools.ietf.org/html/rfc6749#section-4.1.1)
[4.1.2.  Authorization Response](https://tools.ietf.org/html/rfc6749#section-4.1.2)

paurav-munshi:
@jgrandja I would like to work on this issue. Please let me know if its good to start work on this ?

jgrandja:
Thanks @paurav-munshi. The issue is yours.

paurav-munshi:
@jgrandja 

Can you please clarify few things ? 

1. How will we handle the authentication part of the authorize endpoint ? Is is the scope of this issue ? If so then we will have to redirect the user-agent to a login endpoint

2. Once the code is generation, it will be the responsibility of the filter to redirect the user-agent to the configured redirect-uri along with code in query string. Is this understanding correct ?

jgrandja:
@paurav-munshi 

> How will we handle the authentication part of the authorize endpoint ?

The default security configuration (`WebSecurityConfigurerAdapter`) will be configured with `authenticated()` for the `/oauth2/authorize` endpoint, so the existing features of Spring Security will be leveraged here - in a later story.

[...]

Issue #42, opened 2020-04-20
Add Authorization (User) Consent page


jgrandja:

paurav-munshi:
@jgrandja 

Is this related to the resource owner consent in authorization code grant flow ? If so, why is this on hold. Shouln't it be available for contribution ?

anoopgarlapati:
@paurav-munshi Please check the project roadmap in ZenHub. Authorization consent epic is planned to be available for contribution in the upcoming week.

paurav-munshi:
@anoopgarlapati I think I missed it in the road map when I was looking at it. Thanks for clarifying.

jgrandja:
Thank you for pointing that out @anoopgarlapati.

@paurav-munshi 

> Is this related to the resource owner consent in authorization code grant flow ?

Yes it is. However, it's not needed at the moment to realize the full `authorization_code` grant flow.  My intention is to auto-approve the requested scopes for the initial iteration and then bring in the consent page afterwards.

PR #71, opened 2020-04-26
Add in-memory implementation for OAuth2AuthorizationService

watsta:
Fixes gh-43

<s>Haven't provided unit tests yet,</s> I am unsure about a few things, see comments.

### DIFF
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization;
+
+import org.springframework.util.Assert;
+
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
+
+/**
+ * In-memory implementation of {@link OAuth2AuthorizationService}.
+ *
+ * @author Krisztian Toth
+ */
+public final class InMemoryOAuth2AuthorizationService implements OAuth2AuthorizationService {
+	private final List<OAuth2Authorization> authorizations;
+
+	/**
+	 * Creates an {@link InMemoryOAuth2AuthorizationService}.
+	 */
+	public InMemoryOAuth2AuthorizationService() {
+		this(Collections.emptyList());
+	}
+
+	/**
+	 * Creates an {@link InMemoryOAuth2AuthorizationService} with the provided {@link List}<{@link OAuth2Authorization}>
+	 * as the in-memory store.
+	 *
+	 * @param authorizations a {@link List}<{@link OAuth2Authorization}> object to use as the store
+	 */
+	public InMemoryOAuth2AuthorizationService(List<OAuth2Authorization> authorizations) {
+		Assert.notNull(authorizations, "authorizations cannot be null");
+		this.authorizations = new CopyOnWriteArrayList<>(authorizations);
+	}
+
+	@Override
+	public void save(OAuth2Authorization authorization) {
+		Assert.notNull(authorization, "authorization cannot be null");
+		this.authorizations.add(authorization);
+	}
+
+	@Override
+	public OAuth2Authorization findByTokenAndTokenType(String token, TokenType tokenType) {
+		Assert.hasText(token, "token cannot be empty");
+		Assert.notNull(tokenType, "tokenType cannot be null");
+		return this.authorizations.stream()
+				.filter(authorization -> doesMatch(authorization, token, tokenType))
+				.findFirst()
+				.orElse(null);
+
+	}
+
+	private boolean doesMatch(OAuth2Authorization authorization, String token, TokenType tokenType) {
+		if (tokenType.equals(TokenType.ACCESS_TOKEN)) {
+			return isAccessTokenEqual(token, authorization);
+		} else if (tokenType.equals(TokenType.AUTHORIZATION_CODE)) {
+			return isAuthorizationCodeEqual(token, authorization);
+		}
+		return false;
+	}
+
+	private boolean isAccessTokenEqual(String token, OAuth2Authorization authorization) {
+		return authorization.getAccessToken() != null && token.equals(authorization.getAccessToken().getTokenValue());
+	}
+
+	private boolean isAuthorizationCodeEqual(String token, OAuth2Authorization authorization) {
+		return token.equals(authorization.getAttributes().get(TokenType.AUTHORIZATION_CODE.getValue()));
+	}
+}
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
@@ -16,11 +16,19 @@
 package org.springframework.security.oauth2.server.authorization;
 
 import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.util.Assert;
 
+import java.util.Collections;
+import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
+import java.util.function.Consumer;
 
 /**
+ * Represents a collection of attributes which describe an OAuth 2.0 authorization context.
+ *
  * @author Joe Grandja
+ * @author Krisztian Toth
  */
 public class OAuth2Authorization {
 	private String registeredClientId;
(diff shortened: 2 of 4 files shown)


core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -15,17 +15,127 @@
...
+		return "OAuth2Authorization{" +
+				"registeredClientId='" + registeredClientId + '\'' +
+				", principalName='" + principalName + '\'' +
+				", accessToken=" + accessToken +
+				", attributes=" + attributes +
+				", tokenType=" + tokenType +
+				'}';
+	}
 
+	private TokenType determineTokenType(OAuth2AccessToken accessToken, Map<String, Object> attributes) {
```
watsta, line 131: If I understand correctly, there is either an `OAuth2AccessToken` set, or an `authorization_code` entry should exist in the `attributes` map. In this case I assume it would be best to verify it during instantiation, hence the solution. Should this be done elsewhere, e.g. a dedicated class? Either way, this feels really clunky.
Another solution I can think of would be to create an interface `OAuth2Authorization` which has methods `getToken()` and `getTokenType()` with 2 implementations, `AccessTokenOAuth2Authorization` and `AuthorizationCodeOAuth2Authorization`, or something along these lines. Thoughts?

Captain-P-Goldfish: From my opinion creating the interface is definetely the better solution. Besides the usage of bearer tokens (RFC6750) for example is simply a recommendation of RFC6749. The structure of an access token is not defined at all and it is allowed to use any type of token. I do not know of any case in which someone used another token type as bearer but if someone would need an alternative to the bearer token another approach than the interface would be problematic.

watsta: Thanks for the input. I updated the PR with what I think is a cleaner solution.

core/spring-authorization-server-core.gradle
```diff
@@ -18,3 +19,7 @@ dependencies {
 
 	provided 'javax.servlet:javax.servlet-api'
 }
+
+jacoco {
```
r331, line 23: Is it necessary to include this lib?

anoopgarlapati: Yes. This is the same issue I faced in my PR gh-70. Build fails with JDK 14 if jacoco `0.8.5` is not used.

core/src/main/java/org/springframework/security/oauth2/server/authorization/AuthorizationCodeOAuth2Authorization.java
```diff
@@ -0,0 +1,52 @@
...
+
+import java.util.Map;
+
+/**
+ * An {@link OAuth2Authorization} implementation for authorization code grant. The authorization code must be provided
+ * in the {@code attributes} map.
+ *
+ * @author Krisztian Toth
+ */
+public class AuthorizationCodeOAuth2Authorization extends OAuth2Authorization {
```
jgrandja, line 28: This class is not needed. The intended purpose for `OAuth2Authorization.attributes` is to store **non-common** attributes. The `code` parameter is `authorization_code` grant specific so it would be stored here using a well-known key.

Another example of what would be stored here is the `scope(s)` consented by the user during the authorization consent step #42.

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -15,17 +15,79 @@
...
+ *
  * @author Joe Grandja
+ * @author Krisztian Toth
  */
-public class OAuth2Authorization {
-	private String registeredClientId;
-	private String principalName;
-	private OAuth2AccessToken accessToken;
-	private Map<String, Object> attributes;
+public abstract class OAuth2Authorization implements TokenContainer {
```
jgrandja, line 31: Please remove `abstract`. This domain object does not need to be extended at the moment.

core/src/main/java/org/springframework/security/oauth2/server/authorization/TokenContainer.java
```diff
@@ -0,0 +1,38 @@
...
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization;
+
+/**
+ * An interface for retrieving a token value and its type.
+ *
+ * @author Krisztian Toth
+ */
+public interface TokenContainer {
```
jgrandja, line 23: Can you clarify why this is needed? I'd prefer to remove this.

core/src/main/java/org/springframework/security/oauth2/server/authorization/AccessTokenOAuth2Authorization.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.Collections;
+import java.util.Map;
+import java.util.Objects;
+
+/**
+ * An {@link OAuth2Authorization} implementation with {@link OAuth2AccessToken} access token.
+ *
+ * @author Krisztian Toth
+ */
+public class AccessTokenOAuth2Authorization extends OAuth2Authorization {
```
jgrandja, line 31: Please remove this class as `OAuth2AccessToken` should be a member in `OAuth2Authorization`

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
watsta: Thanks for the review @jgrandja. I wasn't exactly sure what would be the best way to derive the `TokenType` from an `OAuth2Authorization` object. If i am correct there is either a `code` entry in the `attributes` map or an `OAuthAccessToken` field in this class - what do you think would be the best way to get the token and the TokenType attributes in the `InMemoryOAuth2AuthorizationService`? As I can see, we either 
* do that in this class with a `TokenType` field that is assigned to when the object is created:
```java
	public OAuth2Authorization(String registeredClientId, String principalName,
			@Nullable OAuth2AccessToken accessToken, @Nullable Map<String, Object> attributes) {
		// ...
                this.tokenType = // AUTHORIZATION_CODE if accessToken is null and code is set in attributes, 
                                 // ACCESS_TOKEN otherwise, throw exception if none
	}
```
This feels really clunky as this class would have to be modified every time there is a new TokenType (and TokenType also has a public constructor). We would also have to fetch the token from different sources - if I am not mistaken - depending on the token type, so there should also be a field for the token itself.
* or we create an external class with a method signature something along the lines of
```java
public TokenAndType getTokenAndTypeFrom(OAuth2AccessToken accessToken) { ... }
```
But perhaps I am way off, that could also very well be. Could you please advise? Thank you!

jgrandja: @watsta This is what I envision.

Given the current design:

```java
public class OAuth2Authorization {
	private String registeredClientId;
	private String principalName;
	private OAuth2AccessToken accessToken;
	private Map<String, Object> attributes;

	...

}

public interface OAuth2AuthorizationService {

	void save(OAuth2Authorization authorization);

	OAuth2Authorization findByTokenAndTokenType(String token, TokenType tokenType);

}

```

When the Authorization Endpoint #66 responds with the [Authorization Response](https://tools.ietf.org/html/rfc6749#section-4.1.2), the _in-flight_ `OAuth2Authorization` should be `OAuth2AuthorizationService.save()` as such:

```java
public class OAuth2Authorization {
	private String registeredClientId;     // auto-assigned identifier
	private String principalName;     // current `Authentication.getName()`
	private OAuth2AccessToken accessToken;     // null since access token is not available yet
	private Map<String, Object> attributes;     // `code` response parameter assigned to well-known key

	...

}

```

Now when the client calls the Token Endpoint #67,  the [Access Token Request ](https://tools.ietf.org/html/rfc6749#section-4.1.3) will include the `code` parameter issued by #66, which can be looked up using:

```java
OAuth2AuthorizationService.findByTokenAndTokenType(code, TokenType.AUTHORIZATION_CODE)
```

The implementation would use the well-known key (eg. `OAuth2AuthorizationAttributeNames.AUTHORIZATION_CODE`) to lookup in `OAuth2Authorization.attributes` for each `OAuth2Authorization`, and if found matching `code`, return it for further validation by the caller.

Assuming the caller #67 validates successfully, then the `OAuth2Authorization` is updated with the `OAuth2AccessToken` resulting in this state at the end of the grant flow.

```java
public class OAuth2Authorization {
	private String registeredClientId;     // previously assigned
	private String principalName;     // previously assigned
	private OAuth2AccessToken accessToken;     // set the granted access token
	private Map<String, Object> attributes;    // previously assigned

	...

}
```

Does this clarify things?

By the way, have you had a chance to read the [Getting Started](https://github.com/spring-projects-experimental/spring-authorization-server#getting-started)? If not, then I strongly encourage you to read the [OAuth 2.0 Authorization Framework](https://tools.ietf.org/html/rfc6749) to help clarify things further.

watsta: Thank you for taking your time to clarify @jgrandja , it certainly does help. My confusion comes from the fact that for different `TokenType`s we will have to use different logic to save the authorization object. 

In your example, when #66 calls `#save(OAuth2Authorization)` we want to know what the derived `TokenType` of the object is so we can use the lookup method `#findByTokenAndTokenType(token, TokenType)`. This is obviously not a problem, when we want to save the authorization we can just grab the `code` entry from the `attributes` map. When it becomes a problem is when you want to save authorizations with different `TokenType`s than `AUTHORIZATION_CODE`. My initial question was, where should this derivation logic exist? This is what I wanted to solve with the current implementation. 
Should I implement the in-memory service which always saves the authorization object with `AUTHORIZATION_CODE` token type for now?

I think I understand the role of the `OAuth2AuthorizationService` in the application and I am familiar with the OAuth framework to an extent, but I will make sure to read through the specifications and the resources in the Getting Started page, thanks!.

jgrandja: > Should I implement the in-memory service which always saves the authorization object with AUTHORIZATION_CODE token type for now

The `OAuth2AccessToken` (only available in 2nd phase of flow) will also be `save(OAuth2Authorization)` along with the `AUTHORIZATION_CODE` in the 1st phase of the flow. So as far as looking up the `OAuth2AccessToken` using `OAuth2AuthorizationService.findByTokenAndTokenType(code, TokenType.ACCESS_TOKEN)`, we simply search for it via the `OAuth2AccessToken` member.

watsta:
@jgrandja I updated the PR, let me know if its headed in the right direction. Thanks!

core/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.Objects;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * In-memory implementation of {@link OAuth2AuthorizationService}.
+ *
+ * @author Krisztian Toth
+ */
+public class InMemoryOAuth2AuthorizationService implements OAuth2AuthorizationService {
+	private final Map<TokenWithType, OAuth2Authorization> tokenAuthorizationMap = new ConcurrentHashMap<>();
```
jgrandja, line 31: The `TokenWithType` key will limit the search capabilities. What if we need to provide a finder for `principalName`, eg. find all authorizations for a specific user.

Let's change this to a `List` and use `list.stream().filter()` in `findByTokenAndTokenType()`

core/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
```diff
@@ -0,0 +1,88 @@
...
+	public OAuth2Authorization findByTokenAndTokenType(String token, TokenType tokenType) {
+		Assert.hasText(token, "token cannot be empty");
+		Assert.notNull(tokenType, "tokenType cannot be null");
+		return tokenAuthorizationMap.get(new TokenWithType(token, tokenType));
+	}
+
+	/**
+	 * Visible for testing.
+	 */
+	Map<TokenWithType, OAuth2Authorization> getTokenAuthorizationMap() {
```
jgrandja, line 57: I'm not a fan of exposing a method strictly for testing. There usually is a way to test without exposing this method.

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -15,17 +15,108 @@
...
+	public OAuth2Authorization(String registeredClientId, String principalName, OAuth2AccessToken accessToken) {
+		this(registeredClientId, principalName, accessToken, Collections.emptyMap());
+	}
+
+	/**
+	 * Creates a {@link OAuth2Authorization} object without an {@link OAuth2AccessToken}.
+	 *
+	 * @see #OAuth2Authorization(String, String, OAuth2AccessToken, Map)
+	 */
+	public OAuth2Authorization(String registeredClientId, String principalName, Map<String, Object> attributes) {
```
jgrandja, line 53: Instead of overloaded constructors, lets a `Builder`. Take a look at the recent merge of `RegisteredClient` and apply the same `Builder` pattern here.

core/src/test/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationServiceTest.java
```diff
@@ -0,0 +1,138 @@
...
+import java.util.Map;
+
+import static org.junit.Assert.*;
+
+/**
+ * Unit tests for {@link InMemoryOAuth2AuthorizationService}.
+ *
+ * @author Krisztian Toth
+ */
+public class InMemoryOAuth2AuthorizationServiceTest {
```
jgrandja, line 34: Please take a look at `InMemoryRegisteredClientRepositoryTests` and `InMemoryOAuth2AuthorizedClientServiceTests` (`spring-security-oauth2-client` module) to see how our tests are written.

The main points are:
- use AssertJ - not JUnit's assert
- use Mockito for mocking expectations
- the naming conventions for methods name follow a pattern

Please apply the changes here to align with the coding conventions/style that Spring Security uses.

core/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
watsta: I agree, though the implementation requirements stated that `Map` should be used. Updated in 
fd7ab91

core/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
watsta: Using constructor injection in `fd7ab91`, though I am still not sure if this is the best solution. Let me know. Also, which `List` implementation should be used, taking concurrency into account? I went with `CopyOnWriteArrayList` in the default constructor, but as far as I know its performance when writing to the list is really bad.

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
watsta: Added Builder in `9b748c4`

core/src/test/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationServiceTest.java
watsta: Sorry for that, I should have checked. Updated in 48cc835

watsta:
@jgrandja I updated the PR, let me know of any shortcomings.

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -16,16 +16,182 @@
...
+				Objects.equals(attributes, that.attributes);
+	}
+
+	@Override
+	public int hashCode() {
+		return Objects.hash(registeredClientId, principalName, accessToken, attributes);
+	}
+
+	@Override
+	public String toString() {
```
jgrandja, line 78: Please remove `toString()` as there will be sensitive information here and we don't want it displayed

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -16,16 +16,182 @@
...
+		return new Builder();
+	}
+
+	/**
+	 * Returns a new {@link Builder}, initialized with the provided {@link OAuth2Authorization}.
+	 *
+	 * @param authorization the {@link OAuth2Authorization} to copy from
+	 * @return the {@link Builder}
+	 */
+	public static Builder withOAuth2Authorization(OAuth2Authorization authorization) {
```
jgrandja, line 102: Let's rename to `withAuthorization`

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -16,16 +16,182 @@
...
+		Assert.notNull(authorization, "authorization cannot be null");
+		return new Builder(authorization);
+	}
+
+	/**
+	 * Builder class for {@link OAuth2Authorization}.
+	 */
+	public static final class Builder {
```
jgrandja, line 110: Please remove `final`

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -16,16 +16,182 @@
...
+			return this;
+		}
+
+		/**
+		 * Sets the attributes map.
+		 *
+		 * @param attributes the attributes map
+		 * @return the {@link Builder}
+		 */
+		public Builder attributes(Map<String, Object> attributes) {
```
jgrandja, line 165: Please remove this method and add:
- `public Builder attribute(String name, String value)`
- `public Builder attributes(Consumer<Map<String, Object>> attributesConsumer)`

core/src/test/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationServiceTest.java
```diff
@@ -0,0 +1,123 @@
+/*
```
jgrandja, line 1: Test classes should be plural `*Tests`

core/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
```diff
@@ -0,0 +1,84 @@
...
+ * @author Krisztian Toth
+ */
+public class InMemoryOAuth2AuthorizationService implements OAuth2AuthorizationService {
+	private final List<OAuth2Authorization> authorizationStore;
+
+	/**
+	 * Creates an {@link InMemoryOAuth2AuthorizationService}.
+	 */
+	public InMemoryOAuth2AuthorizationService() {
+		this(new CopyOnWriteArrayList<>());
```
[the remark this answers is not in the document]
watsta: Applied, though I think this defeats the purpose of this class, you won't be able to save authorizations into this in-memory store. Is that fine?

watsta:
@jgrandja see my reply, PR is also updated

jgrandja:
Thanks for all the updates @watsta ! FYI, I added a polish commit to get this merged. This is now in master.

PR #72, opened 2020-04-27
Implement Client Credentials Authentication

### LINKED ISSUE #39: Implement Client Authentication
The client must authenticate when calling the authorization server's token endpoint.

The `OAuth2ClientAuthenticationFilter` should be implemented as a `Filter`. The initial implementation should support HTTP Basic only.

- the `Filter` should process requests for the (default) path `/oauth2/token` and if HTTP Basic credentials are available in the request
- the `OAuth2ClientAuthenticationToken` should be passed to the `AuthenticationManager`
- the `AuthenticationManager` should be composed of `OAuth2ClientAuthenticationProvider` (in a later story)
- the `OAuth2ClientAuthenticationProvider` should use the `RegisteredClientRepository` #40 to look up and validate the client credentials
- the `RegisteredClient` should be returned in a new `OAuth2ClientAuthenticationToken` if the authentication succeeds
- the `Filter` should save the `OAuth2ClientAuthenticationToken` in the `SecurityContext`
- Unit tests

### Specification References

[2.3.  Client Authentication](https://tools.ietf.org/html/rfc6749#section-2.3)
[3.1.  Token Endpoint](https://tools.ietf.org/html/rfc6749#section-3.2)
[4.1.3.  Access Token Request](https://tools.ietf.org/html/rfc6749#section-4.1.3)

pkostrzewa:
- [x] The `Filter` should process requests for the (default) path `/oauth2/token` and check if credentials are available in the request.
- [X]  The `OAuth2ClientAuthenticationToken` should be passed to the `AuthenticationManager`.
- [x]  The `AuthenticationManager` should be composed of `OAuth2ClientAuthenticationProvider`.
- [x]  The `OAuth2ClientAuthenticationProvider` should use the `RegisteredClientRepository`.
- [x]  The `RegisteredClient` should be returned in a new `OAuth2ClientAuthenticationToken` if the authentication succeeds.
- [X] The `Filter` should save the  `OAuth2ClientAuthenticationToken` in the `SecurityContext`.
- [x] Add javadoc for `class` and `public` methods.
- [x] Unit tests.

The `ClientAuthenticationConverter` is based on `BasicAuthenticationFilter`.

Fixes #39 .

### DIFF
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationProvider.java
@@ -18,17 +18,52 @@ package org.springframework.security.oauth2.server.authorization.authentication;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
 import org.springframework.security.oauth2.server.authorization.client.RegisteredClientRepository;
+import org.springframework.util.Assert;
 
 /**
+ * An {@link AuthenticationProvider} implementation that validates {@link OAuth2ClientAuthenticationToken}s.
+ *
  * @author Joe Grandja
+ * @author Patryk Kostrzewa
  */
 public class OAuth2ClientAuthenticationProvider implements AuthenticationProvider {
-	private RegisteredClientRepository registeredClientRepository;
+	private final RegisteredClientRepository registeredClientRepository;
+
+	/**
+	 * @param registeredClientRepository
+	 * 		the bean to lookup the client details from
+	 */
+	public OAuth2ClientAuthenticationProvider(RegisteredClientRepository registeredClientRepository) {
+		Assert.notNull(registeredClientRepository, "registeredClientRepository cannot be null");
+		this.registeredClientRepository = registeredClientRepository;
+	}
 
 	@Override
 	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
-		return authentication;
+		String clientId = authentication.getName();
+		if (authentication.getCredentials() == null) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_CLIENT));
+		}
+
+		RegisteredClient registeredClient = this.registeredClientRepository.findByClientId(clientId);
+		// https://tools.ietf.org/html/rfc6749#section-2.4
+		if (registeredClient == null) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_CLIENT));
+		}
+
+		String presentedSecret = authentication.getCredentials()
+				.toString();
+		if (!registeredClient.getClientSecret()
+				.equals(presentedSecret)) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_CLIENT));
+		}
+
+		return new OAuth2ClientAuthenticationToken(registeredClient);
 	}
 
 	@Override
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationToken.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationToken.java
@@ -18,11 +18,11 @@ package org.springframework.security.oauth2.server.authorization.authentication;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.core.SpringSecurityCoreVersion;
 import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
-
 import java.util.Collections;
 
 /**
  * @author Joe Grandja
+ * @author Patryk Kostrzewa
  */
 public class OAuth2ClientAuthenticationToken extends AbstractAuthenticationToken {
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
@@ -44,11 +44,11 @@ public class OAuth2ClientAuthenticationToken extends AbstractAuthenticationToken
 
 	@Override
 	public Object getCredentials() {
-		return null;
+		return this.clientSecret;
 	}
 
 	@Override
 	public Object getPrincipal() {
-		return null;
+		return this.clientId;
 	}
 }
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DefaultOAuth2ClientAuthenticationConverter.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DefaultOAuth2ClientAuthenticationConverter.java
@@ -1,0 +1,72 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.web;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientAuthenticationToken;
+import org.springframework.security.web.authentication.AuthenticationConverter;
+import org.springframework.util.StringUtils;
+import javax.servlet.http.HttpServletRequest;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
+/**
+ * Converts from {@link HttpServletRequest} to {@link OAuth2ClientAuthenticationToken} that can be authenticated.
+ *
+ * @author Patryk Kostrzewa
+ */
+public class DefaultOAuth2ClientAuthenticationConverter implements AuthenticationConverter {
+
+	private static final String AUTHENTICATION_SCHEME_BASIC = "Basic";
+
+	@Override
+	public OAuth2ClientAuthenticationToken convert(HttpServletRequest request) {
+		String header = request.getHeader(HttpHeaders.AUTHORIZATION);
+
+		if (header == null) {
+			return null;
+		}
+
+		header = header.trim();
+		if (!StringUtils.startsWithIgnoreCase(header, AUTHENTICATION_SCHEME_BASIC)) {
+			return null;
+		}
+
+		if (header.equalsIgnoreCase(AUTHENTICATION_SCHEME_BASIC)) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST));
+		}
+
+		byte[] decoded;
+		try {
+			byte[] base64Token = header.substring(6)
+					.getBytes(StandardCharsets.UTF_8);
+			decoded = Base64.getDecoder()
+					.decode(base64Token);
+		} catch (IllegalArgumentException e) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST));
+		}
+
+		String token = new String(decoded, StandardCharsets.UTF_8);
+		String[] credentials = token.split(":");
+		if (credentials.length != 2) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_TOKEN));
+		}
+		return new OAuth2ClientAuthenticationToken(credentials[0], credentials[1]);
+	}
+}
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
@@ -15,9 +15,21 @@
  */
 package org.springframework.security.oauth2.server.authorization.web;
 
+import org.springframework.http.MediaType;
+import org.springframework.http.server.ServletServerHttpResponse;
 import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.http.converter.OAuth2ErrorHttpMessageConverter;
+import org.springframework.security.web.authentication.AuthenticationConverter;
+import org.springframework.security.web.authentication.AuthenticationFailureHandler;
+import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
+import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
 import org.springframework.web.filter.OncePerRequestFilter;
-
 import javax.servlet.FilterChain;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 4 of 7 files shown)


pkostrzewa:
@jgrandja Should `AuthenticationSuccessHandler` and `AuthenticationFailureHandler` be used in early implementation?

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationProvider.java
```diff
@@ -15,20 +15,64 @@
...
+		String clientId = authentication.getName();
+		if (authentication.getCredentials() == null) {
+			logger.debug("Authentication failed: no credentials provided");
+			throw new BadCredentialsException("Bad credentials");
+		}
+
+		RegisteredClient registeredClient = registeredClientRepository.findByClientId(clientId);
+		// https://tools.ietf.org/html/rfc6749#section-2.4
+		if (registeredClient == null) {
+			logger.debug("Authentication failed: client not registered");
```
watsta, line 62: wouldn't it be helpful to include the clientId in the debug message?

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationProvider.java
```diff
@@ -15,20 +15,64 @@
...
+			throw new BadCredentialsException("Bad credentials");
+		}
+
+		String presentedSecret = authentication.getCredentials()
+				.toString();
+		if (!registeredClient.getClientSecret()
```
watsta, line 68: I think clients are not required to have a secret. Obviously, such clients should not be used with the client credentials flow, but that could be something to mitigate here. I think a simple null-check would make sense

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientAuthenticationConverter.java
```diff
@@ -0,0 +1,70 @@
...
+					.getBytes(StandardCharsets.UTF_8);
+			decoded = Base64.getDecoder()
+					.decode(base64Token);
+		} catch (IllegalArgumentException e) {
+			throw new BadCredentialsException("Failed to decode basic authentication token");
+		}
+
+		String token = new String(decoded, StandardCharsets.UTF_8);
+		String[] credentials = token.split(":");
+		if (credentials.length == 0) {
```
watsta, line 65: shouldn't this rather be `credentials.length != 2`? Otherwise you can still get an `ArrayIndexOutOfBoundsException` on line 68, e.g when there is no `:` present in the Authorization header, length will be 1

nifrez:
@watsta It's draft at the moment ;)

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
 	}
 
+	protected void onSuccessfulAuthentication(HttpServletRequest request, HttpServletResponse response,
+			Authentication result) throws IOException {
+
+		SecurityContextHolder.getContext()
+				.setAuthentication(result);
+	}
+
+	protected void onUnsuccessfulAuthentication(HttpServletRequest request, HttpServletResponse response,
```
Captain-P-Goldfish, line 113: renaming proposal: "onFailedAuthentication"
the current state made me misread this 2 times.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationProvider.java
```diff
@@ -15,20 +15,64 @@
...
 /**
+ * An {@link AuthenticationProvider} implementation that validates {@link OAuth2ClientAuthenticationToken}s.
+ *
  * @author Joe Grandja
+ * @author Patryk Kostrzewa
  */
 public class OAuth2ClientAuthenticationProvider implements AuthenticationProvider {
 	private RegisteredClientRepository registeredClientRepository;
 
+	protected final Log logger = LogFactory.getLog(getClass());
```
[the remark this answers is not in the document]
pkostrzewa: I'd rather stick to that is already configured and present rather than adding slf4j in this PR.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
pkostrzewa: Agree.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientAuthenticationConverter.java
pkostrzewa: That's true.

pkostrzewa:
@jgrandja Could you please elaborate on "The `AuthenticationManager` should be composed of `OAuth2ClientAuthenticationProvider`?

Isn't `OAuth2ClientAuthenticationProvider` automatically registered in `AuthenticationManager`? If not, what should be done to register a provider?

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationToken.java
```diff
@@ -23,6 +23,7 @@
 
 /**
  * @author Joe Grandja
+ * @author Patryk Kostrzewa
  */
 public class OAuth2ClientAuthenticationToken extends AbstractAuthenticationToken {
```
jgrandja, line 28: For constructor arguments, apply `Assert.notNull()` or `Assert.hasText()` or whatever makes sense.

For private members, mark `final` whenever applicable (set in constructor and no corresponding setter)

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
 import org.springframework.web.filter.OncePerRequestFilter;
-
 import javax.servlet.FilterChain;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 
 /**
+ * Processes a HTTP request's BASIC authorization headers, putting the result into the
```
jgrandja, line 32: Please remove this javadoc that came from `BasicAuthenticationFilter` as not all details apply to this filter. For example, this `Filter` will support multiple authentication schemes (not just basic) via a composite `AuthenticationConverter`.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
+ * <p>
+ * If authentication fails, the <code>SecurityContextHolder</code> will be cleared.
+ *
  * @author Joe Grandja
+ * @author Patryk Kostrzewa
  */
 public class OAuth2ClientAuthenticationFilter extends OncePerRequestFilter {
-	private AuthenticationManager authenticationManager;
+
+	private static final String TOKEN_ENDPOINT = "/oauth2/token";
```
jgrandja, line 63: The default token endpoint is "/oauth2/token" but it should also be configurable via a constructor. Take a look at `OAuth2AuthorizationRequestRedirectFilter` to see how a custom path can be supplied.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
+	private final AuthenticationManager authenticationManager;
+	private final ClientAuthenticationConverter authenticationConverter = new ClientAuthenticationConverter();
```
jgrandja, line 65: Let's reference the interface `private AuthenticationConverter authenticationConverter` and provide a setter to enable customization `public final void setAuthenticationConverter(AuthenticationConverter authenticationConverter)`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
+	 */
+	public OAuth2ClientAuthenticationFilter(AuthenticationManager authenticationManager) {
+		Assert.notNull(authenticationManager, "authenticationManager cannot be null");
+		this.authenticationManager = authenticationManager;
+	}
 
 	@Override
-	protected void doFilterInternal(HttpServletRequest request,
-			HttpServletResponse response, FilterChain filterChain)
+	public void afterPropertiesSet() {
```
jgrandja, line 81: Please remove as `AuthenticationManager` is required as per constructor.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
+		Assert.notNull(authenticationManager, "An AuthenticationManager is required");
+	}
+
+	@Override
+	protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
 			throws ServletException, IOException {
 
+		if (TOKEN_ENDPOINT.equals(request.getRequestURI())) {
```
jgrandja, line 89: Let's use `RequestMatcher` specifically `AntPathRequestMatcher`. See `OAuth2AuthorizationRequestRedirectFilter` -> `DefaultOAuth2AuthorizationRequestResolver` -> `RequestMatcher`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
+			if (authentication == null) {
+				filterChain.doFilter(request, response);
+				return;
+			}
+			try {
+				final Authentication result = authenticationManager.authenticate(authentication);
+				onSuccessfulAuthentication(request, response, result);
+			} catch (AuthenticationException failed) {
+				onUnsuccessfulAuthentication(request, response, failed);
+				filterChain.doFilter(request, response);
```
jgrandja, line 100: Failed authentication should not proceed through the chain. The failure handler should respond with error response.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
+			}
+		}
+		filterChain.doFilter(request, response);
 	}
 
+	protected void onSuccessfulAuthentication(HttpServletRequest request, HttpServletResponse response,
```
jgrandja, line 106: Replace with `AuthenticationSuccessHandler` delegate and add `final` setter to allow for customization.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
jgrandja: Replace with `AuthenticationFailureHandler` delegate and add `final` setter to allow for customization.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -16,25 +16,103 @@
...
+			Authentication result) throws IOException {
+
+		SecurityContextHolder.getContext()
+				.setAuthentication(result);
+	}
+
+	protected void onUnsuccessfulAuthentication(HttpServletRequest request, HttpServletResponse response,
+			AuthenticationException failed) throws IOException {
+
+		SecurityContextHolder.clearContext();
```
jgrandja, line 116: We also need to write the OAuth 2.0 Error Response based on the incoming `OAuth2AuthenticationException`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientAuthenticationConverter.java
```diff
@@ -0,0 +1,70 @@
...
+import javax.servlet.http.HttpServletRequest;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+
+/**
+ * Converts from {@link HttpServletRequest} to {@link OAuth2ClientAuthenticationToken} that can be authenticated.
+ *
+ * @author Patryk Kostrzewa
+ */
+public class ClientAuthenticationConverter implements AuthenticationConverter {
```
jgrandja, line 32: Rename to `OAuth2ClientAuthenticationConverter`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientAuthenticationConverter.java
```diff
@@ -0,0 +1,70 @@
...
+			return null;
+		}
+
+		header = header.trim();
+		if (!StringUtils.startsWithIgnoreCase(header, AUTHENTICATION_SCHEME_BASIC)) {
+			return null;
+		}
+
+		if (header.equalsIgnoreCase(AUTHENTICATION_SCHEME_BASIC)) {
+			throw new BadCredentialsException("Empty basic authentication token");
```
jgrandja, line 50: We should throw `OAuth2AuthenticationException` using `invalid_client` as the `OAuth2Error.errorCode`. See [5.2.  Error Response](https://tools.ietf.org/html/rfc6749#section-5.2)

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientAuthenticationConverter.java
```diff
@@ -0,0 +1,70 @@
...
+		}
+
+		byte[] decoded;
+		try {
+			byte[] base64Token = header.substring(6)
+					.getBytes(StandardCharsets.UTF_8);
+			decoded = Base64.getDecoder()
+					.decode(base64Token);
+		} catch (IllegalArgumentException e) {
+			throw new BadCredentialsException("Failed to decode basic authentication token");
```
jgrandja, line 60: throw `OAuth2AuthenticationException `

jgrandja:
@pkostrzewa 

> Isn't `OAuth2ClientAuthenticationProvider` automatically registered in `AuthenticationManager`?

Yes, it will be in a later story. Nothing to do as far as this issue goes. I was simply stating the runtime expectations. I updated the main issue to be clear.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientAuthenticationConverter.java
pkostrzewa: Shouldn't the error be `invalid_request` in this case?

pkostrzewa:
@jgrandja 

Thanks for the review! I will update the PR with your comments later on today.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
pkostrzewa: I've changed this to simply return 401 status for now in default `AuthenticationFailureHandler`. Should I write an `OAuth2Error` to `response` instead?

pkostrzewa:
@jgrandja I've updated the PR. There are still missing tests and documentation and I will provide both as soon as I can.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/DefaultOAuth2ClientAuthenticationFailureHandler.java
```diff
@@ -0,0 +1,39 @@
...
+ * @author Patryk Kostrzewa
+ */
+public class DefaultOAuth2ClientAuthenticationFailureHandler implements AuthenticationFailureHandler {
+
+	@Override
+	public void onAuthenticationFailure(final HttpServletRequest request, final HttpServletResponse response,
+			final AuthenticationException exception) throws IOException, ServletException {
+
+		SecurityContextHolder.clearContext();
+		response.sendError(HttpStatus.UNAUTHORIZED.value());
```
jgrandja, line 37: We should also write out the `OAuth2Error` in the response. Try to reuse `OAuth2ErrorHttpMessageConverter` from `spring-security-oauth2-client` module.

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationConverterTests.java
```diff
@@ -0,0 +1,116 @@
...
+
+	private OAuth2ClientAuthenticationConverter converter;
+
+	@Before
+	public void setup() {
+		converter = new OAuth2ClientAuthenticationConverter();
+	}
+
+	@Test
+	public void shouldConvertToAuthenticationToken() {
```
jgrandja, line 41: Please see existing tests in `spring-security-oauth2-client` module and follow the method naming pattern and general test style. eg. `DefaultOAuth2AuthorizationRequestResolverTests`, `OAuth2AuthorizationCodeAuthenticationProviderTests`, etc.

pkostrzewa:
Thanks for the review @jgrandja. I will update the PR soon.

pkostrzewa:
> For the next update, can you please rebase on `master` and avoid merge commits. And after you add all the tests please change the Draft to Ready and I think we'll be close to merging.

Do you mean I should squash all commits from this PR into single one?

jgrandja:
@pkostrzewa Yes please squash to 1 commit and rebase on latest master.

pkostrzewa:
I've added missing tests and rebased on top of the `master` branch.

If there will be any changes requested I can address them next wednesday at earliest.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DefaultOAuth2ClientAuthenticationConverter.java
```diff
@@ -0,0 +1,72 @@
...
+		} catch (IllegalArgumentException e) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST));
+		}
+
+		String token = new String(decoded, StandardCharsets.UTF_8);
+		String[] credentials = token.split(":");
+		if (credentials.length != 2) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_TOKEN));
+		}
+		return new OAuth2ClientAuthenticationToken(credentials[0], credentials[1]);
```
anoopgarlapati, line 70: @pkostrzewa cc: @jgrandja 
As per [RFC 6749 - section 2.3.1](https://tools.ietf.org/html/rfc6749#section-2.3.1), the client identifier and client password are further encoded using `application/x-www-form-urlencoded` encoding algorithm and then used as username and password .
> The client identifier is encoded using the
   "application/x-www-form-urlencoded" encoding algorithm per
   Appendix B, and the encoded value is used as the username; the client
   password is encoded using the same algorithm and used as the
   password. 

There is an issue already logged in legacy oauth2 project spring-projects/spring-security-oauth#1826 and we should implement this according to specification in this project.

dfcoffin: @anoopgarlapati
Although [RFC 6749 - section 2.3.1,](https://tools.ietf.org/html/rfc6749#section-2.3.1) indeed does say:

> The client identifier is encoded using the
> "application/x-www-form-urlencoded" encoding algorithm per
> Appendix B, and the encoded value is used as the username; the client
> password is encoded using the same algorithm and used as the
> password.

It further states:

> Including the client credentials in the request-body using the two parameters is NOT RECOMMENDED and SHOULD be limited to clients unable to directly utilize the HTTP Basic authentication scheme (or other password-based HTTP authentication schemes).  the parameters can only be transmitted in the request-body and MUST NOT be included in the request URI.
> 
> The authorization server MUST require the of TLS as described in [Section 1.6](https://tools.ietf.org/html/rfc6749#section-1.6) when sending requests using password authentication.
> 
> Since this client authentication method involves a password, the authorization server MUST protect any endpoint utilizing it against brute force attacks.

jgrandja:
Thank you for all the updates @pkostrzewa !

Overall, the implementation looked good, however, I did add a polish commit in order to get this merged.

Thank you again and I look forward to another contribution 👍

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DefaultOAuth2ClientAuthenticationConverter.java
OrangeDog: Those sections do not contradict each other. The encoding needs to be applied when using Basic authentication.

PR #77, opened 2020-05-04
Authorization Endpoint filter for Authorization Code flow

### LINKED ISSUE #66: Implement Authorization Endpoint

paurav-munshi:
Fixes #66 

Functionality Covered
-------------------------
- Validation of client id & response type
- Generate authorization code
- Save Authorization
- Redirect to client redirect uri
- Send error in http response
- Send error as query params in redirect uri

Validations covered
-------------------
- Client Id should be mandatory
- Client Id should be registered
- Client Id should be configured with Authorization grant type
- Response type is mandaotry
- Response type should be code
- Redirect Uri should be specified if client has multiple redirect uris configured
- If supplied, redirect uri should be same as the one configured on client
- User should be authenticated
- Endpoint should be /oauth2/authorize


core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+		this.authorizationRedirectStrategy.sendRedirect(request, response, finalRedirectURI);
+	}
+
+	public Converter<HttpServletRequest, OAuth2AuthorizationRequest> getAuthorizationRequestConverter() {
+		return authorizationRequestConverter;
+	}
+
+	@Override
+	protected boolean shouldNotFilter(HttpServletRequest request) throws ServletException {
+		return !authorizationEndpiontMatcher.matches(request);
```
kratostaine, line 226: Minor typo in `authorizationEndpiontMatcher`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+		this.authorizationRedirectStrategy.sendRedirect(request, response, finalRedirectUri);
+	}
+
+	private void sendErrorInResponse(HttpServletResponse response, OAuth2Error authorizationError) throws IOException {
+		int errorStatus = -1;
+		String errorCode = authorizationError.getErrorCode();
+		if (errorCode.equals(OAuth2ErrorCodes.ACCESS_DENIED))
+			errorStatus=HttpStatus.FORBIDDEN.value();
+		else errorStatus=HttpStatus.INTERNAL_SERVER_ERROR.value();
+		response.sendError(errorStatus, authorizationError.getErrorCode()+":"+authorizationError.getDescription());
```
pgerhard, line 203: I find not using curly braces in "if" statements to be very bug prone. I'd suggest introducing them here and in other other locations in this PR. I've noticed this in numerous locations within this PR

pgerhard: @jgrandja do you have any suggestions on which coding guidelines we should follow? I found [this document](https://github.com/spring-projects/spring-framework/wiki/Code-Style) in the spring-framework repository. But it does say its still a work-in-progress

jgrandja: @pgerhard Take a look at the [Getting Started](https://github.com/spring-projects-experimental/spring-authorization-server#getting-started), specifically:

> Submitted work via pull requests should follow the same coding style/conventions and adopt the same or similar design patterns that have been established in Spring Security’s OAuth 2.0 support

Essentially, we are looking for the same coding style/patterns that are implemented in the OAuth 2.0 codebase in Spring Security 5.x. Consistency is key, so the Authorization Server code should align as well.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
 
 	@Override
 	protected void doFilterInternal(HttpServletRequest request,
 			HttpServletResponse response, FilterChain filterChain)
 			throws ServletException, IOException {
 
+		RegisteredClient client = null;
+		OAuth2AuthorizationRequest authorizationRequest = null;
+		OAuth2Authorization authorization = null;
+
```
jgrandja, line 96: I would have expected to see the use of `authorizationEndpointMatcher` as the very first call here. However, it's not being used at all. Please review [4.1.1.  Authorization Request](https://tools.ietf.org/html/rfc6749#section-4.1.1). At a minimum, there should be a `RequestMatcher` that matches on a **valid Authorization Request**. A valid authorization request should at least match on the configured endpoint path (using `AntPathRequestMatcher`) and the required parameters `response_type=code`, `client_id=*`. This is a common pattern used throughout Spring Security, where `Filter` use a `RequestMatcher` to filter out request or process them.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+		try {
+			checkUserAuthenticated();
```
jgrandja, line 98: After the `RequestMatcher` determines to process the request, it should delegate to `authorizationRequestConverter` to convert to `OAuth2AuthorizationRequest`. Then the `OAuth2AuthorizationRequest` should be used to perform the required validations as the next step.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+			client = fetchRegisteredClient(request);
+
+			authorizationRequest = authorizationRequestConverter.convert(request);
+			validateAuthorizationRequest(request, client);
```
jgrandja, line 102: Why not pass `authorizationRequest` here instead of `request`?

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+		return client;
+
+	}
+
+	protected OAuth2Authorization buildOAuth2Authorization(RegisteredClient client,
+			OAuth2AuthorizationRequest authorizationRequest, String code) {
+		OAuth2Authorization authorization = OAuth2Authorization.createBuilder()
+					.clientId(authorizationRequest.getClientId())
+					.addAttribute(OAuth2ParameterNames.CODE, code)
+					.addAttribute(OAuth2Authorization.ISSUED_AT, Instant.now())
```
jgrandja, line 157: Please remove `ISSUED_AT`, `CODE_USED` and `SCOPE`. These will be added in a later story. We only need to store `CODE` and `OAuth2AuthorizationRequest` `attributes` as part of this story.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+		boolean isAuthoirzationGrantAllowed = Stream.of(client.getAuthorizationGrantTypes())
+				.anyMatch(grantType -> grantType.contains(AuthorizationGrantType.AUTHORIZATION_CODE));
+			if (!isAuthoirzationGrantAllowed)
+				throw new OAuth2AuthorizationException(AUTHZ_CODE_NOT_SUPPORTED_ERROR);
+
+		return client;
+
+	}
+
+	protected OAuth2Authorization buildOAuth2Authorization(RegisteredClient client,
```
jgrandja, line 152: We typically encapsulate in the initial design and only expose when needed. Please change to `private` for all exposed methods.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+	protected boolean shouldNotFilter(HttpServletRequest request) throws ServletException {
+		return !authorizationEndpiontMatcher.matches(request);
+	}
+
+	public void setAuthorizationRequestConverter(
+			Converter<HttpServletRequest, OAuth2AuthorizationRequest> authorizationRequestConverter) {
+		this.authorizationRequestConverter = authorizationRequestConverter;
+	}
+
+	public RegisteredClientRepository getRegisteredClientRepository() {
```
jgrandja, line 234: Please remove getters. The convention is to provide `final` setters for **optional** attributes and **required** attributes are provided via constructors. See `OAuth2AuthorizationRequestRedirectFilter` and `DefaultOAuth2AuthorizedClientManager` for example

core/src/main/java/org/springframework/security/oauth2/server/authorization/util/AuthorizationCodeKeyGenerator.java
```diff
@@ -0,0 +1,34 @@
...
+
+import java.util.UUID;
+
+import org.springframework.security.crypto.keygen.StringKeyGenerator;
+
+/**
+ * @author Paurav Munshi
+ * @since 0.0.1
+ */
+public class AuthorizationCodeKeyGenerator implements StringKeyGenerator {
```
jgrandja, line 26: Please remove this and instead leverage the same that is used in `DefaultOAuth2AuthorizationRequestResolver.stateGenerator`

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilterTest.java
```diff
@@ -0,0 +1,380 @@
...
+
+		filter.setAuthorizationService(authorizationService);
+		filter.setCodeGenerator(codeGenerator);
+		filter.setRegisteredClientRepository(registeredClientRepository);
+
+		SecurityContextHolder.getContext().setAuthentication(authentication);
+	}
+
+	@Test
+	public void testFilterRedirectsWithCodeOnValidReq() throws Exception {
```
jgrandja, line 87: There is a naming convention for test methods. For an example, please see `DefaultOAuth2AuthorizationRequestResolverTests` and `OAuth2AuthorizationCodeGrantFilterTests`

core/src/main/java/org/springframework/security/oauth2/server/authorization/util/AuthorizationCodeKeyGenerator.java
paurav-munshi: That state generator uses Base64StringKeyGenerator internally. Earlier I looked at that option but because Base64StringKeyGenerator uses SecureRandom internally I thought its better to use something which explicitly guarantees of uniqueness. Therefore I used a generator based on UUID. But it seems UUID also uses SecureRandom internally for type 4 UUID. So I would change my code to use Base64StringKeyGenerator.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: I have used the RequestMatcher in the shouldNotFilter method. That method is called by OncePerRequestFilter.doFilter method effectively calling it before doFilterInternal. I thought it is much more natural to filter this way for checking endpoint. And I also think that validity of the query parameters is not supposed to be done in RequestMatcher because it is only responsible to check if filter should be invoked for that request or not. Once invoked, I do not think RequestMatcher should validate the request and therefore I have those validations in separate methods. Let me know if it is not to be done in that way.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: Actually, when we build `authorizationRequest`, it automatically populates authorization code when creating the Builder using the `OAuth2AuthorizationRequest.authorizationCode` method. So even if we get blank `response_type `from `request`, `authorizationRequest `would still have response type as code. So instead of doing one validation on `request `and other validations on `authorizationRequest`, it felt more obvious to do all validations on `request`.

I had to separate out client id validation to be done in `fetchRegisteredClient` because other validation like `redirect_uri ` validations or `scope ` validations(future story) are dependent on registered client configuration. Otherwise I earlier planned to also perform client id validation in `validateAuthorizationRequest ` method. Infact I was just thinking, I can move 
```
if (client==null) {
    throw new OAuth2AuthorizationException(new OAuth2Error(OAuth2ErrorCodes.ACCESS_DENIED));
}

boolean isAuthorizationGrantAllowed = Stream.of(client.getAuthorizationGrantTypes())
	.anyMatch(grantType -> grantType.contains(AuthorizationGrantType.AUTHORIZATION_CODE));
if (!isAuthorizationGrantAllowed) {
	throw new OAuth2AuthorizationException(new OAuth2Error(OAuth2ErrorCodes.ACCESS_DENIED));
}
```
from fetchRegisteredClient to validateAuthorizationRequest method so that there is only one universal validation method.

Hope if makes sense. Please suggest if you think there is a better approach.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: Please see my [comment ](https://github.com/spring-projects-experimental/spring-authorization-server/pull/77#discussion_r422573402) below. There I have tried to explain why I have used request instead of OAuth2AuthorizationRequest.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
jgrandja: > I have used the RequestMatcher in the shouldNotFilter method

Sorry  missed that.

> I also think that validity of the query parameters is not supposed to be done in RequestMatcher

A valid Authorization Request requires `response_type=code` and `client_id != empty` so both of these need to be applied in `RequestMatcher`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
jgrandja: > So even if we get blank response_type from request

As mentioned in other comment, a valid Authorization Request requires `response_type=code` so this should be applied in `RequestMatcher`. So when `OAuth2AuthorizationRequest.authorizationCode()` is called, we know that `response_type=code` or else `authorizationRequestConverter` would never be called.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: > A valid Authorization Request requires response_type=code and client_id != empty so both of these need to be applied in RequestMatcher

So here is where my confusion lies. `RequestMatcher `by its nature, it seems, is supposed to decide whether request is supported by that filter or not. So if the `RequestMatcher `does not find a match it will silently allow the request to proceed to `FilterChain`. But if we look at first line of [4.1.2.1 Error Response](https://tools.ietf.org/html/rfc6749#section-4.1.2.1), it says :- 

> If the request fails due to a missing, invalid, or mismatching
>    redirection URI, or if the client identifier is missing or invalid,
>    the authorization server SHOULD inform the resource owner of the
>    error

So if we are putting` response_type=code and client_id != empty` in `RequestMatcher` it will not send the error response back to resource owner's user-agent. So how are supposed to handle this scenario if we want to perform this validation in `RequestMatcher`. I think that in future if Implicit grant is going to be implemented using a different filter then we can add `response_type=code` in `RequestMatcher` but I would still not think of validating client id in `ReqeustMatcher`. 

Can you please provide some more light here ?

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: Fixed. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+	protected RegisteredClient fetchRegisteredClient(HttpServletRequest request) throws OAuth2AuthorizationException {
+		String clientId = request.getParameter(OAuth2ParameterNames.CLIENT_ID);
+		if (StringUtils.isEmpty(clientId))
+			throw new OAuth2AuthorizationException(CLIENT_ID_ABSENT_ERROR);
+
+		RegisteredClient client = registeredClientRepository.findByClientId(clientId);
+		if (client==null)
+			throw new OAuth2AuthorizationException(CLIENT_ID_NOT_FOUND_ERROR);
+
+		boolean isAuthoirzationGrantAllowed = Stream.of(client.getAuthorizationGrantTypes())
```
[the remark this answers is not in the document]
paurav-munshi: Fixed. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
 	private Converter<HttpServletRequest, OAuth2AuthorizationRequest> authorizationRequestConverter;
 	private RegisteredClientRepository registeredClientRepository;
 	private OAuth2AuthorizationService authorizationService;
 	private StringKeyGenerator codeGenerator;
+	private RedirectStrategy authorizationRedirectStrategy;
+	private RequestMatcher authorizationEndpiontMatcher;
+
+	private static final String DEFAULT_ENDPOINT = "/oauth2/authorize";
+
+	private static final OAuth2Error CLIENT_ID_ABSENT_ERROR = new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST, OAuth2AuthorizationServerMessages.REQUEST_MISSING_CLIENT_ID, null);
```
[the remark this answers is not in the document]
paurav-munshi: Fixed in PR. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,252 @@
...
+	private static final OAuth2Error INVALID_REDIRECT_URI_REQUESTED = new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST, OAuth2AuthorizationServerMessages.REQUESTED_REDIRECT_URI_INVALID, null);
+	private static final OAuth2Error CLIENT_ID_NOT_FOUND_ERROR = new OAuth2Error(OAuth2ErrorCodes.ACCESS_DENIED, OAuth2AuthorizationServerMessages.CLIENT_ID_NOT_FOUND, null);
+	private static final OAuth2Error USER_NOT_AUTHENTICATED_ERROR = new OAuth2Error(OAuth2ErrorCodes.ACCESS_DENIED, OAuth2AuthorizationServerMessages.USER_NOT_AUTHENTICATED, null);
+	private static final OAuth2Error AUTHZ_CODE_NOT_SUPPORTED_ERROR = new OAuth2Error(OAuth2ErrorCodes.ACCESS_DENIED, OAuth2AuthorizationServerMessages.CLIENT_ID_UNAUTHORIZED_FOR_CODE, null);
+	private static final OAuth2Error RESPONSE_TYPE_NOT_FOUND_ERROR = new OAuth2Error(OAuth2ErrorCodes.UNSUPPORTED_RESPONSE_TYPE, OAuth2AuthorizationServerMessages.RESPONSE_TYPE_MISSING_OR_INVALID, null);
+
+
+
+	public OAuth2AuthorizationEndpointFilter() {
+		authorizationEndpiontMatcher = new AntPathRequestMatcher(DEFAULT_ENDPOINT);
```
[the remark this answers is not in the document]
paurav-munshi: Fixed in PR. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: Fixed in PR. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: Fixed in PR. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: Fixed in PR. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/util/AuthorizationCodeKeyGenerator.java
paurav-munshi: Fixed in PR. Thanks.

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilterTest.java
paurav-munshi: Fixed in PR. Thanks.

paurav-munshi:
> Thanks for the updates @paurav-munshi. Please see my review comments.
> 
> Also, I noticed quite a few issues with code formatting and style, eg. missing curly braces in `if` blocks, missing space between `if` and `(`, etc. Again, please review [Getting Started](https://github.com/spring-projects-experimental/spring-authorization-server#getting-started), as it's very imprortant that the code style/conventions are adopted from what exists today in Spring Security 5 OAuth 2.0 support classes.

I have not found a definitive guide for code formatting but I mainly followed [this document](https://github.com/spring-projects/spring-security/blob/master/CONTRIBUTING.adoc) and [this document](https://github.com/spring-projects/spring-framework/wiki/Code-Style). And I looked on other existing classes from Spring Security OAuth Core for validating my understanding of code formatting. I have applied following changes which are now incorporated in PR :- 

- Line width of 120 characters with few exceptions
- Method sequence to be constructors, setters, over-ridden methods and then private methods
- Add braces to if conditions with single statement
- Moved else and catch to the next line of closing brace of preview try / if block
- this keyword for accessing instance members
- Constructors for mandatory members not initiated by default
- final setters for optional or default initiated members
- Assert non null for constructor and setter arguments
- Renamed test methods to methodName-When-Then format

Do let me know if you find more discrepancies. Thanks.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: At a minimum I have added the check for response_type in RequestMatcher. So if the response_type is blank or if its not code then this filter will not be invoked and filter chain will proceeded with. Tests have been added to check the same. 

Can you please review the PR ?

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: Modified the validation method to validate on `OAuth2AuthorizationRequest `instead of `request`. Please review. Thanks.

kratostaine:
@paurav-munshi I don't think you should follow this particular code formatting:
> Moved else and catch to the next line of closing brace of preview try / if block

Reference files: [OAuth2AuthorizationRequestRedirectFilter.java](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilter.java) and [HttpSessionOAuth2AuthorizationRequestRepository.java](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java)
The else/else if block and catch block should be on the same line of the closing brace of previous try/if block.

paurav-munshi:
> @paurav-munshi I don't think you should follow this particular code formatting:
> 
> > Moved else and catch to the next line of closing brace of preview try / if block
> 
> Reference files: [OAuth2AuthorizationRequestRedirectFilter.java](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/OAuth2AuthorizationRequestRedirectFilter.java) and [HttpSessionOAuth2AuthorizationRequestRepository.java](https://github.com/spring-projects/spring-security/blob/master/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/web/HttpSessionOAuth2AuthorizationRequestRepository.java)
> The else/else if block and catch block should be on the same line of the closing brace of previous try/if block.

Actually I saw class OncePerRequestFilter from which I followed the styling and also I came across link https://github.com/spring-projects/spring-framework/wiki/Code-Style which states 

> - Line break after the closing brace if that brace terminates a statement or the body of a method, constructor, or named class
> - Line break before else, catch and finally statements

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,231 @@
...
-import javax.servlet.ServletException;
-import javax.servlet.http.HttpServletRequest;
-import javax.servlet.http.HttpServletResponse;
-import java.io.IOException;
+import org.springframework.web.util.UriComponentsBuilder;
 
 /**
  * @author Joe Grandja
+ * @author Paurav Munshi
+ * @since 0.0.1
```
kratostaine, line 55: Javadoc to be added

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,231 @@
...
  */
 public class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilter {
-	private Converter<HttpServletRequest, OAuth2AuthorizationRequest> authorizationRequestConverter;
+
+	private static final String DEFAULT_ENDPOINT = "/oauth2/authorize";
```
kratostaine, line 59: This should probably be named `DEFAULT_AUTHORIZATION_ENDPOINT_URI `

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,231 @@
...
+
+	public final void setAuthorizationEndpointMatcher(RequestMatcher authorizationEndpointMatcher) {
+		Assert.notNull(authorizationEndpointMatcher, "authorizationEndpointMatcher cannot be set to null");
+		this.authorizationEndpointMatcher = authorizationEndpointMatcher;
+	}
+
+	@Override
+	protected boolean shouldNotFilter(HttpServletRequest request) throws ServletException {
+		boolean pathMatch = this.authorizationEndpointMatcher.matches(request);
+		String responseType = request.getParameter(OAuth2ParameterNames.RESPONSE_TYPE);
```
kratostaine, line 100: According to the spec of [Error response](https://tools.ietf.org/html/rfc6749#section-4.1.2.1) for Authorization request, the server should respond with `invalid_request` if missing an required parameter or `unsupported_response_type` for an invalid response type. So ideally the check for `response_type` should be added at the place of validating the auth request.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -15,33 +15,231 @@
...
+		return client;
+
+	}
+
+	private OAuth2Authorization buildOAuth2Authorization(Authentication auth, RegisteredClient client,
+			OAuth2AuthorizationRequest authorizationRequest, String code) {
+		OAuth2Authorization authorization = OAuth2Authorization.withRegisteredClient(client)
+					.principalName(auth.getPrincipal().toString())
+					.attribute(TokenType.AUTHORIZATION_CODE.getValue(), code)
+					.attributes(attirbutesMap -> attirbutesMap.putAll(authorizationRequest.getAttributes()))
```
kratostaine, line 185: Minor typo in `attirbutesMap`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: I am thinking of adding Javadoc once everything is finalized

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
paurav-munshi: That was how it was. I, initially, had the request validation also perform response_type check. But @jgrandja suggested to perform this check at the time of verifying if filter is applicable or not. Please look at this [comment](https://github.com/spring-projects-experimental/spring-authorization-server/pull/77#discussion_r423099177). And I think this is also a good approach because `response_type `could also be `token ` and in that case a different `Filter` will handle the request and I think for erroneous `response_type` we can have an separate error filter. I am fine with both the options, but just wanted to point out that this was kept this way deliberately.

jgrandja:
Thanks for all the updates @paurav-munshi.

I decided to add a polish commit in order to get this merged, as I'm targeting to complete the first iteration of `authorization_code` grant later next week. The implementation had some missing/invalid  validation checks as per spec. As well, I didn't feel the implementation had the same _"look-and-feel"_ as the OAuth 2.0 support in Spring Security 5.x classes, eg. coding style/patterns/conventions.

I would also recommend doing a comparison of `OAuth2AuthorizationEndpointFilter` against `OAuth2AuthorizationRequestRedirectFilter` and `OAuth2AuthorizationCodeGrantFilter` to better understand the implementation we are looking for. We are really looking to align this codebase with the OAuth 2.0 codebase in Spring Security 5.x.

Issue #68, opened 2020-04-24
Implement authorization_code AuthenticationProvider


jgrandja:
The `OAuth2AuthorizationCodeAuthenticationProvider` is responsible for authenticating the authorization `code` parameter..

### Implementation Requirements

- the `OAuth2TokenEndpointFilter` indirectly calls this `AuthenticationProvider`
- the `RegisteredClientRepository` should be used to validate the `client_id` parameter if the client was not previously authenticated
- the `OAuth2AuthorizationService` should be used to lookup the `OAuth2Authorization` based on the `code` parameter
- the `accessTokenGenerator` should be used to generate an opaque access token. NOTE: This will later be re-factored to generate a JWT
- javadoc `class` and `public` methods
- Unit tests

### Specification References

[3.1.  Token Endpoint](https://tools.ietf.org/html/rfc6749#section-3.2)
[4.1.  Authorization Code Grant](https://tools.ietf.org/html/rfc6749#section-4.1)
[4.1.3.  Access Token Request](https://tools.ietf.org/html/rfc6749#section-4.1.3)
[4.1.4.  Access Token Response](https://tools.ietf.org/html/rfc6749#section-4.1.4)

dfcoffin:
@jgrandja We had discussed the ability to allow an implementation to enhance the generated access token response, but I don't see that described above in the `Implementation Requirements`

dfcoffin:
@jgrandja Is it possible to update the `Implementation Requirements` section with Issue numbers?

- the OAuth2TokenEndpointFilter #67 indirectly calls this AuthenticationProvider

Multiple `Implementation Requirements` require the use of `RegisteredClientRepository`, and `OAuth2AuthorizationService`.  Are implementers of issues that require these classes to be implemented expected to develop their own versions or should these be considered as additional issues?

It would be nice if we had a method to collaborate with individuals developing dependent issues other than through comments in the issue they are developing.

jgrandja:
@dfcoffin I updated the issues to cross reference the dependent issues.

Customizing the token response will come in a later iteration. The implementation requirements are for this iteration only. We will likely need to go through 3 or more iterations to full realize the `authorization_code` grant, which will include full customizations as per the framework.

dfcoffin:
@jgrandja Thanks for the update.  After reading several of your PR review comments, I'm getting a better feeling for the overall project plan and code development approach.  Is this issue still available?

jgrandja:
[...]

PR #79, opened 2020-05-07
Added token endpoint implementation

### LINKED ISSUE #67: Implement Token Endpoint

kratostaine:
Fixes #67 

**Implementation Requirements**

- [x] the Filter should process requests for the (default) path `/oauth2/token`
- [x] `authorizationGrantConverter` should convert a valid `authorization_code` Access Token Request to `OAuth2AuthorizationCodeAuthenticationToken`
- [x] the `OAuth2AuthorizationCodeAuthenticationToken` should be passed to the `AuthenticationManager`
- [x] the `AuthenticationManager` should be composed of `OAuth2AuthorizationCodeAuthenticationProvider`
- [x] `OAuth2AccessTokenAuthenticationToken` should be returned from `AuthenticationManager`
- [x] the `OAuth2AccessToken` should be updated in the in-flight `OAuth2Authorization` 
- [x] failure scenarios
- [x] javadoc class and public methods
- [x] Unit tests

### DIFF
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
@@ -33,6 +33,7 @@ import java.util.function.Consumer;
  *
  * @author Joe Grandja
  * @author Krisztian Toth
+ * @author Madhu Bhat
  * @since 0.0.1
  * @see RegisteredClient
  * @see OAuth2AccessToken
@@ -72,6 +73,15 @@ public class OAuth2Authorization implements Serializable {
 	 */
 	public OAuth2AccessToken getAccessToken() {
 		return this.accessToken;
+	}
+
+	/**
+	 * Sets the access token {@link OAuth2AccessToken} in the {@link OAuth2Authorization}.
+	 *
+	 * @param accessToken the access token
+	 */
+	public final void setAccessToken(OAuth2AccessToken accessToken) {
+		this.accessToken = accessToken;
 	}
 
 	/**
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationToken.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationToken.java
@@ -25,6 +25,7 @@ import java.util.Collections;
 
 /**
  * @author Joe Grandja
+ * @author Madhu Bhat
  */
 public class OAuth2AccessTokenAuthenticationToken extends AbstractAuthenticationToken {
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
@@ -48,5 +49,14 @@ public class OAuth2AccessTokenAuthenticationToken extends AbstractAuthentication
 	@Override
 	public Object getPrincipal() {
 		return null;
+	}
+
+	/**
+	 * Returns the access token {@link OAuth2AccessToken}.
+	 *
+	 * @return the access token
+	 */
+	public OAuth2AccessToken getAccessToken() {
+		return this.accessToken;
 	}
 }
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
@@ -24,6 +24,7 @@ import java.util.Collections;
 
 /**
  * @author Joe Grandja
+ * @author Madhu Bhat
  */
 public class OAuth2AuthorizationCodeAuthenticationToken extends AbstractAuthenticationToken {
 	private static final long serialVersionUID = SpringSecurityCoreVersion.SERIAL_VERSION_UID;
@@ -56,5 +57,14 @@ public class OAuth2AuthorizationCodeAuthenticationToken extends AbstractAuthenti
 	@Override
 	public Object getPrincipal() {
 		return null;
+	}
+
+	/**
+	 * Returns the code.
+	 *
+	 * @return the code
+	 */
+	public String getCode() {
+		return this.code;
 	}
 }
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
@@ -15,10 +15,31 @@
  */
 package org.springframework.security.oauth2.server.authorization.web;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.core.convert.converter.Converter;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
+import org.springframework.security.oauth2.server.authorization.OAuth2Authorization;
 import org.springframework.security.oauth2.server.authorization.OAuth2AuthorizationService;
+import org.springframework.security.oauth2.server.authorization.TokenType;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AccessTokenAuthenticationToken;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeAuthenticationToken;
+import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
+import org.springframework.security.web.util.matcher.RequestMatcher;
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.web.filter.OncePerRequestFilter;
 
 import javax.servlet.FilterChain;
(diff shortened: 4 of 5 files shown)


kratostaine:
@jgrandja Have added the initial changes for early feedback. Tests, javadoc to follow. Also need your help to understand the required failure scenarios to be handled here so that I don't add unnecessary checks and failure handling.

jgrandja:
@kratostaine Please review [4.1.3.  Access Token Request](https://tools.ietf.org/html/rfc6749#section-4.1.3) to understand all the validations that need to be applied. At minimum, you will need to validate the REQUIRED parameters. However, the some of the parameter values will be validated by the `OAuth2AuthorizationCodeAuthenticationProvider` #68.

If you are still unsure e on what to validate, please send me the reference in the spec and I could let you know if it should be performed in  `OAuth2TokenEndpointFilter` or `OAuth2AuthorizationCodeAuthenticationProvider`.

kratostaine:
```
  grant_type
         REQUIRED.  Value MUST be set to "authorization_code".

   code
         REQUIRED.  The authorization code received from the
         authorization server.

   redirect_uri
         REQUIRED, if the "redirect_uri" parameter was included in the
         authorization request as described in Section 4.1.1, and their
         values MUST be identical.

   client_id
         REQUIRED, if the client is not authenticating with the
         authorization server as described in Section 3.2.1.
```
@jgrandja the above 4 are the reqd parameters as per the spec. I have done a validation on `grant_type` already. I also need to do a non-nullable string validation on `code`, `redirect_uri` and `client_id` in the `OAuth2TokenEndpointFilter` right? Also, how do I validate here the `redirect_uri` to have value identical to the one as part of the authorization request?

jgrandja:
@kratostaine 

> I also need to do a non-nullable string validation on `code`, `redirect_uri` and `client_id` in the `OAuth2TokenEndpointFilter` right? 

Yes for `code` and `redirect_uri`. The (confidential) client is required to authenticate in the `authorization_code` flow,  so the expectation is that `OAuth2ClientAuthenticationToken` #72 is available in the `SecurityContext`.

> how do I validate here the r`edirect_uri` to have value identical to the one as part of the authorization request?

The `OAuth2AuthorizationCodeAuthenticationProvider` #68 is responsible for this validation.

kratostaine:
Tests to follow. @jgrandja Please review this so that if there are any changes required, I won't have to rework on the tests again.

kratostaine:
@jgrandja Also regarding `AuthenticationManager` implementation, I have created a custom class implementing it as `OAuth2ClientAuthenticationManager`. Is that fine or should I be using the default implementation of `ProviderManager`?

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationManager.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+
+/**
+ * @author Madhu Bhat
+ */
+public class OAuth2ClientAuthenticationManager implements AuthenticationManager {
```
jgrandja, line 26: Please remove this class and add a constructor in `OAuth2TokenEndpointFilter` that accepts an `AuthenticationManager` as the expectation is that the `AuthenticationManager` will be composed of `OAuth2AuthorizationCodeAuthenticationProvider` - in a later story.

core/src/main/java/org/springframework/security/oauth2/server/authorization/converter/OAuth2AuthorizationCodeAuthenticationTokenConverter.java
```diff
@@ -0,0 +1,39 @@
...
+ * converter does not validate the request but only performs a conversion.
+ * @author Madhu Bhat
+ */
+public class OAuth2AuthorizationCodeAuthenticationTokenConverter implements Converter<HttpServletRequest, OAuth2AuthorizationCodeAuthenticationToken> {
+
+	@Override
+	public OAuth2AuthorizationCodeAuthenticationToken convert(HttpServletRequest request) {
+		return new OAuth2AuthorizationCodeAuthenticationToken(
+				request.getParameter(OAuth2ParameterNames.CODE),
+				request.getParameter(OAuth2ParameterNames.CLIENT_ID),
```
jgrandja, line 35: Please see my earlier [comment](https://github.com/spring-projects-experimental/spring-authorization-server/pull/79#issuecomment-625897548):

> so the expectation is that `OAuth2ClientAuthenticationToken` #72 is available in the `SecurityContext` - no need to validate client_id

The other constructor should be called - the one accepting `Authentication clientPrincipal`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -29,16 +44,56 @@
 
 /**
  * @author Joe Grandja
+ * @author Madhu Bhat
  */
 public class OAuth2TokenEndpointFilter extends OncePerRequestFilter {
-	private Converter<HttpServletRequest, Authentication> authorizationGrantConverter;
+	private final String DEFAULT_TOKEN_URI = "/oauth2/token";
```
jgrandja, line 50: Rename to `DEFAULT_TOKEN_ENDPOINT_URI` and make `static`. We should also provide a constructor that accepts `tokenEndpointUri` allowing for customization. See `OAuth2AuthorizationRequestRedirectFilter` in `spring-security-oauth2-client` for an example.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -29,16 +44,56 @@
...
+	/**
+	 * Constructs an {@code OAuth2TokenEndpointFilter} using the provided parameter.
+	 *
+	 * @param authorizationService the authorization service implementation
+	 */
+	public OAuth2TokenEndpointFilter(OAuth2AuthorizationService authorizationService) {
+		Assert.notNull(authorizationService, "authorizationService cannot be null");
+		this.authorizationGrantConverter = new OAuth2AuthorizationCodeAuthenticationTokenConverter();
+		this.authenticationManager = new OAuth2ClientAuthenticationManager();
+		this.uriMatcher = new AntPathRequestMatcher(DEFAULT_TOKEN_URI, HttpMethod.POST.name());
```
jgrandja, line 65: In addition to the path, the `RequestMatcher` should also match on parameter `grant_type=authorization_code`

core/src/main/java/org/springframework/security/oauth2/server/authorization/converter/OAuth2AuthorizationCodeAuthenticationTokenConverter.java
```diff
@@ -0,0 +1,39 @@
...
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeAuthenticationToken;
+
+import javax.servlet.http.HttpServletRequest;
+
+/**
+ * Converts from a {@link HttpServletRequest} to an {@link OAuth2AuthorizationCodeAuthenticationToken} that can be authenticated. The
+ * converter does not validate the request but only performs a conversion.
+ * @author Madhu Bhat
+ */
+public class OAuth2AuthorizationCodeAuthenticationTokenConverter implements Converter<HttpServletRequest, OAuth2AuthorizationCodeAuthenticationToken> {
```
jgrandja, line 29: Let's make this a private method in `OAuth2TokenEndpointFilter`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -29,16 +44,56 @@
...
+		}
+		filterChain.doFilter(request, response);
+	}
 
+	private String validateAccessTokenRequest(HttpServletRequest request) {
+		if (!AuthorizationGrantType.AUTHORIZATION_CODE.getValue().equals(request.getParameter(OAuth2ParameterNames.GRANT_TYPE))) {
+			return OAuth2ErrorCodes.UNSUPPORTED_GRANT_TYPE;
+		} else if (StringUtils.isEmpty(request.getParameter(OAuth2ParameterNames.CODE))
+				|| StringUtils.isEmpty(request.getParameter(OAuth2ParameterNames.REDIRECT_URI))) {
+			return OAuth2ErrorCodes.INVALID_REQUEST;
```
jgrandja, line 95: Instead of returning `String`, just `throw OAuth2AuthenticationException`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -29,16 +44,56 @@
...
+		if (uriMatcher.matches(request)) {
+			String errorCode = validateAccessTokenRequest(request);
+			if (errorCode == null) {
+				OAuth2AuthorizationCodeAuthenticationToken authCodeAuthToken = authorizationGrantConverter.convert(request);
+				OAuth2AccessTokenAuthenticationToken accessTokenAuthenticationToken =
+						(OAuth2AccessTokenAuthenticationToken) authenticationManager.authenticate(authCodeAuthToken);
+				OAuth2Authorization authorization = authorizationService
+						.findByTokenAndTokenType(authCodeAuthToken.getCode(), TokenType.AUTHORIZATION_CODE);
+				authorization.setAccessToken(accessTokenAuthenticationToken.getAccessToken());
+				authorizationService.save(authorization);
```
jgrandja, line 82: The access token response should be written to `response` and then `return`.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
kratostaine: So in case `grant_type!=authorization_code` shouldn't the auth server send back `unsupported_grant_type` as an error response as per [5.2 Error Response](https://tools.ietf.org/html/rfc6749#section-5.2)? Having the grant_type param check in the requestMatcher would just ignore the request in case the request has an invalid `grant_type`. Is that fine?

core/src/main/java/org/springframework/security/oauth2/server/authorization/converter/OAuth2AuthorizationCodeAuthenticationTokenConverter.java
kratostaine: So no need of the existing `authorizationGrantConverter` converter object in `OAuth2TokenEndpointFilter` right?

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
kratostaine: But the issue requirements mention
> the `OAuth2AccessToken` should be updated in the in-flight `OAuth2Authorization`

So should the `OAuth2AccessToken` be updated in `OAuth2Authorization` or be written to the `response`?

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
jgrandja: > So in case grant_type!=authorization_code shouldn't the auth server send back unsupported_grant_type as an error response

Yes, you are correct. So let's just match on the  `grant_type ` parameter and return `unsupported_grant_type` if there isn't a match on the value of the parameter.

core/src/main/java/org/springframework/security/oauth2/server/authorization/converter/OAuth2AuthorizationCodeAuthenticationTokenConverter.java
jgrandja: Let's keep `authorizationGrantConverter` and simply point by method reference to private method. We will likely expose a setter (later) to allow for customizations of `authorizationGrantConverter`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
jgrandja: The `OAuth2AccessToken` needs to be updated in `OAuth2Authorization` AND be written to the response? I didn't specify the requirement to write to response because it's implied as per [4.1.4.  Access Token Response](https://tools.ietf.org/html/rfc6749#section-4.1.4)

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
kratostaine: Just ensuring that we're on the same page, we don't need the grant_type added to the requestMatcher, but validated as part of the request validation, correct?

jgrandja: Sorry for not being clear. The `RequestMatcher` should also match on the existence of the `grant_type` parameter name (not value). The parameter value will be used after to determine if it can be converted

kratostaine: @jgrandja just one doubt though. We can match with RequestMatcher if `grant_type` was passed as a query param using a regex with `RegexRequestMatcher`. However, if the `grant_type` was passed within the posted form data, the same cannot be matched with `RequestMatcher` right? Your thoughts on that please?

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
kratostaine: @jgrandja and should the exception be just thrown or also caught and written to the response with `response.sendError`?

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
jgrandja: Take a look at `OAuth2AuthorizationCodeGrantFilter.matchesAuthorizationResponse()`, which demonstrates more advanced request matching. **NOTE:** Although `OAuth2AuthorizationCodeGrantFilter` does not use a `RequestMatcher`, the `matchesAuthorizationResponse()` method can be referenced as a `RequestMatcher`.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
jgrandja: The `OAuth2Error` should also be written to the response.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
kratostaine: @jgrandja as per [section-4.1.3](https://tools.ietf.org/html/rfc6749#section-4.1.3), `grant_type` is a required parameter for an access token request and as per [section-5.2](https://tools.ietf.org/html/rfc6749#section-5.2) which is referenced by [section-4.1.4](https://tools.ietf.org/html/rfc6749#section-4.1.4) of access token response, the response should have error as
>invalid_request
               The request is missing a required parameter

Doesn't this mean that the `RequestMatcher` shouldn't include existence of `grant_type`, but rather be validated post the request being matched and validated along with the rest of the required params for the existence of `grant_type` param?

jgrandja: @kratostaine Thank you for reading the details in the spec! And for keeping me in check :) You are correct. If the `grant_type`  parameter is not provided then an `OAuth2Error` should be returned with `invalid_request`. So no need to use `grant_type` in the `RequestMatcher`.

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -21,11 +21,15 @@
 
 /**
  * @author Joe Grandja
+ * @author Madhu Bhat
  */
 public class OAuth2Authorization {
```
anoopgarlapati, line 26: Javadoc for class is required as per issue requirements.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationToken.java
```diff
@@ -25,6 +25,7 @@
 
 /**
  * @author Joe Grandja
+ * @author Madhu Bhat
  */
 public class OAuth2AccessTokenAuthenticationToken extends AbstractAuthenticationToken {
```
anoopgarlapati, line 30: Javadoc for class is required as per issue requirements.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
```diff
@@ -24,6 +24,7 @@
 
 /**
  * @author Joe Grandja
+ * @author Madhu Bhat
  */
 public class OAuth2AuthorizationCodeAuthenticationToken extends AbstractAuthenticationToken {
```
anoopgarlapati, line 29: Javadoc for class is required as per issue requirements.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -15,30 +15,150 @@
...
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
+import java.io.Writer;
 
 /**
  * @author Joe Grandja
+ * @author Madhu Bhat
  */
 public class OAuth2TokenEndpointFilter extends OncePerRequestFilter {
```
anoopgarlapati, line 52: Javadoc for class is required as per issue requirements.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -15,30 +15,150 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
 import org.springframework.core.convert.converter.Converter;
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.MediaType;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.core.Authentication;
+import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.oauth2.core.*;
```
anoopgarlapati, line 28: As per [Spring framework code style document](https://github.com/spring-projects/spring-framework/wiki/Code-Style#import-statements), wildcard imports should be avoided.
> Wildcard imports such as `import java.util.*`; or` import static org.assertj.core.api.Assertions.*` are forbidden, even in test code.

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilterTests.java
```diff
@@ -0,0 +1,215 @@
...
+
+import javax.servlet.FilterChain;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+import java.time.Instant;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.Mockito.*;
```
anoopgarlapati, line 43: As per [Spring framework code style document](https://github.com/spring-projects/spring-framework/wiki/Code-Style#import-statements), wildcard imports should be avoided.
> Wildcard imports such as `import java.util.*`; or` import static org.assertj.core.api.Assertions.*` are forbidden, even in test code.

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
kratostaine: Added only to the setter that I added and not to the class as there is an open PR with javadoc for the class.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationToken.java
kratostaine: Added only to the getter that I added and not to the class as there is an open PR with javadoc for the class.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationToken.java
kratostaine: Added only to the getter that I added and not to the class as there is an open PR with javadoc for the class.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
kratostaine: Added

kratostaine:
@jgrandja could you please help with reviewing this?

jgrandja:
Thank you for all the updates @kratostaine!

Overall, the implementation looked good and complied to spec for the most part.

I did add a polish commit in order to get this merged, as I'm hoping to complete the first iteration of `authorization_code` grant later this week. I added some missing validation checks and re-organized some of the code.

Thank you again and I look forward to another contribution 👍

Issue #83, opened 2020-05-28
Implement Token Revocation Endpoint


jgrandja:
An authorization server should provide an endpoint that allows revoking an access token or refresh token.

The `OAuth2TokenRevocationEndpointFilter` should be implemented as a `Filter`.

### Implementation Requirements

- the `Filter` should process `POST` requests for the (default) path `/oauth2/revoke`
- The client must be authenticated #39 before the Revocation Request can be accepted
- the implementation must support the revocation of access tokens and refresh tokens
- the `Filter` should use the `OAuth2AuthorizationService` to _invalidate_ the token
- javadoc `class` and `public` methods
- Unit tests

### Specification References

[OAuth 2.0 Token Revocation](https://tools.ietf.org/html/rfc7009)

synek:
I like to have a go at this one @jgrandja if that's possible.

jgrandja:
Thanks for your interest @synek. This has already been assigned to an internal team member. Please keep an eye out for other issues coming up.

PR #84, opened 2020-06-05
Implement Token Revocation Endpoint

babuv2:
Fixes gh-83

### DIFF
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/DefaultOAuth2TokenRevocationService.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/DefaultOAuth2TokenRevocationService.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization;
+
+import org.springframework.util.Assert;
+
+/**
+ * An {@link OAuth2TokenRevocationService} that revokes tokens.
+ *
+ * @author Vivek Babu
+ * @see OAuth2AuthorizationService
+ * @since 0.0.1
+ */
+public final class DefaultOAuth2TokenRevocationService implements OAuth2TokenRevocationService {
+
+	private OAuth2AuthorizationService authorizationService;
+
+	/**
+	 * Constructs an {@code DefaultOAuth2TokenRevocationService}.
+	 */
+	public DefaultOAuth2TokenRevocationService(OAuth2AuthorizationService authorizationService) {
+		Assert.notNull(authorizationService, "authorizationService cannot be null");
+		this.authorizationService = authorizationService;
+	}
+
+	@Override
+	public void revoke(String token, TokenType tokenType) {
+		final OAuth2Authorization authorization = this.authorizationService.findByTokenAndTokenType(token, tokenType);
+		if (authorization != null) {
+			final OAuth2Authorization revokedAuthorization = OAuth2Authorization.from(authorization)
+					.revoked(true).build();
+			this.authorizationService.save(revokedAuthorization);
+		}
+	}
+}
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
@@ -1,0 +1,99 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.authentication;
+
+import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+import org.springframework.security.oauth2.server.authorization.OAuth2Authorization;
+import org.springframework.security.oauth2.server.authorization.OAuth2TokenRevocationService;
+import org.springframework.security.oauth2.server.authorization.TokenType;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
+import org.springframework.security.oauth2.server.authorization.OAuth2AuthorizationService;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link AuthenticationProvider} implementation for the OAuth 2.0 Token Revocation.
+ *
+ * @author Vivek Babu
+ * @since 0.0.1
+ * @see OAuth2TokenRevocationAuthenticationToken
+ * @see OAuth2AuthorizationService
+ * @see OAuth2TokenRevocationService
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7009#section-2.1">Section 2.1 Revocation Request</a>
+ */
+public class OAuth2TokenRevocationAuthenticationProvider implements AuthenticationProvider {
+
+	private OAuth2AuthorizationService authorizationService;
+	private OAuth2TokenRevocationService tokenRevocationService;
+
+	/**
+	 * Constructs an {@code OAuth2TokenRevocationAuthenticationProvider} using the provided parameters.
+	 *
+	 * @param authorizationService the authorization service
+	 * @param tokenRevocationService the token revocation service
+	 */
+	public OAuth2TokenRevocationAuthenticationProvider(OAuth2AuthorizationService authorizationService,
+			OAuth2TokenRevocationService tokenRevocationService) {
+		Assert.notNull(authorizationService, "authorizationService cannot be null");
+		Assert.notNull(tokenRevocationService, "tokenRevocationService cannot be null");
+		this.authorizationService = authorizationService;
+		this.tokenRevocationService = tokenRevocationService;
+	}
+
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		OAuth2TokenRevocationAuthenticationToken tokenRevocationAuthenticationToken =
+				(OAuth2TokenRevocationAuthenticationToken) authentication;
+
+		OAuth2ClientAuthenticationToken clientPrincipal = null;
+		if (OAuth2ClientAuthenticationToken.class.isAssignableFrom(tokenRevocationAuthenticationToken.getPrincipal()
+				.getClass())) {
+			clientPrincipal = (OAuth2ClientAuthenticationToken) tokenRevocationAuthenticationToken.getPrincipal();
+		}
+		if (clientPrincipal == null || !clientPrincipal.isAuthenticated()) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_CLIENT));
+		}
+
+		final RegisteredClient registeredClient = clientPrincipal.getRegisteredClient();
+		final String tokenTypeHint = tokenRevocationAuthenticationToken.getTokenTypeHint();
+		final String token = tokenRevocationAuthenticationToken.getToken();
+		final OAuth2Authorization authorization = authorizationService.findByTokenAndTokenType(token,
+				TokenType.ACCESS_TOKEN);
+
+		OAuth2TokenRevocationAuthenticationToken successfulAuthentication =
+				new OAuth2TokenRevocationAuthenticationToken(token, registeredClient, tokenTypeHint);
+
+		if (authorization == null) {
+			return successfulAuthentication;
+		}
+
+		if (!registeredClient.getClientId().equals(authorization.getRegisteredClientId())) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_GRANT));
+		}
+
+		tokenRevocationService.revoke(token, TokenType.ACCESS_TOKEN);
+		return successfulAuthentication;
+	}
+
+	@Override
+	public boolean supports(Class<?> authentication) {
+		return OAuth2TokenRevocationAuthenticationToken.class.isAssignableFrom(authentication);
+	}
+}
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
@@ -1,0 +1,102 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.authentication;
+
+import org.springframework.lang.Nullable;
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.server.authorization.Version;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
+import org.springframework.util.Assert;
+
+import java.util.Collections;
+
+/**
+ * An {@link Authentication} implementation used for OAuth 2.0 Client Authentication.
+ *
+ * @author Vivek Babu
+ * @since 0.0.1
+ * @see AbstractAuthenticationToken
+ * @see RegisteredClient
+ * @see OAuth2TokenRevocationAuthenticationProvider
+ */
+public class OAuth2TokenRevocationAuthenticationToken extends AbstractAuthenticationToken {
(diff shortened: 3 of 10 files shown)


pivotal-issuemaster:
@babuv2 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects-experimental/spring-authorization-server&pullRequestId=84)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects-experimental/spring-authorization-server&pullRequestId=84) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@babuv2 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects-experimental/spring-authorization-server&pullRequestId=84)!

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
```diff
@@ -0,0 +1,103 @@
...
+	private Authentication authentication;
+	private String clientId;
+	private String token;
+	private String credentials;
+	private RegisteredClient registeredClient;
+
+	public OAuth2TokenRevocationAuthenticationToken(String token,
+			Authentication clientPrincipal, String tokenTypeHint) {
+		super(Collections.emptyList());
+		this.token = token;
```
jgrandja, line 41: Add `Assert.notNull` or `Assert.hasText` etc. for all required arguments

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
```diff
@@ -0,0 +1,107 @@
...
+import org.springframework.security.oauth2.server.authorization.OAuth2AuthorizationService;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClientRepository;
+import org.springframework.util.Assert;
+
+/**
+ * @author Vivek Babu
+ */
+public class OAuth2TokenRevocationAuthenticationProvider implements AuthenticationProvider {
+
+	public static final String TOKEN_TYPE_HINT = "token_type_hint";
```
jgrandja, line 37: Change modifier to `private` for `TOKEN_TYPE_HINT` and `UNSUPPORTED_TOKEN_TYPE`.

As well, rename:

`TOKEN_TYPE_HINT` -> `TOKEN_TYPE_HINT_ERROR_CODE`
`UNSUPPORTED_TOKEN_TYPE` -> `UNSUPPORTED_TOKEN_TYPE_ERROR_CODE`

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
```diff
@@ -0,0 +1,107 @@
...
+		this.registeredClientRepository = registeredClientRepository;
+		this.authorizationService = authorizationService;
+	}
+
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		OAuth2TokenRevocationAuthenticationToken authenticationToken = (OAuth2TokenRevocationAuthenticationToken)
+				authentication;
+
+		String clientId = authenticationToken.getPrincipal().toString();
```
jgrandja, line 62: We should assert on `OAuth2ClientAuthenticationToken`. Take a look at [OAuth2AuthorizationCodeAuthenticationProvider](https://github.com/spring-projects-experimental/spring-authorization-server/blob/master/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java#L76) and apply the same check for client auth.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
```diff
@@ -0,0 +1,107 @@
...
+
+		if (authorization == null) {
+			return successfulAuthentication;
+		}
+
+		if (!registeredClient.getClientId().equals(authorization.getRegisteredClientId())) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_GRANT));
+		}
+
+		if (token.equals(authorization.getAttribute(OAuth2AuthorizationAttributeNames.CODE))) {
```
jgrandja, line 84: It's not clear to me what this check is for?

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
```diff
@@ -0,0 +1,107 @@
...
+
+		this.authorizationService.invalidate(authorization);
+		return successfulAuthentication;
+	}
+
+	private OAuth2Authorization findAuthorizationUsingHint(String tokenTypeHint, String token) {
+		OAuth2Authorization authorization = null;
+		TokenType tokenType = new TokenType(tokenTypeHint);
+
+		if (TokenType.REFRESH_TOKEN.equals(tokenType) || TokenType.ACCESS_TOKEN.equals(tokenType)) {
```
jgrandja, line 96: Apologies as I did mention to allow the capability to revoke a refresh token but at the moment we don't have refresh token grant feature. Therefore, please remove all references to refresh token. We will add this in after we have refresh token feature implemented.

For now, we can assume the revocation request is for access tokens only. Any other token type hint should return 400 for now.

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2AuthorizationService.java
```diff
@@ -42,4 +43,19 @@
...
 	OAuth2Authorization findByTokenAndTokenType(String token, TokenType tokenType);
 
+	/**
+	 * Returns the {@link OAuth2Authorization} containing the provided {@code token},
+	 * or {@code null} if not found.
+	 *
+	 * @param token the token credential
+	 * @return the {@link OAuth2Authorization} if found, otherwise {@code null}
+	 */
+	OAuth2Authorization findByToken(String token);
```
jgrandja, line 53: I'm wondering if we need to add this operation? Instead we could consider changing `findByTokenAndTokenType` to `String token, @Nullable TokenType tokenType`? I think the name would change to. Thoughts?

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2AuthorizationService.java
```diff
@@ -42,4 +43,19 @@
...
+
+	/**
+	 * Invalidates the given {@link OAuth2Authorization}
+	 *
+	 * @param authorization {@link OAuth2Authorization} to be invalidated
+	 */
+	void invalidate(OAuth2Authorization authorization);
```
jgrandja, line 60: I'm not sure we should add this operation here. We will need the ability to revoke access tokens, refresh tokens and possibly authorization codes. This operation doesn't allow us to specify which token to revoke. I wonder if we should have a `revoke` on `OAuth2Authorization` instead. Thoughts?

babuv2:
@jgrandja Joe, Thanks for all the inputs 
I am working on all the changes that is requested. I will try to get them done by tomorrow.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenRevocationEndpointFilter.java
```diff
@@ -0,0 +1,169 @@
...
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7009#section-2.1">Section 2.1 Revocation Request</a>
+ * @since 0.0.1
+ */
+public class OAuth2TokenRevocationEndpointFilter extends OncePerRequestFilter {
+
+	/**
+	 * The default endpoint {@code URI} for token revocation request.
+	 */
+	public static final String DEFAULT_TOKEN_REVOCATION_ENDPOINT_URI = "/oauth2/revoke";
+	public static final String TOKEN_TYPE_HINT = "token_type_hint";
```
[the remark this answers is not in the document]
babuv2: Fixed

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenRevocationEndpointFilter.java
```diff
@@ -0,0 +1,169 @@
...
+
+	private static class TokenRevocationAuthenticationConverter implements
+			Converter<HttpServletRequest, Authentication> {
+
+		@Override
+		public Authentication convert(HttpServletRequest request) {
+			MultiValueMap<String, String> parameters = OAuth2EndpointUtils.getParameters(request);
+
+			// client_id (REQUIRED)
+			String clientId = parameters.getFirst(OAuth2ParameterNames.CLIENT_ID);
```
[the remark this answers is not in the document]
babuv2: Fixed

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
```diff
@@ -0,0 +1,103 @@
...
+		this.authentication = clientPrincipal;
+		this.tokenTypeHint = tokenTypeHint;
+		this.credentials = "";
+	}
+
+	public OAuth2TokenRevocationAuthenticationToken(String token,
```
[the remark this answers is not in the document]
babuv2: Fixed

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
```diff
@@ -0,0 +1,103 @@
...
+	private static final long serialVersionUID = Version.SERIAL_VERSION_UID;
+	private final String tokenTypeHint;
+	private Authentication authentication;
+	private String clientId;
+	private String token;
+	private String credentials;
+	private RegisteredClient registeredClient;
+
+	public OAuth2TokenRevocationAuthenticationToken(String token,
+			Authentication clientPrincipal, String tokenTypeHint) {
```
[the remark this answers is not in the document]
babuv2: Fixed

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * @author Vivek Babu
+ */
+public class OAuth2TokenRevocationAuthenticationToken extends AbstractAuthenticationToken {
+	private static final long serialVersionUID = Version.SERIAL_VERSION_UID;
+	private final String tokenTypeHint;
+	private Authentication authentication;
+	private String clientId;
+	private String token;
+	private String credentials;
```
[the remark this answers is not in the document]
babuv2: Fixed

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
babuv2: Added

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
babuv2: Done

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
babuv2: This is not needed Joe. I have removed it. As of now we ignore the hint and only look for access token

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java
babuv2: FiJoe the spec says that hint can be any value. if it is of value access_token or refresh_token the authorization server CAN use it to speed up its lookup. If any other value of token_type_hint is give it should not lead to any error

"An invalid token type hint value is ignored by the authorization
   server and does not influence the revocation response."

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2AuthorizationService.java
babuv2: Just looking up for access_token for now

core/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2AuthorizationService.java
babuv2: @jgrandja Joe this is the only place where I have a doubt, you mean to say that have a revoke method on the authorization and then we should do
authorization.revoke() and then authorizationService.save(authorization)? Because in that case with our current inmemory implementation one more record will get inserted with revoked status. Instead I thought we can have an invalidate method on the service and let the implementation decide how to handle the revoke. In this default inmemory implementation 
1. we remove the authorization and then
2. add it in revoked status
Other implementations may remove it from DB and handle other revocations etc. 
We can discuss this further Joe

jgrandja:
@babuv2 Regarding our discussion around renaming `OAuth2AuthorizationService` to `OAuth2AuthorizationRepository`, I'll hold off on this for now until we merge this PR. I'd like to see how things look after we merge and then determine if the rename makes sense or not.

babuv2:
> @babuv2 Regarding our discussion around renaming `OAuth2AuthorizationService` to `OAuth2AuthorizationRepository`, I'll hold off on this for now until we merge this PR. I'd like to see how things look after we merge and then determine if the rename makes sense or not.

@jgrandja Joe should I make any more changes to this PR?

jgrandja:
@babuv2 Yes. Do you recall our discussion around `OAuth2AuthorizationService.invalidate()`? As per [comment](https://github.com/spring-projects-experimental/spring-authorization-server/pull/84/files#r437102895), this operation does not specify which token to revoke. An authorization code, access token and refresh token may be revoked and the logic will differ depending on which token type it is.

We also agreed that we would extract `OAuth2AuthorizationService.invalidate()` to a new interface, named `OAuth2TokenRevocationService.revoke(...)`. The `OAuth2AuthorizationService` will likely be renamed (later) to `OAuth2AuthorizationRepository` as it will serve mainly for CRUD operations. The `OAuth2TokenRevocationService` implementation would use the `OAuth2AuthorizationService` to lookup the token to revoke and then update the "status" of the token before it saves it back.

babuv2:
@jgrandja  Joe, I have added the revocation service implementation

jgrandja:
Thanks @babuv2. I will try to review today if not first thing tomorrow.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenRevocationEndpointFilter.java
```diff
@@ -0,0 +1,149 @@
...
+		}
+	}
+
+	private void sendErrorResponse(HttpServletResponse response, OAuth2Error error) throws IOException {
+		ServletServerHttpResponse httpResponse = new ServletServerHttpResponse(response);
+		httpResponse.setStatusCode(HttpStatus.BAD_REQUEST);
+		this.errorHttpResponseConverter.write(error, null, httpResponse);
+	}
+
+	private static OAuth2AuthenticationException throwError(String errorCode, String parameterName) {
```
jgrandja, line 121: Should return `void`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenRevocationEndpointFilter.java
```diff
@@ -0,0 +1,149 @@
...
+
+			// token (REQUIRED)
+			String token = parameters.getFirst(TOKEN);
+			if (!StringUtils.hasText(token) ||
+					parameters.get(TOKEN).size() != 1) {
+				throwError(OAuth2ErrorCodes.INVALID_REQUEST, TOKEN);
+			}
+
+			// token_type_hint (OPTIONAL)
+			String tokenTypeHint = parameters.getFirst(TOKEN_TYPE_HINT);
```
jgrandja, line 144: We should validate that `TOKEN_TYPE_HINT` was not supplied with multiple values

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenRevocationEndpointFilter.java
```diff
@@ -0,0 +1,149 @@
...
+
+		if (!this.revocationEndpointMatcher.matches(request)) {
+			filterChain.doFilter(request, response);
+			return;
+		}
+
+		try {
+			Authentication tokenRevocationRequestAuthentication =
+					this.tokenRevocationAuthenticationConverter.convert(request);
+			this.authenticationManager.authenticate(tokenRevocationRequestAuthentication);
```
jgrandja, line 108: We should explicitly set status to `200` - even though this will default. However, it will ensure the response is written and therefore will commit.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationToken.java
```diff
@@ -0,0 +1,102 @@
...
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.server.authorization.Version;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
+import org.springframework.util.Assert;
+
+import java.util.Collections;
+
+/**
+ * An {@link Authentication} implementation used for OAuth 2.0 Client Authentication.
```
jgrandja, line 28: Typo in javadoc

core/src/main/java/org/springframework/security/oauth2/server/authorization/DefaultOAuth2TokenRevocationService.java
```diff
@@ -0,0 +1,48 @@
...
+		Assert.notNull(authorizationService, "authorizationService cannot be null");
+		this.authorizationService = authorizationService;
+	}
+
+	@Override
+	public void revoke(String token, TokenType tokenType) {
+		final OAuth2Authorization authorization = this.authorizationService.findByTokenAndTokenType(token, tokenType);
+		if (authorization != null) {
+			final OAuth2Authorization revokedAuthorization = OAuth2Authorization.from(authorization)
+					.revoked(true).build();
```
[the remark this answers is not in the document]
jgrandja: @babuv2 I've been giving this some further thought and I'd like to propose the following class as a holder of token metadata. The name maps nicely as referenced in the [Token Introspection RFC](https://tools.ietf.org/html/rfc7662#section-1):

> This specification defines a protocol that allows authorized
   protected resources to query the authorization server to determine
   the set of metadata for a given token that was presented to them by
   an OAuth 2.0 client.  This metadata includes whether or not the token
   is currently active (or if it has expired or otherwise been revoked)...

```java
public class OAuth2TokenMetadata<T extends AbstractOAuth2Token> implements Serializable {
	private final T token;
	private boolean revoked;
	private Instant revokedAt;

	public OAuth2TokenMetadata(T token) {
	}

	public T getToken() {
	}

	public boolean isExpired() {
	}

	public boolean isRevoked() {
	}

	public Instant getRevokedAt() {
	}

	public boolean isActive() {
	}
}
```

How about we start with this and see how it shapes up. Sounds good?

jgrandja:
Thanks for the PR @babuv2 ! This is now merged along with a follow-up polish commit to complete the feature.

Issue #82, opened 2020-05-28
Implement JWK Set Endpoint


jgrandja:
The authorization server should provide an endpoint that responds with the public key(s) used to verify the signature of a JWS. The public key(s) should be returned in the JWK Set format.

The `JwkSetEndpointFilter` should be implemented as a `Filter` similar to #31.

### Implementation Requirements

- the `Filter` should process `GET` requests for the (default) path `/oauth2/jwks`
- only public key(s) should be returned - private keys and symmetric keys must NOT be returned
- the `Filter` should have a _"KeyManager"_ that provides the public key(s) (also required by #81 )
- javadoc `class` and `public` methods
- Unit tests

### Specification References

[JSON Web Key (JWK)](https://tools.ietf.org/html/rfc7517)

synek:
How exactly does this relate to #2? Should this replace the existing implementation in #31 or should it be a separate filter matching `/oauth/jwks` instead of `.well-known/jwk_uris`?

Perhaps @ovidiupopa91 is interested in this as he requested #2 originally :)

ovidiupopa91:
Hi, yes, I'm interested in working on this issue.

jgrandja:
@synek The goal of #2 was to flush out the implementation so it can be used in #4. It really was meant to be a simplified implementation compared to what is required here. Ultimately, #2 will be replaced by this implementation.

@ovidiupopa91 I have an internal team that will be delivering the implementations for the JOSE stack. I plan on logging a couple of new issues next week so take a look there to see if anything interests you.

ovidiupopa91:
@jgrandja  ok, got it! I will take a look next week.

Issue #81, opened 2020-05-27
Add JwtEncoder to support JWT/JWS


jgrandja:
The authorization server should support JSON Web Token (JWT) as an access token format, as well, it MUST be signed - JSON Web Signature (JWS).

`JwtEncoder` is the counterpart to `JwtDecoder` in the `spring-security-oauth2-jose` module.

### Implementation Requirements

- define `interface JwtEncoder`
- provide an implementation of `JwtEncoder` - `NimbusJwtEncoder` **NOTE:** The implementation should use the [Nimbus JOSE + JWT](https://connect2id.com/products/nimbus-jose-jwt) library 
- the `NimbusJwtEncoder` should have a `Builder`
- the `NimbusJwtEncoder` should support `HS256` (HMAC using SHA-256)
- the `NimbusJwtEncoder` should support `RS256` (RSASSA-PKCS1-v1_5 using SHA-256)
- the `NimbusJwtEncoder` should have a _"KeyManager"_ that provides the public/private/symmetric keys (also required by #82 )
- the `NimbusJwtEncoder` should be used by `OAuth2AuthorizationCodeAuthenticationProvider` #68 to produce a JWS
- javadoc `class` and `public` methods
- Unit tests

### Specification References
[JSON Web Token (JWT)](https://tools.ietf.org/html/rfc7519)
[JSON Web Signature (JWS)](https://tools.ietf.org/html/rfc7515)
[Cryptographic Algorithms for Digital Signatures and MACs](https://tools.ietf.org/html/rfc7518#section-3)

idosal:
I'd love to be considered to take this when it's up for development.

jgrandja:
Thank you for your interest @idosal. It looks like I have a team (internal to VMware) that will work on delivering the JOSE stack. If anything changes, I'll be sure to let you know.

PR #87, opened 2020-06-11
Introduce JwtEncoder

anoopgarlapati:
Issue gh-81

**Implementation Requirements**
- [x] define interface JwtEncoder
- [x] provide an implementation of JwtEncoder - NimbusJwtEncoder NOTE: The implementation should use the Nimbus JOSE + JWT library
- [x] the NimbusJwtEncoder should have a Builder
- [x] the NimbusJwtEncoder should support HS256 (HMAC using SHA-256)
- [x] the NimbusJwtEncoder should support RS256 (RSASSA-PKCS1-v1_5 using SHA-256)
- [x] the NimbusJwtEncoder should have a "KeyManager" that provides the public/private/symmetric keys (also required by #82)
- [ ] the NimbusJwtEncoder should be used by OAuth2AuthorizationCodeAuthenticationProvider #68 to produce a JWS
- [ ] javadoc class and public methods
- [ ] Unit tests

### DIFF
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/JoseHeaders.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/JoseHeaders.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.jose;
+
+import org.springframework.core.convert.TypeDescriptor;
+import org.springframework.security.oauth2.core.converter.ClaimConversionService;
+import org.springframework.security.oauth2.jose.jws.MacAlgorithm;
+import org.springframework.security.oauth2.jose.jws.SignatureAlgorithm;
+import org.springframework.util.Assert;
+
+import java.net.URI;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Consumer;
+
+/**
+ * A representation of &quot;headers&quot; that may be contained
+ * in the JSON object JOSE Header of a JSON Web Signature (JWS).
+ *
+ * @author Anoop Garlapati
+ * @since 0.0.1
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7515#section-4">JOSE Header of JWS</a>
+ */
+public class JoseHeaders {
+
+	private final Map<String, Object> headers;
+
+	protected JoseHeaders(Map<String, Object> headers) {
+		Assert.notEmpty(headers, "headers cannot be empty");
+		this.headers = Collections.unmodifiableMap(new LinkedHashMap<>(headers));
+	}
+
+	public SignatureAlgorithm getSignatureAlgorithm() {
+		return SignatureAlgorithm.from(getHeaderAsString("alg"));
+	}
+
+	public MacAlgorithm getMacAlgorithm() {
+		return MacAlgorithm.from(getHeaderAsString("alg"));
+	}
+
+	public String getType() {
+		return getHeaderAsString("typ");
+	}
+
+	public Map<String, Object> getHeaders() {
+		return this.headers;
+	}
+
+	public String getHeaderAsString(String header) {
+		return !containsHeader(header) ? null :
+				ClaimConversionService.getSharedInstance().convert(getHeaders().get(header), String.class);
+	}
+
+	@SuppressWarnings("unchecked")
+	public List<String> getHeaderAsStringList(String header) {
+		if (!containsHeader(header)) {
+			return null;
+		}
+		final TypeDescriptor sourceDescriptor = TypeDescriptor.valueOf(Object.class);
+		final TypeDescriptor targetDescriptor = TypeDescriptor.collection(
+				List.class, TypeDescriptor.valueOf(String.class));
+		Object headerValue = getHeaders().get(header);
+		List<String> convertedValue = (List<String>) ClaimConversionService.getSharedInstance().convert(
+				headerValue, sourceDescriptor, targetDescriptor);
+		if (convertedValue == null) {
+			throw new IllegalArgumentException("Unable to convert header '" + header +
+					"' of type '" + headerValue.getClass() + "' to List.");
+		}
+		return convertedValue;
+	}
+
+	public URI getHeaderAsURI(String header) {
+		if (!containsHeader(header)) {
+			return null;
+		}
+		Object headerValue = getHeaders().get(header);
+		URI convertedValue = ClaimConversionService.getSharedInstance().convert(headerValue, URI.class);
+		if (convertedValue == null) {
+			throw new IllegalArgumentException("Unable to convert header '" + header +
+					"' of type '" + headerValue.getClass() + "' to URI.");
+		}
+		return convertedValue;
+	}
+
+	public Boolean getHeaderAsBoolean(String header) {
+		return !containsHeader(header) ? null :
+				ClaimConversionService.getSharedInstance().convert(getHeaders().get(header), Boolean.class);
+	}
+
+	private Boolean containsHeader(String header) {
+		Assert.notNull(header, "header cannot be null");
+		return getHeaders().containsKey(header);
+	}
+
+	public static class Builder {
+		private Map<String, Object> headers = new LinkedHashMap<>();
+
+		public Builder() {
+		}
+
+		public Builder(JoseHeaders joseHeaders) {
+			Assert.notNull(joseHeaders, "joseHeaders cannot be null");
+			this.headers = joseHeaders.headers;
+		}
+
+		public Builder header(String name, Object value) {
+			Assert.notNull(name, "name cannot be null");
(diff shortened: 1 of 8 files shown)


core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/KeyManager.java
```diff
@@ -0,0 +1,32 @@
...
+import java.security.Key;
+import java.util.List;
+
+/**
+ * A manager for the {@link Key}s.
+ *
+ * @author Anoop Garlapati
+ * @since 0.0.1
+ */
+public interface KeyManager {
```
jzheaux, line 27: I wonder if this should be called `KeyService`. Several other interfaces that retrieve resources in Spring Security are services, like `UserDetailsService`, `OAuth2UserService`, `AclService`, etc.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/JoseHeaders.java
```diff
@@ -0,0 +1,155 @@
...
+
+	public String getType() {
+		return getHeaderAsString("typ");
+	}
+
+	public Map<String, Object> getHeaders() {
+		return this.headers;
+	}
+
+	public String getHeaderAsString(String header) {
```
jzheaux, line 64: I realize that these are following the `JwtClaimAccessor` pattern; however, I think that it's better to coerce the values to the correct type on write instead of converting them on each read. 

Further, I think the `JwtClaimAccessor` makes a bit more sense when we are reading a JWT that another library has deserialized and chosen the types for each claim. Since we are the ones serializing in this case, I'm not seeing a huge need for these converters.

Is there any reason that we can't just have one method for now:

```java
public <H> H getHeader(String name) {
    return (H) this.headers.get(name)
}
```

and add the others later on, if needed?

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/NimbusJwtEncoder.java
```diff
@@ -0,0 +1,232 @@
...
+		}
+
+		public KeyManagerJwtEncoderBuilder jwsAlgorithm(JwsAlgorithm jwsAlgorithm) {
+			Assert.notNull(jwsAlgorithm, "jwsAlgorithm cannot be null");
+			this.jwsAlgorithm = jwsAlgorithm;
+			return this;
+		}
+
+		public NimbusJwtEncoder build() {
+			Key key = keyManager.getActiveKey();
```
jzheaux, line 206: Will `getActiveKey` guarantee non-null or should there also be an exception here to indicate that no key was returned? One way to address this would be to change the existing exception message from "Unsupported key provided." to "No supported key provided".

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/NimbusJwtEncoder.java
```diff
@@ -0,0 +1,232 @@
...
+				} else if (X5T.equals(name)) {
+					jwsHeaderBuilder = jwsHeaderBuilder.x509CertThumbprint(
+							Base64URL.from(headers.getHeaderAsString(X5T)));
+				} else if (X5T256.equals(name)) {
+					jwsHeaderBuilder = jwsHeaderBuilder.x509CertSHA256Thumbprint(
+							Base64URL.from(headers.getHeaderAsString(X5T256)));
+				} else if (X5C.equals(name)) {
+					jwsHeaderBuilder = jwsHeaderBuilder.x509CertChain(
+							X509CertChainUtils.toBase64List(JSONArrayUtils.parse(headers.getHeaderAsString(X5C))));
+				} else if (KID.equals(name)) {
```
jzheaux, line 169: How would the caller know this? It seems intrinsically tied to which key is used.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/NimbusJwtEncoder.java
jzheaux, line 206: Is the `build` method the right place to get the key? I'm thinking about key rotation.

It makes me wonder if the `KeyManager` should be in the `NimbusJwtEncoder` constructor.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/JoseHeaders.java
anoopgarlapati: I agree that since we are the ones serializing we might not need converters. I will simplify this in my next draft.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/NimbusJwtEncoder.java
anoopgarlapati: I'll add an exception here if a key is not returned and also will change the message.

anoopgarlapati: I missed this use-case. Yes, we do need to have `KeyManager` in the `NimbusJwtEncoder` and fetch the active key to create the signer before signing the JWT. I'll make the appropriate changes in the next draft.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/JoseHeaders.java
```diff
@@ -0,0 +1,155 @@
...
+		}
+		return convertedValue;
+	}
+
+	public Boolean getHeaderAsBoolean(String header) {
+		return !containsHeader(header) ? null :
+				ClaimConversionService.getSharedInstance().convert(getHeaders().get(header), Boolean.class);
+	}
+
+	private Boolean containsHeader(String header) {
```
krajcsovszkig-ms, line 105: Can this ever be null? Does it have to be a `Boolean`?

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/JwtClaimsSet.java
```diff
@@ -0,0 +1,75 @@
...
+
+	public static class Builder {
+		private Map<String, Object> claims = new LinkedHashMap<>();
+
+		public Builder() {
+		}
+
+		public Builder(JwtClaimsSet jwtClaimsSet) {
+			Assert.notNull(jwtClaimsSet, "jwtClaimsSet cannot be null");
+			this.claims = jwtClaimsSet.claims;
```
krajcsovszkig-ms, line 56: `claims` is an unmodifiable map, you should add all its entries to `this.claims` instead.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/NimbusJwtEncoder.java
```diff
@@ -0,0 +1,232 @@
...
+	}
+
+	@SuppressWarnings("StatementWithEmptyBody")
+	private JWSHeader createJwsHeader(JoseHeaders headers) {
+		JWSHeader.Builder jwsHeaderBuilder = new JWSHeader.Builder(
+				JWSAlgorithm.parse(headers.getHeaderAsString(ALG)));
+		// Parse optional and custom header parameters
+		Map<String, Object> headersMap = headers.getHeaders();
+		try {
+			for (final String name: headersMap.keySet()) {
```
krajcsovszkig-ms, line 136: I wonder if copying the map and getting each known key via `remove`, then only looping over the rest (the custom params) would be a bit nicer.

But if you do it in a loop, IMO you should loop over the `entrySet()` to avoid having to get the values separately for the keys you find, and perhaps a `switch` would be a little nicer than an `if` with this many branches.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/NimbusJwtEncoder.java
```diff
@@ -0,0 +1,232 @@
...
+					jwsHeaderBuilder = jwsHeaderBuilder.customParam(name, headersMap.get(name));
+				}
+			}
+		} catch (Exception ex) {
+			throw new JwtException(String.format(ENCODING_ERROR_MESSAGE_TEMPLATE, ex.getMessage()), ex);
+		}
+		return jwsHeaderBuilder.build();
+	}
+
+	private JWTClaimsSet createJwtClaimsSet(JwtClaimsSet claimsSet) {
```
krajcsovszkig-ms, line 183: This `JWTClaimsSet` vs `JwtClaimsSet` is a little confusing, it might be a good idea to write out the package name for one of them for clarity.

core/src/main/java/org/springframework/security/oauth2/server/authorization/jose/NimbusJwtEncoder.java
krajcsovszkig-ms: Do you think it would be worth trying to find a key from the manager that works with the selected algorithm?

anoopgarlapati:
> Given that the client needs this functionality as well, should this go into spring-security-oauth2-jose? Or are the two codebases going to be separate?

@krajcsovszkig-ms Thanks for the review comments. I am close to revising my draft. This is something brought to my attention by @jgrandja as well. I am refactoring the relevant classes into the spring-security-oauth-jose package in my next draft so that they can be extracted and merged into spring-security seamlessly.

krajcsovszkig-ms:
@anoopgarlapati great, thanks for your work on this!

krajcsovszkig-ms:
Hi @anoopgarlapati , are there any updates on this PR? Do you have an approximate ETA for when it will be ready?

Thanks!

anoopgarlapati:
@krajcsovszkig-ms Unfortunately due to my work related project I will not be able to meet the timeline for this task. I have discussed this with @jgrandja and he will be taking this task forward.

krajcsovszkig-ms:
Thanks for getting back!

@jgrandja do you have an approximate timeline for this task? Let me know if I can help to move it forward.

jgrandja:
@krajcsovszkig-ms I'll be working on it this week and next week. My goal is to have this merged later next week.

krajcsovszkig-ms:
That's great, thanks!

jgrandja:
Closing in favour of #96

Issue #51, opened 2020-04-21
Add support for Client Credentials Grant


jgrandja:
Add support for [Client Credentials Grant](https://tools.ietf.org/html/rfc6749#section-4.4).

aDelNogal:
Are you planning to add JWT authentication? [rfc7523](https://tools.ietf.org/html/rfc7523#section-2.2)

jgrandja:
@aDelNogal Yes. See the [roadmap](https://app.zenhub.com/workspaces/authorization-server-5e8f3182b5e8f5841bfc4902/roadmap) for planned features.

PR #88, opened 2020-06-11
Add client_credentials grant type support

alek-sys:
Fixes gh-51

### DIFF
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
@@ -1,0 +1,85 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.authentication;
+
+import java.time.Instant;
+import java.time.temporal.ChronoUnit;
+import java.util.Base64;
+import java.util.Set;
+
+import org.springframework.security.authentication.AuthenticationProvider;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.crypto.keygen.Base64StringKeyGenerator;
+import org.springframework.security.crypto.keygen.StringKeyGenerator;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+
+/**
+ * An {@link AuthenticationProvider} implementation for the OAuth 2.0 Client Credentials Grant.
+ *
+ * @author Alexey Nesterov
+ * @since 0.0.1
+ * @see OAuth2ClientCredentialsAuthenticationToken
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.4">Section 4.4 Client Credentials Grant</a>
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.4.2">Section 4.4.2 Access Token Request</a>
+ */
+
+public class OAuth2ClientCredentialsAuthenticationProvider implements AuthenticationProvider {
+
+	private final StringKeyGenerator accessTokenGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder());
+
+	@Override
+	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
+		OAuth2ClientCredentialsAuthenticationToken clientCredentialsAuthenticationToken =
+				(OAuth2ClientCredentialsAuthenticationToken) authentication;
+
+		OAuth2ClientAuthenticationToken clientPrincipal = null;
+		if (OAuth2ClientAuthenticationToken.class.isAssignableFrom(clientCredentialsAuthenticationToken.getPrincipal().getClass())) {
+			clientPrincipal = (OAuth2ClientAuthenticationToken) clientCredentialsAuthenticationToken.getPrincipal();
+		}
+
+		if (clientPrincipal == null || !clientPrincipal.isAuthenticated()) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_CLIENT));
+		}
+
+		Set<String> clientScopes = clientPrincipal.getRegisteredClient().getScopes();
+		Set<String> requestedScopes = clientCredentialsAuthenticationToken.getScopes();
+		if (!clientScopes.containsAll(requestedScopes)) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_SCOPE));
+		}
+
+		if (requestedScopes == null || requestedScopes.isEmpty()) {
+			requestedScopes = clientScopes;
+		}
+
+		String tokenValue = this.accessTokenGenerator.generateKey();
+		Instant issuedAt = Instant.now();
+		Instant expiresAt = issuedAt.plus(1, ChronoUnit.HOURS);		// TODO Allow configuration for access token lifespan
+		OAuth2AccessToken accessToken = new OAuth2AccessToken(OAuth2AccessToken.TokenType.BEARER,
+				tokenValue, issuedAt, expiresAt, requestedScopes);
+
+		return new OAuth2AccessTokenAuthenticationToken(
+				clientPrincipal.getRegisteredClient(), clientPrincipal, accessToken);
+	}
+
+	@Override
+	public boolean supports(Class<?> authentication) {
+		return OAuth2ClientCredentialsAuthenticationToken.class.isAssignableFrom(authentication);
+	}
+}
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
@@ -1,0 +1,68 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.security.oauth2.server.authorization.authentication;
+
+import java.util.Collections;
+import java.util.Set;
+
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.server.authorization.Version;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link Authentication} implementation used for the OAuth 2.0 Client Credentials Grant.
+ *
+ * @author Alexey Nesterov
+ * @since 0.0.1
+ * @see Authentication
+ * @see OAuth2ClientCredentialsAuthenticationProvider
+ */
+public class OAuth2ClientCredentialsAuthenticationToken extends AbstractAuthenticationToken {
+
+	private static final long serialVersionUID = Version.SERIAL_VERSION_UID;
+
+	private final Authentication clientPrincipal;
+	private final Set<String> scopes;
+
+	public OAuth2ClientCredentialsAuthenticationToken(Authentication clientPrincipal, Set<String> scopes) {
+		super(Collections.emptyList());
+		Assert.notNull(clientPrincipal, "clientPrincipal cannot be null");
+		Assert.notNull(scopes, "scopes cannot be null");
+		this.clientPrincipal = clientPrincipal;
+		this.scopes = scopes;
+	}
+
+	@SuppressWarnings("unchecked")
+	public OAuth2ClientCredentialsAuthenticationToken(OAuth2ClientAuthenticationToken clientPrincipal) {
+		this(clientPrincipal, Collections.EMPTY_SET);
+	}
+
+	@Override
+	public Object getCredentials() {
+		return "";
+	}
+
+	@Override
+	public Object getPrincipal() {
+		return this.clientPrincipal;
+	}
+
+	public Set<String> getScopes() {
+		return this.scopes;
+	}
+}
--- a/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
+++ b/core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
@@ -35,6 +35,8 @@ import org.springframework.security.oauth2.core.http.converter.OAuth2ErrorHttpMe
 import org.springframework.security.oauth2.server.authorization.OAuth2AuthorizationService;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AccessTokenAuthenticationToken;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeAuthenticationToken;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientAuthenticationToken;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientCredentialsAuthenticationToken;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
@@ -48,6 +50,11 @@ import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 import java.io.IOException;
 import java.time.temporal.ChronoUnit;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.Set;
 
 /**
  * A {@code Filter} for the OAuth 2.0 Authorization Code Grant,
@@ -86,8 +93,8 @@ public class OAuth2TokenEndpointFilter extends OncePerRequestFilter {
 	private final AuthenticationManager authenticationManager;
 	private final OAuth2AuthorizationService authorizationService;
 	private final RequestMatcher tokenEndpointMatcher;
-	private final Converter<HttpServletRequest, Authentication> authorizationGrantAuthenticationConverter =
-			new AuthorizationCodeAuthenticationConverter();
+	private final Converter<HttpServletRequest, Authentication> authorizationGrantAuthenticationConverter;
+
 	private final HttpMessageConverter<OAuth2AccessTokenResponse> accessTokenHttpResponseConverter =
 			new OAuth2AccessTokenResponseHttpMessageConverter();
 	private final HttpMessageConverter<OAuth2Error> errorHttpResponseConverter =
@@ -119,6 +126,11 @@ public class OAuth2TokenEndpointFilter extends OncePerRequestFilter {
 		this.authenticationManager = authenticationManager;
 		this.authorizationService = authorizationService;
 		this.tokenEndpointMatcher = new AntPathRequestMatcher(tokenEndpointUri, HttpMethod.POST.name());
+
+		Map<AuthorizationGrantType, Converter<HttpServletRequest, Authentication>> converters = new HashMap<>();
+		converters.put(AuthorizationGrantType.AUTHORIZATION_CODE, new AuthorizationCodeAuthenticationConverter());
+		converters.put(AuthorizationGrantType.CLIENT_CREDENTIALS, new ClientCredentialsAuthenticationConverter());
+		this.authorizationGrantAuthenticationConverter = new DelegatingAuthorizationGrantAuthenticationConverter(converters);
 	}
 
 	@Override
@@ -131,8 +143,16 @@ public class OAuth2TokenEndpointFilter extends OncePerRequestFilter {
 		}
 
 		try {
-			Authentication authorizationGrantAuthentication =
-					this.authorizationGrantAuthenticationConverter.convert(request);
+			String[] grantTypes = request.getParameterValues(OAuth2ParameterNames.GRANT_TYPE);
+			if (grantTypes == null || grantTypes.length == 0) {
+				throwError(OAuth2ErrorCodes.INVALID_REQUEST, "grant_type");
+			}
+
+			Authentication authorizationGrantAuthentication = this.authorizationGrantAuthenticationConverter.convert(request);
+			if (authorizationGrantAuthentication == null) {
+				throwError(OAuth2ErrorCodes.UNSUPPORTED_GRANT_TYPE, "grant_type");
+			}
+
 			OAuth2AccessTokenAuthenticationToken accessTokenAuthentication =
 					(OAuth2AccessTokenAuthenticationToken) this.authenticationManager.authenticate(authorizationGrantAuthentication);
 			sendAccessTokenResponse(response, accessTokenAuthentication.getAccessToken());
@@ -161,7 +181,7 @@ public class OAuth2TokenEndpointFilter extends OncePerRequestFilter {
 		this.errorHttpResponseConverter.write(error, null, httpResponse);
 	}
 
-	private static OAuth2AuthenticationException throwError(String errorCode, String parameterName) {
+	private static void throwError(String errorCode, String parameterName) {
 		OAuth2Error error = new OAuth2Error(errorCode, "OAuth 2.0 Parameter: " + parameterName,
 				"https://tools.ietf.org/html/rfc6749#section-5.2");
 		throw new OAuth2AuthenticationException(error);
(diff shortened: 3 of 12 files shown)


alek-sys:
Hey @jgrandja, I still have a couple of to-dos:
- [ ] Support basic auth for client id / secret
- [ ] Update sample app to use JWK to sign the token. This should allow to put together an end-to-end demo

With that in mind, feel free to provide any initial feedback you have.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
```diff
@@ -0,0 +1,66 @@
...
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClientRepository;
+import org.springframework.security.oauth2.server.authorization.token.TokenIssuer;
+
+/**
+ * An {@link AuthenticationProvider} that converts give {@link OAuth2ClientCredentialsAuthenticationToken} to a {@link OAuth2AccessTokenAuthenticationToken}
+ * using the provided token generator.
+ *
+ * @author Alexey Nesterov
+ */
+public class OAuth2ClientCredentialsAuthenticationProvider implements AuthenticationProvider {
```
jgrandja, line 34: #72 introduced `OAuth2ClientAuthenticationProvider` and `OAuth2ClientAuthenticationToken`, which handles client authentication. The `OAuth2ClientCredentialsAuthenticationProvider` should simply assert that the current `Authentication` is `OAuth2ClientAuthenticationToken`. Take a look at `OAuth2AuthorizationCodeAuthenticationProvider` and apply the same validation check for `OAuth2ClientAuthenticationToken`.

Also, take a look at the integration test `OAuth2AuthorizationCodeGrantTests.requestWhenTokenRequestValidThenResponseIncludesCacheHeaders()` to see how the client authenticates via Basic before it gets to `OAuth2AuthorizationCodeAuthenticationProvider`

core/src/main/java/org/springframework/security/oauth2/server/authorization/token/TokenIssuer.java
```diff
@@ -0,0 +1,25 @@
...
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.token;
+
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+
+/**
+ * @author Alexey Nesterov
+ */
+public interface TokenIssuer<PRINCIPAL> {
```
jgrandja, line 23: Please remove this as it's not needed as part of this issue. The `OAuth2ClientCredentialsAuthenticationProvider` should simply leverage `StringKeyGenerator accessTokenGenerator` the same as in `OAuth2AuthorizationCodeAuthenticationProvider`

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
```diff
@@ -0,0 +1,66 @@
...
+		if (!clientSecret.equals(registeredClient.getClientSecret())) {
+			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_CLIENT));
+		}
+
+		return new OAuth2AccessTokenAuthenticationToken(registeredClient, token, this.tokenIssuer.issue(registeredClient));
+	}
+
+	@Override
+	public boolean supports(Class<?> authentication) {
+		return authentication == OAuth2ClientCredentialsAuthenticationToken.class;
```
jgrandja, line 64: Change to `return OAuth2ClientCredentialsAuthenticationToken.class.isAssignableFrom(authentication)`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -86,8 +84,10 @@
 	private final AuthenticationManager authenticationManager;
 	private final OAuth2AuthorizationService authorizationService;
 	private final RequestMatcher tokenEndpointMatcher;
-	private final Converter<HttpServletRequest, Authentication> authorizationGrantAuthenticationConverter =
```
jgrandja, line 89: The way I envisioned `authorizationGrantAuthenticationConverter` is to be a delegation-based implementation that simply delegates to the specific authorization grant implementation. Spring Security uses this strategy in a number of places. Take a look at `DelegatingAccessDeniedHandler` and `DelegatingOAuth2UserService` for examples.

For now, both `AuthorizationCodeAuthenticationConverter` and `ClientCredentialsAuthenticationConverter` should be both `private static class` and then create a `delegate` `Converter` that simply delegates to the appropriate impl.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/AuthorizationCodeAuthenticationConverter.java
```diff
@@ -0,0 +1,85 @@
...
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeAuthenticationToken;
+import org.springframework.util.MultiValueMap;
+import org.springframework.util.StringUtils;
+
+class AuthorizationCodeAuthenticationConverter implements Converter<HttpServletRequest, Authentication> {
```
jgrandja, line 33: Please revert this change and keep `AuthorizationCodeAuthenticationConverter` as `private static class` in `OAuth2TokenEndpointFilter`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientCredentialsAuthenticationConverter.java
```diff
@@ -0,0 +1,68 @@
...
+		if (!StringUtils.hasText(grantType) ||
+				parameters.get(OAuth2ParameterNames.GRANT_TYPE).size() != 1) {
+			throwError(OAuth2ErrorCodes.INVALID_REQUEST, OAuth2ParameterNames.GRANT_TYPE);
+		}
+		if (!AuthorizationGrantType.CLIENT_CREDENTIALS.getValue().equals(grantType)) {
+			throwError(OAuth2ErrorCodes.UNSUPPORTED_GRANT_TYPE, OAuth2ParameterNames.GRANT_TYPE);
+		}
+
+		// client_id (REQUIRED)
+		String clientId = parameters.getFirst(OAuth2ParameterNames.CLIENT_ID);
```
jgrandja, line 49: `OAuth2ClientAuthenticationFilter` and `OAuth2ClientAuthenticationProvider` will handle authenticating the client. Both `client_id` and `client_secret` should be removed from the validation checks. Take a look at `AuthorizationCodeAuthenticationConverter` and how it returns `clientPrincipal = SecurityContextHolder.getContext().getAuthentication()` in the `Authentication` and apply the same.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientCredentialsAuthenticationConverter.java
```diff
@@ -0,0 +1,68 @@
...
+			throwError(OAuth2ErrorCodes.INVALID_REQUEST, OAuth2ParameterNames.CLIENT_ID);
+		}
+
+		// client_secret (REQUIRED)
+		String clientSecret = parameters.getFirst(OAuth2ParameterNames.CLIENT_SECRET);
+		if (!StringUtils.hasText(clientSecret)) {
+			throwError(OAuth2ErrorCodes.INVALID_REQUEST, OAuth2ParameterNames.CLIENT_SECRET);
+		}
+
+		return new OAuth2ClientCredentialsAuthenticationToken(clientId, clientSecret);
```
jgrandja, line 60: `scope` is an optional parameter but should also be returned in `OAuth2ClientCredentialsAuthenticationToken`

jgrandja:
@alek-sys 

Regarding

> Support basic auth for client id / secret

This is already implemented in `ClientSecretBasicAuthenticationConverter`

alek-sys:
Thanks for feedback @jgrandja! I'm now re-working this PR based on your comments.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientCredentialsAuthenticationConverter.java
alek-sys: Do you expect section [3.3. Access Token](https://tools.ietf.org/html/rfc6749#section-3.3) to be implemented as a part of this PR? E.g. returning `scope` in the response, and filtering requested scopes based on a list of scopes supported by a client?

jgrandja: Yes. The `OAuth2ClientCredentialsAuthenticationProvider` should receive the "request scopes" and the provider should ensure the "request scopes"  are either a subset (or all) of the "assigned scopes". If the client requests a scope that is not in the "assigned scopes" then it should fail. If no scopes are requested then we should auto-approve. However, I'm not sure auto-approving is a good idea but we can change that later if need be.

alek-sys:
Updated based on your comments.

Also, [the spec](https://tools.ietf.org/html/rfc6749#section-3.3) states:
> If the issued access token scope is different from the one requested by the client, the authorization server MUST include the "scope" response parameter to inform the client of the actual scope granted.

But currently Token Endpoint always returns `scope` in the response. Shoud l I fix it in this PR or create a separate story for this (unless there is one already)?

jgrandja:
The returned scope will come into play in #42 at a later point. No need to address it in this PR. But feel free to add notes to #42.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -214,4 +222,57 @@ public Authentication convert(HttpServletRequest request) {
...
+			// "scope" is optional, see https://tools.ietf.org/html/rfc6749#section-4.4.2 and https://tools.ietf.org/html/rfc6749#section-3.3
+			// if "scope" is not provided, client scopes are used
+			String scopeParameter = request.getParameter(SCOPE_PARAMETER_NAME);
+			if (StringUtils.isEmpty(scopeParameter)) {
+				return new OAuth2ClientCredentialsAuthenticationToken(clientAuthenticationToken, clientScopes);
+			}
+
+			Set<String> requestedScopes = null;
+			try {
+				requestedScopes = scopesParser.parse(scopeParameter);
```
jgrandja, line 265: Please move `scope` validation in `OAuth2ClientCredentialsAuthenticationProvider`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DelegateGrantTypeAuthenticationConverter.java
```diff
@@ -0,0 +1,59 @@
...
+
+		Converter<HttpServletRequest, Authentication> converter = this.converters.get(new AuthorizationGrantType(grantType));
+		if (converter == null) {
+			throwError(OAuth2ErrorCodes.UNSUPPORTED_GRANT_TYPE, "grant_type");
+		}
+
+		return converter.convert(request);
+	}
+
+	public DelegateGrantTypeAuthenticationConverter registerConverter(AuthorizationGrantType authorizationGrantType, Converter<HttpServletRequest, Authentication> authenticationConverter) {
```
jgrandja, line 49: Please remove this method and  provide constructor that accepts the `Map` making it immutable.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DelegateGrantTypeAuthenticationConverter.java
```diff
@@ -0,0 +1,59 @@
...
+
+import org.springframework.core.convert.converter.Converter;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+import org.springframework.util.StringUtils;
+
+public class DelegateGrantTypeAuthenticationConverter implements Converter<HttpServletRequest, Authentication> {
```
jgrandja, line 30: Rename to `DelegatingAuthorizationGrantAuthenticationConverter` and make `final`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DelegateGrantTypeAuthenticationConverter.java
```diff
@@ -0,0 +1,59 @@
...
+
+	private final HashMap<AuthorizationGrantType, Converter<HttpServletRequest, Authentication>> converters = new HashMap<>();
```
jgrandja, line 32: Reference `Map` instead

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DelegateGrantTypeAuthenticationConverter.java
```diff
@@ -0,0 +1,59 @@
...
+
+	@Override
+	public Authentication convert(HttpServletRequest request) {
+		String grantType = request.getParameter("grant_type");
```
jgrandja, line 36: This class should not throw any errors. It simply delegates to the correct `Converter` or returns `null` . Similar to how `DelegatingOAuth2UserService` is implemented.

jgrandja, line 36: Use constants in `OAuth2ParameterNames`

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
```diff
@@ -0,0 +1,74 @@
...
+ * An {@link AuthenticationProvider} that converts give {@link OAuth2ClientAuthenticationToken} to a {@link OAuth2AccessTokenAuthenticationToken}
+ * using the provided token generator.
+ *
+ * @author Alexey Nesterov
+ */
+public class OAuth2ClientCredentialsAuthenticationProvider implements AuthenticationProvider {
+
+	private final StringKeyGenerator accessTokenGenerator = new Base64StringKeyGenerator(Base64.getUrlEncoder());
+
+	public OAuth2ClientCredentialsAuthenticationProvider() {
```
jgrandja, line 42: Please remove default constructor since it's redundant

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
```diff
@@ -0,0 +1,57 @@
...
+		return "";
+	}
+
+	@Override
+	public Object getPrincipal() {
+		return this.clientPrincipal;
+	}
+
+	public Set<String> getScopes() {
+		return scopes;
```
jgrandja, line 55: Prefix `this.`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -214,4 +222,57 @@ public Authentication convert(HttpServletRequest request) {
...
+				return null;
+			}
+
+			if (!OAuth2ClientAuthenticationToken.class.isAssignableFrom(authentication.getClass())) {
+				throwError(OAuth2ErrorCodes.INVALID_CLIENT, "client_id");
+			}
+
+			OAuth2ClientAuthenticationToken clientAuthenticationToken = (OAuth2ClientAuthenticationToken) authentication;
+			if (!clientAuthenticationToken.isAuthenticated()) {
+				throwError(OAuth2ErrorCodes.INVALID_CLIENT, "client_id");
```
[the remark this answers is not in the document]
alek-sys: I see now. Converter belongs to `web` package so it shouldn't really cotain any auth-n logic. So I refactored it a bit - now converted only extracts scopes from the request plus gets current authentication from the context. The rest is done in the manager, including validating scopes, using default scopes and validating client authentication. It is much more clear now, thanks for feedback.

alek-sys:
Thanks @jgrandja, I've updated this PR. Now there is much better split between converted and provider.

jgrandja:
Great thanks @alek-sys. I will review this tomorrow.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -214,4 +222,57 @@ public Authentication convert(HttpServletRequest request) {
...
+	/**
+	 * Validates that client is authenticated by {@link OAuth2ClientAuthenticationFilter}
+	 *
+	 * @see OAuth2ClientAuthenticationToken
+	 * @see OAuth2ClientAuthenticationFilter
+	 * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.4">Section 4.4 Client Credentials Grant</a>
+	 */
+	private static class ClientCredentialsAuthenticationConverter implements Converter<HttpServletRequest, Authentication> {
+
+		private static final String SCOPE_PARAMETER_NAME = "scope";
```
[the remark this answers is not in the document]
jgrandja: Please remove `SCOPE_PARAMETER_NAME` and use `OAuth2ParameterNames.SCOPE` in `convert`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DelegateGrantTypeAuthenticationConverter.java
jgrandja: Looks like this was missed

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
jgrandja: Actually the more I thought about this one, I don't think we need to validate here. The `scope` validation should be done when the `RegisteredClient` is created - similar to `ClientRegistration` in `oauth2-client`. We know then that the configured scopes are valid and all we need to do is match the configured ones against the requested ones (as-is). So let's remove `OAuth2ScopeParser` and pass in the requested scopes as is, eg:

`new HashSet<>(Arrays.asList(StringUtils.delimitedListToStringArray(scope, " ")))`

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/DelegateGrantTypeAuthenticationConverter.java
alek-sys: Sorry for that, I missed this comment totally - for whatever reason it is not shown on `Conversations` tab, only on `Files`. So I get notifcations about your comments but have to search for them. New GitHub UI is not so polished I guess. This class is now updated.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
alek-sys: The validation itself (e.g. check that client has all requested scopes) is happening in the provider. The purpose of`OAuth2ScopeParser`  is only parsing `scope` string request param to a set according to RFC. It is slightly more complicated than `delimitedListToStringArray` so I decided to introduce more strict parser to cover special characters, double spaces, etc. I think it still has some value, unless you absolutely want to remove it. 

Maybe exception for cases when `scope` is not a valid string should be different then, just general BAD_REQUEST?

alek-sys:
@jgrandja updated based on your latest comments. Also, see my comment regarding `OAuth2ScopeParser` - I think it has some value as it validates `scope` parameter to be compliant with RFC.

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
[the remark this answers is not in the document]
jgrandja: Looks like this was missed

core/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
jgrandja: I still feel that `scope` validation should happen when a client is registered, eg. when a `RegisteredClient` is created. This ensures that the `scope`(s) the client is allowed to request `RegisteredClient.scopes` are valid from the start. Even if the requested `scope` has invalid characters, it still won't match against the registered (valid) scope. So I feel this extra validation at runtime is redundant so please remove `OAuth2ScopeParser`. FYI, this validation already exists in `RegisteredClient.validateScopes()`

jgrandja:
@alek-sys Can you please update the test methods to following the naming convention `**methodName**When**Condition**Then**Result**`. Take a look at `OAuth2TokenEndpointFilterTests` and `OAuth2AuthorizationCodeAuthenticationProviderTests` for examples.

Also, please add integration tests in `OAuth2ClientCredentailsGrantTests` (new). See `OAuth2AuthorizationCodeGrantTests` for example.

alek-sys:
@jgrandja done. Actually there was `OAuth2ClientCredentialsGrantTests` class with an integration test already. I updated it a bit to make it more like `OAuth2AuthorizationCodeGrantTests`. I also changed security configuration for it to be more realistic by overriding authentication entry point to return 401 when client is not authenticated and `client_credentials` grant is used.

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
```diff
@@ -0,0 +1,57 @@
...
+package org.springframework.security.oauth2.server.authorization.authentication;
+
+import java.util.Collections;
+import java.util.Set;
+
+import org.springframework.security.authentication.AbstractAuthenticationToken;
+import org.springframework.security.core.Authentication;
+import org.springframework.util.Assert;
+
+public class OAuth2ClientCredentialsAuthenticationToken extends AbstractAuthenticationToken {
```
jgrandja, line 26: - Add class-level javadoc with `@author` and `@since`
- Add `OAuth2ClientCredentialsAuthenticationTokenTests`

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
```diff
@@ -0,0 +1,57 @@
...
+
+	private final Authentication clientPrincipal;
```
jgrandja, line 28: Add `private static final long serialVersionUID = Version.SERIAL_VERSION_UID;`

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
```diff
@@ -0,0 +1,57 @@
...
+	private final Set<String> scopes;
+
+	public OAuth2ClientCredentialsAuthenticationToken(OAuth2ClientAuthenticationToken clientPrincipal, Set<String> scopes) {
```
jgrandja, line 31: Use `Authentication` type instead of `OAuth2ClientAuthenticationToken `

jgrandja, line 31: Update to `@Nullable Set<String> scopes` and remove the other constructor

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
```diff
@@ -0,0 +1,57 @@
...
+	public Object getCredentials() {
+		return "";
+	}
+
+	@Override
+	public Object getPrincipal() {
+		return this.clientPrincipal;
+	}
+
+	public Set<String> getScopes() {
```
jgrandja, line 54: Add `@Nullable`

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
```diff
@@ -0,0 +1,81 @@
...
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.core.OAuth2Error;
+import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
+
+/**
+ * An {@link AuthenticationProvider} that converts give {@link OAuth2ClientAuthenticationToken} to a {@link OAuth2AccessTokenAuthenticationToken}
+ * using the provided token generator.
+ *
+ * @author Alexey Nesterov
```
jgrandja, line 37: Add `@since`

core/src/test/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProviderTest.java
```diff
@@ -0,0 +1,113 @@
...
+
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
+import org.springframework.security.oauth2.server.authorization.client.TestRegisteredClients;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+public class OAuth2ClientCredentialsAuthenticationProviderTest {
```
jgrandja, line 32: Rename to `...Tests`

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientCredentialsGrantTests.java
```diff
@@ -0,0 +1,128 @@
...
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.reset;
+import static org.mockito.Mockito.when;
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.csrf;
+import static org.springframework.security.test.web.servlet.request.SecurityMockMvcRequestPostProcessors.httpBasic;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+public class OAuth2ClientCredentialsGrantTests {
```
jgrandja, line 48: Add `@author`

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientCredentialsGrantTests.java
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	@Before
+	public void setup() {
+		reset(registeredClientRepository);
+		reset(authorizationService);
+	}
+
+	@Test
+	public void requestWhenAuthorizationRequestAuthenticatedThenThenReturnTokenAndScope() throws Exception {
```
jgrandja, line 72: Change "AuthorizationRequest" to "TokenRequest"

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientCredentialsGrantTests.java
```diff
@@ -0,0 +1,128 @@
...
+					.with(csrf())
+					.param("grant_type", "client_credentials")
+					.param("scope", "email openid"))
+				.andExpect(status().isOk())
+				.andExpect(jsonPath("$.access_token").isNotEmpty())
+				.andExpect(jsonPath("$.scope").value("openid email"));
+	}
+
+	@Test
+	public void requestWhenAuthorizationRequestNotAuthenticatedThenUnauthorized() throws Exception {
```
jgrandja, line 89: Change "AuthorizationRequest" to "TokenRequest"

core/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientCredentialsGrantTests.java
```diff
@@ -0,0 +1,128 @@
...
+				.thenReturn(client);
+
+		this.mvc.perform(post(OAuth2TokenEndpointFilter.DEFAULT_TOKEN_ENDPOINT_URI)
+				.param("grant_type", "client_credentials")
+				.param("scope", "email openid"))
+				.andExpect(status().isUnauthorized());
+	}
+
+	@EnableWebSecurity
+	static class AuthorizationServerConfiguration extends WebSecurityConfigurerAdapter {
```
jgrandja, line 102: Please change this to:

```java
@EnableWebSecurity
@Import(OAuth2AuthorizationServerConfiguration.class)
static class AuthorizationServerConfiguration {
...
```

See `OAuth2AuthorizationCodeGrantTests`

Any changes needed to the default security should be made in `OAuth2AuthorizationServerSecurity`

core/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationToken.java
alek-sys: `scopes` is never null, do you mean `@NonNull`?

alek-sys:
All updated based on your review @jgrandja. I only didn't add `@Nullable` since scope field is never null. Also, I noticed there is no logging anywhere in the project. Is that intentional or should I add at least debug logging to some of the components?

jgrandja:
Thanks for adding this feature @alek-sys !

FYI, I followed up with a polish commit as some changes were needed and I wanted to get this merged.
Let me know if you have any questions regarding the changes in the polish commit.

PR #96, opened 2020-07-22
Introduce JwtEncoder with JWS implementation

jgrandja:
Issue gh-81

### DIFF
--- a/crypto/src/main/java/org/springframework/security/crypto/keys/ManagedKey.java
+++ b/crypto/src/main/java/org/springframework/security/crypto/keys/ManagedKey.java
@@ -1,0 +1,246 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.crypto.keys;
+
+import org.springframework.lang.Nullable;
+import org.springframework.security.core.SpringSecurityCoreVersion2;
+import org.springframework.util.Assert;
+
+import javax.crypto.SecretKey;
+import java.io.Serializable;
+import java.security.Key;
+import java.security.PrivateKey;
+import java.security.PublicKey;
+import java.time.Instant;
+import java.util.Objects;
+
+/**
+ * A {@code java.security.Key} that is managed by a {@link KeyManager}.
+ *
+ * @author Joe Grandja
+ * @since 0.0.1
+ * @see KeyManager
+ */
+public final class ManagedKey implements Serializable {
+	private static final long serialVersionUID = SpringSecurityCoreVersion2.SERIAL_VERSION_UID;
+	private Key key;
+	private PublicKey publicKey;
+	private String keyId;
+	private Instant activatedOn;
+	private Instant deactivatedOn;
+
+	private ManagedKey() {
+	}
+
+	/**
+	 * Returns {@code true} if this is a symmetric key, {@code false} otherwise.
+	 *
+	 * @return {@code true} if this is a symmetric key, {@code false} otherwise
+	 */
+	public boolean isSymmetric() {
+		return SecretKey.class.isAssignableFrom(this.key.getClass());
+	}
+
+	/**
+	 * Returns {@code true} if this is a asymmetric key, {@code false} otherwise.
+	 *
+	 * @return {@code true} if this is a asymmetric key, {@code false} otherwise
+	 */
+	public boolean isAsymmetric() {
+		return PrivateKey.class.isAssignableFrom(this.key.getClass());
+	}
+
+	/**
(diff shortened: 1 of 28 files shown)


jgrandja:
@krajcsovszkig-ms @anoopgarlapati @jzheaux Here is the initial draft of `JwtEncoder` with a `JWS` implementation.

You can run the integration test `OAuth2AuthorizationCodeGrantTests.requestWhenTokenRequestValidThenResponseIncludesCacheHeaders()` to test the `authorization_code` grant flow that returns a `JWS`.

Feedback would be great.

**NOTE:** Tests and javadoc have not been done yet.

crypto/src/main/java/org/springframework/security/crypto/keys/ManagedKey.java
```diff
@@ -0,0 +1,138 @@
...
+	public boolean isSymmetric() {
+		return SecretKey.class.isAssignableFrom(this.key.getClass());
+	}
+
+	public boolean isAsymmetric() {
+		return PrivateKey.class.isAssignableFrom(this.key.getClass());
+	}
+
+	@SuppressWarnings("unchecked")
+	public <T> T getKey() {
```
krajcsovszkig-ms, line 54: I'd prefer returning a `Key` here and let the caller cast it explicitly to what they think it is, or having separate `getPrivateKey()` and `getSymmetricKey()` methods that return null or throw an exception if the `key` field is not the correct type.

crypto/src/main/java/org/springframework/security/crypto/keys/StaticKeyGeneratingKeyManager.java
```diff
@@ -0,0 +1,99 @@
...
+import java.util.UUID;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+/**
+ *
+ * @author Joe Grandja
+ * @since 0.0.1
+ */
+public final class StaticKeyGeneratingKeyManager implements KeyManager {
```
krajcsovszkig-ms, line 39: I'd call this `DefaultKeyManager` and document that it generates HMAC and RSA keys. Or alternatively make the generated keys configurable.

jose/src/main/java/org/springframework/security/oauth2/jose/JoseHeader.java
```diff
@@ -0,0 +1,183 @@
...
+	public String getContentType() {
+		return getHeader(CTY);
+	}
+
+	public Map<String, Object> getHeaders() {
+		return this.headers;
+	}
+
+	@SuppressWarnings("unchecked")
+	public <T> T getHeader(String name) {
```
krajcsovszkig-ms, line 101: If this is public, I'd prefer returning an `Object` and having the caller cast it explicitly. Perhaps have a separate private one that auto-casts to make the specific getters neater.

jose/src/main/java/org/springframework/security/oauth2/jose/jws/NimbusJwsEncoder.java
```diff
@@ -0,0 +1,174 @@
...
+		}
+
+		headers = JoseHeader.from(headers)
+				.type(JOSEObjectType.JWT.getType())
+				.keyId(managedKey.getKeyId())
+				.build();
+		JWSHeader jwsHeader = convert(headers);
+
+		claims = JwtClaimsSet.from(claims)
+				.id(UUID.randomUUID().toString())	// TODO Is this unique enough?
```
krajcsovszkig-ms, line 113: It seems to be using `secureRandom()` so it's as good as it gets IMO.

krajcsovszkig-ms:
Hi @jgrandja ,

Do you have an approximate ETA on this?

Do you expect the interfaces to change much? If not, I might start building on them in a week or two.

Thanks!

jgrandja:
I'm planning on getting back to this PR this week @krajcsovszkig-ms and merging it later this week or early next week.

As an FYI, this won't make it into Spring Security 5.4 (Sep 2 release) but will be the first priority for 5.5, which will be released approx. 6 months after 5.4.

I'm not expecting major changes to the current API but likely there will be some minor changes.

I'm very interested to see how this works out on client side so please go ahead with your implementation and provide me feedback on how we can improve on the current API. Thanks!

crypto/src/main/java/org/springframework/security/crypto/keys/ManagedKey.java
jgrandja: This was updated to:

```java
	public <T extends Key> T getKey() {
		return (T) this.key;
	}
```

To enforce a `Key` type

crypto/src/main/java/org/springframework/security/crypto/keys/StaticKeyGeneratingKeyManager.java
jgrandja: This implementation is strictly for development/testing and I've javadoc'd it. This is good for now but it will likely go away later on when a new implementation is provided.

jose/src/main/java/org/springframework/security/oauth2/jose/JoseHeader.java
jgrandja: I think this is cleaner than casting:

`JwsAlgorithm jwsAlgorithm = joseHeader.getHeader(JoseHeaderNames.ALG)`

vs.

`JwsAlgorithm jwsAlgorithm = (JwsAlgorithm) joseHeader.getHeader(JoseHeaderNames.ALG)`

The caller will know the type.

jgrandja:
@krajcsovszkig-ms I just merged this so you're good to go on your end.

krajcsovszkig-ms:
@jgrandja 

Great, thanks!

Issue #55, opened 2020-04-21
Implement OpenID Provider Configuration endpoint


jgrandja:
This issue will deliver the [OpenID Provider Configuration](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig) endpoint.

Kehrlann:
I'm interested in taking this one ✋

jgrandja:
It's your @Kehrlann !

Issue #53, opened 2020-04-21
Implement OpenID Connect 1.0 Authorization Code Flow


jgrandja:
This issue will deliver [OpenID Connect 1.0 Authorization Code Flow](https://openid.net/specs/openid-connect-core-1_0.html#CodeFlowAuth).

rajeshrai248:
Hi, when do we expect this feature to complete?

jgrandja:
It's currently on hold (as indicated in the labels) so it hasn't been scheduled as of yet. However, we're aiming before the end of year.

PR #143, opened 2020-11-03
Implement OpenID Provider Configuration endpoint

Kehrlann:
- See https://openid.net/specs/openid-connect-discovery-1_0.html
  sections 3 and 4.
- We introduce here a "ProviderSettings" construct to configure
  the authorization server, starting with endpoint paths (e.g.
  token endpoint, jwk set endpont, ...)

This will deliver #55

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
@@ -34,11 +34,13 @@ import org.springframework.security.oauth2.server.authorization.authentication.O
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2RefreshTokenAuthenticationProvider;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2TokenRevocationAuthenticationProvider;
 import org.springframework.security.oauth2.server.authorization.client.RegisteredClientRepository;
+import org.springframework.security.oauth2.server.authorization.config.ProviderSettings;
 import org.springframework.security.oauth2.server.authorization.web.JwkSetEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2AuthorizationEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2ClientAuthenticationFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2TokenEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2TokenRevocationEndpointFilter;
+import org.springframework.security.oauth2.server.authorization.web.OidcProviderConfigurationEndpointFilter;
 import org.springframework.security.web.AuthenticationEntryPoint;
 import org.springframework.security.web.access.intercept.FilterSecurityInterceptor;
 import org.springframework.security.web.authentication.DelegatingAuthenticationEntryPoint;
@@ -52,6 +54,9 @@ import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
+import java.net.MalformedURLException;
+import java.net.URI;
+import java.net.URISyntaxException;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
@@ -85,6 +90,8 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 			OAuth2TokenRevocationEndpointFilter.DEFAULT_TOKEN_REVOCATION_ENDPOINT_URI, HttpMethod.POST.name());
 	private final RequestMatcher jwkSetEndpointMatcher = new AntPathRequestMatcher(
 			JwkSetEndpointFilter.DEFAULT_JWK_SET_ENDPOINT_URI, HttpMethod.GET.name());
+	private final RequestMatcher oidcProviderConfigurationEndpointMatcher = new AntPathRequestMatcher(
+			OidcProviderConfigurationEndpointFilter.DEFAULT_OIDC_PROVIDER_CONFIGURATION_ENDPOINT_URI, HttpMethod.GET.name());
 
 	/**
 	 * Sets the repository of registered clients.
@@ -123,17 +130,32 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 	}
 
 	/**
+	 * Sets the provider settings.
+	 *
+	 * @param providerSettings the provider settings
+	 * @return the {@link OAuth2AuthorizationServerConfigurer} for further configuration
+	 */
+	public OAuth2AuthorizationServerConfigurer<B> providerSettings(ProviderSettings providerSettings) {
+		Assert.notNull(providerSettings, "providerSettings cannot be null");
+		this.getBuilder().setSharedObject(ProviderSettings.class, providerSettings);
+		return this;
+	}
+
+	/**
 	 * Returns a {@code List} of {@link RequestMatcher}'s for the authorization server endpoints.
 	 *
 	 * @return a {@code List} of {@link RequestMatcher}'s for the authorization server endpoints
 	 */
 	public List<RequestMatcher> getEndpointMatchers() {
+		// TODO: use ProviderSettings instead
 		return Arrays.asList(this.authorizationEndpointMatcher, this.tokenEndpointMatcher,
-				this.tokenRevocationEndpointMatcher, this.jwkSetEndpointMatcher);
+				this.tokenRevocationEndpointMatcher, this.jwkSetEndpointMatcher, this.oidcProviderConfigurationEndpointMatcher);
 	}
 
 	@Override
 	public void init(B builder) {
+		ProviderSettings providerSettings = getProviderSettings(builder);
+		validateProviderSettings(providerSettings);
 		OAuth2ClientAuthenticationProvider clientAuthenticationProvider =
 				new OAuth2ClientAuthenticationProvider(
 						getRegisteredClientRepository(builder),
@@ -186,7 +208,14 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 
 	@Override
 	public void configure(B builder) {
-		JwkSetEndpointFilter jwkSetEndpointFilter = new JwkSetEndpointFilter(getKeySource(builder));
+		if (getProviderSettings(builder).issuer() != null) {
+			OidcProviderConfigurationEndpointFilter oidcProviderConfigurationEndpointFilter = new OidcProviderConfigurationEndpointFilter(getProviderSettings(builder));
+			builder.addFilterBefore(postProcess(oidcProviderConfigurationEndpointFilter), AbstractPreAuthenticatedProcessingFilter.class);
+		}
+
+		JwkSetEndpointFilter jwkSetEndpointFilter = new JwkSetEndpointFilter(
+				getKeySource(builder),
+				getProviderSettings(builder).jwkSetEndpoint());
 		builder.addFilterBefore(postProcess(jwkSetEndpointFilter), AbstractPreAuthenticatedProcessingFilter.class);
 
 		AuthenticationManager authenticationManager = builder.getSharedObject(AuthenticationManager.class);
@@ -200,18 +229,21 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 		OAuth2AuthorizationEndpointFilter authorizationEndpointFilter =
 				new OAuth2AuthorizationEndpointFilter(
 						getRegisteredClientRepository(builder),
-						getAuthorizationService(builder));
+						getAuthorizationService(builder),
+						getProviderSettings(builder).authorizationEndpoint());
 		builder.addFilterBefore(postProcess(authorizationEndpointFilter), AbstractPreAuthenticatedProcessingFilter.class);
 
 		OAuth2TokenEndpointFilter tokenEndpointFilter =
 				new OAuth2TokenEndpointFilter(
 						authenticationManager,
-						getAuthorizationService(builder));
+						getAuthorizationService(builder),
+						getProviderSettings(builder).tokenEndpoint());
 		builder.addFilterAfter(postProcess(tokenEndpointFilter), FilterSecurityInterceptor.class);
 
 		OAuth2TokenRevocationEndpointFilter tokenRevocationEndpointFilter =
 				new OAuth2TokenRevocationEndpointFilter(
-						authenticationManager);
+						authenticationManager,
+						getProviderSettings(builder).tokenRevocationEndpoint());
 		builder.addFilterAfter(postProcess(tokenRevocationEndpointFilter), OAuth2TokenEndpointFilter.class);
 	}
 
@@ -262,5 +294,38 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 
 	private static <B extends HttpSecurityBuilder<B>> CryptoKeySource getKeySourceBean(B builder) {
 		return builder.getSharedObject(ApplicationContext.class).getBean(CryptoKeySource.class);
+	}
+
+	private static <B extends HttpSecurityBuilder<B>> ProviderSettings getProviderSettings(B builder) {
+		ProviderSettings providerSettings = builder.getSharedObject(ProviderSettings.class);
+		if (providerSettings == null) {
+			providerSettings = getProviderSettingsBean(builder);
+			if (providerSettings == null) {
+				providerSettings = new ProviderSettings();
+			}
+			builder.setSharedObject(ProviderSettings.class, providerSettings);
+		}
+		return providerSettings;
+	}
+
+	private static <B extends HttpSecurityBuilder<B>> ProviderSettings getProviderSettingsBean(B builder) {
+		Map<String, ProviderSettings> providerSettingsMap = BeanFactoryUtils.beansOfTypeIncludingAncestors(
+				builder.getSharedObject(ApplicationContext.class), ProviderSettings.class);
+		if (providerSettingsMap.size() > 1) {
+			throw new NoUniqueBeanDefinitionException(ProviderSettings.class, providerSettingsMap.size(),
+					"Expected single matching bean of type '" + ProviderSettings.class.getName() + "' but found " +
+							providerSettingsMap.size() + ": " + StringUtils.collectionToCommaDelimitedString(providerSettingsMap.keySet()));
+		}
+		return (!providerSettingsMap.isEmpty() ? providerSettingsMap.values().iterator().next() : null);
+	}
+
+	private void validateProviderSettings(ProviderSettings providerSettings) {
+		if (providerSettings.issuer() != null) {
+			try {
+				new URI(providerSettings.issuer()).toURL();
+			} catch (MalformedURLException | URISyntaxException e) {
+				throw new IllegalArgumentException("issuer must be a valid URL");
+			}
+		}
 	}
 }
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/oidc/OidcProviderConfiguration.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/oidc/OidcProviderConfiguration.java
@@ -1,0 +1,331 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.core.oidc;
+
+import org.springframework.security.oauth2.server.authorization.Version;
+import org.springframework.util.Assert;
+
+import java.io.Serializable;
+import java.net.URI;
+import java.net.URL;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+
+/**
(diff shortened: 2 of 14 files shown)


Kehrlann:
I have introduced `org.springframework.security.oauth2.core.ClientAuthenticationMethod2` so that the names match "xxx_auth_methods_supported` in OpenID Discovery 1.0. 

As a side note, It does not derive from `org.springframework.security.oauth2.core.ClientAuthenticationMethod` in spring-security because that class is final...

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OidcProviderConfigurationFilter.java
```diff
@@ -0,0 +1,98 @@
...
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see ProviderSettings
+ * @see <a target="_blank" href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig">OpenID Connect Discovery 1.0</a>
+ */
+public class OidcProviderConfigurationFilter extends OncePerRequestFilter {
+	private final RequestMatcher requestMatcher;
+	private final OidcProviderConfiguration providerConfiguration;
+	private final OidcProviderConfigurationHttpMessageConverter messageConverter = new OidcProviderConfigurationHttpMessageConverter();
```
jgrandja, line 54: Rename to `providerConfigurationHttpMessageConverter`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OidcProviderConfigurationFilter.java
```diff
@@ -0,0 +1,98 @@
...
+	@Override
+	protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
+			throws ServletException, IOException {
+		if (!this.requestMatcher.matches(request)) {
+			filterChain.doFilter(request, response);
+			return;
+		}
+
+		ServletServerHttpResponse resp = new ServletServerHttpResponse(response);
+		messageConverter.write(this.providerConfiguration, MediaType.APPLICATION_JSON, resp);
```
jgrandja, line 85: Prefix `this`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/ProviderSettings.java
```diff
@@ -0,0 +1,178 @@
...
+/**
+ * A facility for OpenID Connect Provider Configuration settings.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see Settings
+ * @see <a target="_blank" href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata">OpenID Connect Discovery 1.0</a>
+ */
+public class ProviderSettings extends Settings {
+	private static final String PROVIDER_SETTING_BASE = "spring.security.oauth2.authorization-server.provider.";
```
jgrandja, line 35: Change prefix value to `setting.provider.`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/ProviderSettings.java
```diff
@@ -0,0 +1,178 @@
...
+	public static final String ISSUER = PROVIDER_SETTING_BASE.concat("issuer");
+	public static final String OPENID_CONFIGURATION_ENDPOINT = PROVIDER_SETTING_BASE.concat("openid-configuration-endpoint");
+	public static final String AUTHORIZATION_ENDPOINT = PROVIDER_SETTING_BASE.concat("authorization-endpoint");
+	public static final String TOKEN_ENDPOINT = PROVIDER_SETTING_BASE.concat("token-endpoint");
+	public static final String JWK_SET_ENDPOINT = PROVIDER_SETTING_BASE.concat("jwk-set-endpoint");
```
jgrandja, line 40: Change value to `jwks-endpoint`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/ProviderSettings.java
```diff
@@ -0,0 +1,178 @@
...
+	public URL issuer() {
+		return setting(ISSUER);
+	}
+
+	/**
+	 * Returns the OpenID Provider Configuration endpoint. The default is {@code /.well-known/openid-configuration}.
+	 *
+	 * @return the OpenID Provider Configuration endpoint
+	 */
+	public String openIDProviderConfigurationEndpoint() {
```
jgrandja, line 74: Let's remove this as the spec states it should be `/.well-known/openid-configuration`. I'm not sure if we need to allow this endpoint to be customized.

jgrandja (changes requested):
Thanks for the PR @Kehrlann. Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OidcProviderConfigurationFilter.java
```diff
@@ -0,0 +1,98 @@
...
+	private final OidcProviderConfiguration providerConfiguration;
+	private final OidcProviderConfigurationHttpMessageConverter messageConverter = new OidcProviderConfigurationHttpMessageConverter();
+
+	public OidcProviderConfigurationFilter(ProviderSettings providerSettings) {
+		Assert.notNull(providerSettings, "providerSettings cannot be null");
+		this.requestMatcher = new AntPathRequestMatcher(
+				providerSettings.openIDProviderConfigurationEndpoint(),
+				HttpMethod.GET.name()
+		);
+		this.providerConfiguration = new OidcProviderConfiguration.Builder()
```
jgrandja, line 62: We should build `OidcProviderConfiguration` per request given that `ProviderSettings` may change at runtime since it's mutable.

oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
```diff
@@ -249,4 +274,17 @@ public void configure(B builder) {
 	private static <B extends HttpSecurityBuilder<B>> KeyManager getKeyManagerBean(B builder) {
 		return builder.getSharedObject(ApplicationContext.class).getBean(KeyManager.class);
 	}
+
+	private static <B extends HttpSecurityBuilder<B>> ProviderSettings getProviderSettings(B builder) {
+		ProviderSettings providerSettings = builder.getSharedObject(ProviderSettings.class);
+		if (providerSettings == null) {
+			providerSettings = getProviderSettingsBean(builder);
```
jgrandja, line 281: `ProviderSettings` will not be a required `@Bean` since we can default the values - `getProviderSettingsBean()` will throw if it's not defined

oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
```diff
@@ -80,6 +82,8 @@
 			OAuth2TokenRevocationEndpointFilter.DEFAULT_TOKEN_REVOCATION_ENDPOINT_URI, HttpMethod.POST.name());
 	private final RequestMatcher jwkSetEndpointMatcher = new AntPathRequestMatcher(
 			JwkSetEndpointFilter.DEFAULT_JWK_SET_ENDPOINT_URI, HttpMethod.GET.name());
+	private final RequestMatcher oidcProviderMetadataEndpointMatcher = new AntPathRequestMatcher(
+			"/.well-known/openid-configuration", HttpMethod.GET.name());
```
jgrandja, line 86: Let's add "/.well-known/openid-configuration" as a constant in the `Filter`, same as all other `Filter`'s

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/oidc/OidcProviderMetadataClaimAccessor.java
```diff
@@ -0,0 +1,119 @@
...
+
+	/**
+	 * Returns the Subject Identifier types supported {@code (subject_types_supported)}.
+	 *
+	 * @return the Subject Identifier types supported
+	 */
+	default List<String> getSubjectTypes() {
+		return this.getClaimAsStringList(OidcProviderMetadataClaimNames.SUBJECT_TYPES_SUPPORTED);
+	}
+
```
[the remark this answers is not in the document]
Kehrlann: Good point. ID_TOKEN_SIGNING_ALG_VALUES_SUPPORTED is not used yet. Should we remove it from the claim names instead or provide access ?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/oidc/endpoint/OidcProviderConfiguration.java
```diff
@@ -0,0 +1,312 @@
...
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see OidcProviderMetadataClaimAccessor
+ * @see <a target="_blank" href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata">OpenID Connect Discovery 1.0</a>
+ */
+public class OidcProviderConfiguration implements OidcProviderMetadataClaimAccessor {
+	private final Map<String, Object> claims;
+
+	private OidcProviderConfiguration(Map<String, Object> claims) {
```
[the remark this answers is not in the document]
Kehrlann: Should we somehow validate the claims ? The initial pattern was to make it private and force the use of the builder, which validates the claims.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OidcProviderConfigurationFilter.java
Kehrlann: If the Issuer URL is invalid (e.g. with spaces inside), what's the type of the exception we want to throw ?

It's an edge case, we'll be wrapping `MalformedURLException`.

oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
Kehrlann: Done.

Most of the other filters endpoints will be configurable through ProviderSettings though. The defaults are also in the ProviderSettings (hardcoded as well). Is there something specific we want to do about this ?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/oidc/endpoint/OidcProviderConfiguration.java
```diff
@@ -0,0 +1,312 @@
...
+		}
+
+		/**
+		 * Constructs a new {@link Builder} from the provided claims.
+		 *
+		 * @param claims the claims
+		 * @see <a target="_blank" href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata">OpenID Connect Discovery 1.0</a>
+		 * for required claims
+		 */
+		public Builder(Map<String, Object> claims) {
```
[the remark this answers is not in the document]
Kehrlann: See comment about validating the claims in `OidcProviderConfiguration`. For now, only the Builder validates the claims, and this builder constructor is used when reading a response through `OidcProviderConfigurationHttpMessageConverter`.

oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
Kehrlann: I like the idea.

If we don't register the filter at all, and the user navigates to `/.well-known/openid-configuration` then they get an `HTTP 302` redirecting to `/login`. I'm feeling we should instead return an `HTTP 404`. Thoughts ?

Additionally, `ProviderSettings` is mutable and could change. If the issuer is added after startup, we wouldn't be registering the endpoint. Is that fine ?

I'd love a bit of guidance on writing some tests for that.

Kehrlann:
I've added integration tests to `OAuth2AuthorizationCodeGrantTests`, but I'm not sure it's the best place for it. Maybe it should go in it's own `OidcProviderConfiugrationTests`, or maybe a broader `OidcTests` for oidc-related workflows ?

joshuatcasey:
Hi from the Application SSO team. We had a team goal to implement a customizable Issuer URI - thanks for beating us to it!

We noticed a few things with `64207e9`.

https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata

> issuer
REQUIRED. URL using the https scheme with no query or fragment component that the OP asserts as its Issuer Identifier. If Issuer discovery is supported (see Section 2), this value MUST be identical to the issuer value returned by WebFinger. This also MUST be identical to the iss Claim value in ID Tokens issued from this Issuer.

https://tools.ietf.org/html/rfc8414#section-2

Kehrlann:
Hi @joshuatcasey ! Glad to help. To your points, here are some of the discussions we had around these topics:
1. **`issuer` not required:** The main idea for keeping the issuer optional (for now) is to stand up an authz server with absolute minimal configuration, even if it doesn't implement all possible extensions. I'll let @jgrandja chime in on this.
1. **using the `issuer` in tokens:** Yes the `iss` will be populated from the ProviderSettings, it is on the roadmap, not in the scope of this PR though.
1. **`issuer` validation:** That's correct, for local development `https` is a bit of a pain. We talked about pushing the validation of the URL to the users of the lib instead.

joshuatcasey:
Thanks for the clarifications.

1. Understood. It will likely be required for app SSO, but I understand if that's not part of the library.
1. Can you link this on the roadmap? I see that #148 was closed in favor of this PR. Happy to submit a PR for this soon.
1. Yeah, we had the same thought about local development or deployments where "TLS everywhere" isn't enforced. We'll probably add a flag to allow `http` as well as `https` and make it very obvious to the configurer that this is unsafe / outside of specification compliance.

Kehrlann:
I think @alek-sys will be making sure the `iss` claim is correctly populated, probably based off of #145

alek-sys:
> I think @alek-sys will be making sure the `iss` claim is correctly populated, probably based off of #145 (that PR brings some changes to the API as well)

#145 will abstract JWT token creation behind `TokenIssuer` interface, but current the default implementation for JWT access token doesn't have any validations, not even for `iss` claim. There used to be a mechanism to validate issued token, but I couldn't find any use of it so after chatting with Joe decided to delete it. Maybe `iss` token schema validation is a reason to bring it back.

joshuatcasey:
Speaking as an application developer, I want to validate the user's configuration on application startup, so it makes sense that we would validate `issuer URL` at that time. For now, we'll have some custom logic to perform this validation and we can discuss pushing it upstream to this library later.

oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2ClientCredentialsGrantTests.java
```diff
@@ -70,10 +74,11 @@
 	private MockMvc mvc;
 
 	@BeforeClass
-	public static void init() {
+	public static void init() throws MalformedURLException {
 		registeredClientRepository = mock(RegisteredClientRepository.class);
 		authorizationService = mock(OAuth2AuthorizationService.class);
 		keySource = new StaticKeyGeneratingCryptoKeySource();
+		providerSettings = new ProviderSettings().issuer(new URL("https://example.com"));
```
jgrandja, line 81: Please revert the changes as `providerSettings` is not needed in these tests

oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2RefreshTokenGrantTests.java
```diff
@@ -76,10 +80,11 @@
 	private MockMvc mvc;
 
 	@BeforeClass
-	public static void init() {
+	public static void init() throws MalformedURLException {
 		registeredClientRepository = mock(RegisteredClientRepository.class);
 		authorizationService = mock(OAuth2AuthorizationService.class);
 		keySource = new StaticKeyGeneratingCryptoKeySource();
+		providerSettings = new ProviderSettings().issuer(new URL("https://example.com"));
```
jgrandja, line 87: Please revert the changes as `providerSettings` is not needed in these tests

oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2TokenRevocationTests.java
```diff
@@ -75,10 +79,11 @@
 	private MockMvc mvc;
 
 	@BeforeClass
-	public static void init() {
+	public static void init() throws MalformedURLException {
 		registeredClientRepository = mock(RegisteredClientRepository.class);
 		authorizationService = mock(OAuth2AuthorizationService.class);
 		keySource = new StaticKeyGeneratingCryptoKeySource();
+		providerSettings = new ProviderSettings().issuer(new URL("https://example.com"));
```
jgrandja, line 86: Please revert the changes as `providerSettings` is not needed in these tests

samples/boot/oauth2-integration/authorizationserver/src/main/java/sample/config/AuthorizationServerConfig.java
```diff
@@ -64,6 +67,15 @@ public CryptoKeySource keySource() {
 		return new StaticKeyGeneratingCryptoKeySource();
 	}
 
+	@Bean
+	public ProviderSettings providerSettings() {
```
jgrandja, line 71: Please revert the changes as `ProviderSettings` is not required

jgrandja (changes requested):
Thanks for the updated @Kehrlann. Please see review comments.

jgrandja:
@joshuatcasey 

> I want to validate the user's configuration on application startup

Agreed. We will validate the application supplied configuration at startup and fail-fast on errors. I'm not sure where this will go at the moment but it very well might  go in `OAuth2AuthorizationServerConfigurer`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/oidc/OidcProviderConfiguration.java
```diff
@@ -0,0 +1,317 @@
...
+
+		private Builder() { }
+
+		/**
+		 * Use this {@code issuer} in the resulting {@link OidcProviderConfiguration}, REQUIRED.
+		 *
+		 * @param issuer the issuer URI
+		 * @return the {@link Builder} for further configuration
+		 */
+		public Builder issuer(URL issuer) {
```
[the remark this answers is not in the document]
Kehrlann: Good idea. Questions:
- We keep the claims as URLs, though, correct ?
- And therefore we should validate the URLs upon calling `.build()` ?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/ProviderSettings.java
```diff
@@ -0,0 +1,156 @@
...
+		return setting(ISSUER);
+	}
+
+	/**
+	 * Sets the URL the Provider uses as its Issuer Identity.
+	 *
+	 * @param issuer the URL the Provider uses as its Issuer Identity.
+	 * @return the {@link ProviderSettings} for further configuration
+	 */
+	public ProviderSettings issuer(URL issuer) {
```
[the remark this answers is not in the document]
Kehrlann: Should we run the validation in this PR ?
If so, where exactly ? In `OAuth2AuthorizationServerConfigurer#init` ?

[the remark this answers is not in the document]
jgrandja: Yes, let's validate in `OAuth2AuthorizationServerConfigurer#init`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/oidc/OidcProviderConfiguration.java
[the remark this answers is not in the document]
jgrandja: > We keep the claims as URLs, though, correct

Yes, getters stay as-is. Just change the setters in builder.

> And therefore we should validate the URLs upon calling .build()

I don't think we need to validate here since we're validating `ProviderSettings` in `OAuth2AuthorizationServerConfigurer`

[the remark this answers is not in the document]
Kehrlann: We are validating they're not null and that they're an instance of URL. If we put Strings we could validate they can be cast to URLs.

samples/boot/oauth2-integration/authorizationserver/src/main/java/sample/config/AuthorizationServerConfig.java
Kehrlann: Should we keep it anyway ? Otherwise we wouldn't have the OpenID Provider Configuration endpoint available (and later OAuth 2.0 Server Metadata endpoint).

We could add a comment explaining _why_ we have this Bean.

jgrandja: Let's remove it since we don't have OIDC auth implemented yet #53. I'll be starting #53 next week and I'll be sure to add this back to the sample as part of that PR.

Kehrlann:
Integrated comments.

Could you please take a look at `OAuth2AuthorizationServerConfigurer#validateProviderSettings` for validating the issuer URI ?
Also added `ObjectToSetStringConverter2` here until we add it to Spring Security ; could you please take a look at it before integrating it ? I used `ObjectToListStringConverter` as an inspiration, but I only added temporary tests for `ObjectToSetStringConverter2#convert`, I am a bit unsure about `#getConvertibleTypes` and `#matches`.

jgrandja (approved):

jgrandja:
Great work @Kehrlann ! This is now in master!

Issue #52, opened 2020-04-21
Implement Token Introspection Endpoint


jgrandja:
This will deliver [OAuth 2.0 Token Introspection](https://tools.ietf.org/html/rfc7662), which provides support for opaque access tokens.

lspil:
Hi @jgrandja Can I take this issue?

jgrandja:
Thanks for the offer @lspil. This feature is further down the backlog so we're holding off for now. I'm in the process of setting up the next set of work that will deliver the `authorization_code` grant. I should be organized with the plan by either this Friday or early next week, at which point there will be a few issues you can choose from. I'll be sure to reach out to you when things are ready to go. Thanks!

dfcoffin:
@jgrandja Please reach out to me as well, once the plan has been completed.

jgrandja:
You are definitely on my list @dfcoffin 👍

lspil:
Thanks @jgrandja !

rozagerardo:
Hi @jgrandja! I'm just starting to dedicate some time to this project now, and since I don't see many issues logged to work on, maybe I can help define the specs for the base ticket for this Epic (even maybe work on a draft implementation just to analyze this better, for when we're ready to tackle it)? Or are we planning to move forward with a different track of work?

Here it goes:

======

An Authorization Server should provide an endpoint to determine the active state of an OAuth 2.0 token and to determine meta-information about this token.

An `OAuth2TokenIntrospectionEndpointFilter` should be implemented as a `Filter`.

### Implementation requirements:
* The `Filter` should process `POST` requests for the (default) path `/oauth2/introspect`
* Request parameters are expected to be sent as _application/x-www-form-urlencoded_ data:
  * `token` (required): can be an Access Token or a Refresh Token
  * `token_type_hint` (optional): can be either `access_token` or `refresh_token` (defined in [OAuth 2.0 Token Revocation spec](https://tools.ietf.org/html/rfc7009))
* The client must be authenticated #39 before the Token Introspection Request can be accepted
* Conditions to consider the token 'active':
  * issued by this Authorization Server
  * not expired (and not before the `nbf` timestamp)
  * not revoked or invalidated
  * valid signature (if the token is a JWT)
  * valid resource server introspecting the token
* The metadata included in the response should contain:
  * `active` (required): boolean indicator of whether or not the presented token is currently active.
  * `scope` (optional): a list of space-delimited, case-sensitive strings
[...]

PR #145, opened 2020-11-06
Add Token Issuer interface

alek-sys:
First draft of `OAuth2TokenIssuer` interface


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenIssuer.java
```diff
@@ -0,0 +1,41 @@
...
+ * Base interface for token issuer implementations. Implementations can chose to return any type of an OAuth2 token,
+ * {@code TOKEN_TYPE} which implements {@code AbstractOAuth2Token}.
+ *
+ * @see JwtAccessTokenOAuth2TokenIssuer
+ * @see OpaqueRefreshTokenOAuth2TokenIssuer
+ *
+ * @author Alexey Nesterov
+ * @since 0.0.4
+ */
+public interface OAuth2TokenIssuer<TOKEN_TYPE extends AbstractOAuth2Token> {
```
jgrandja, line 31: Please make a copy of `OAuth2Token` in this repository, similar to this [commit](https://github.com/spring-projects-experimental/spring-authorization-server/commit/bf24cfb19e85ee2dd20701e20baed1bbed23fec0#diff-24bbe651f1e6b03c25ad8cf4853602cdf2422ceb94a2762659ee1b3d5889bdf4R22) and use this for the type instead of `AbstractOAuth2Token`. We cannot do GA releases that are based on milestone dependencies so I had to downgrade Spring Security to 5.4.+.

jgrandja, line 31: Spring Security doesn't use named generic types so please change `TOKEN_TYPE` to `T`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenResult.java
```diff
@@ -0,0 +1,69 @@
...
+/**
+ * Result returned by a {@link OAuth2TokenIssuer}, a combination of a token of {@code TOKEN_TYPE} type and {@link OAuth2TokenMetadata}.
+ *
+ * @see OAuth2TokenIssuer
+ * @see OAuth2TokenMetadata
+ *
+ * @author Alexey Nesterov
+ * @since 0.0.4
+ */
+public class OAuth2TokenResult<TOKEN_TYPE extends AbstractOAuth2Token> {
```
jgrandja, line 30: Rename generic `TOKEN_TYPE` to `T`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenResult.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.security.oauth2.core.AbstractOAuth2Token;
+
+/**
+ * Result returned by a {@link OAuth2TokenIssuer}, a combination of a token of {@code TOKEN_TYPE} type and {@link OAuth2TokenMetadata}.
+ *
+ * @see OAuth2TokenIssuer
+ * @see OAuth2TokenMetadata
+ *
+ * @author Alexey Nesterov
+ * @since 0.0.4
```
jgrandja, line 28: Change to `0.1.0`. Also, please ensure it's added in all other new classes as I noticed it was missing in a couple.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtOAuth2TokenCustomizer.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @see JwtAccessTokenOAuth2TokenIssuer
+ * @see OAuth2TokenIssuer
+ * @see org.springframework.security.oauth2.jwt.Jwt
+ *
+ * @author Alexey Nesterov
+ * @since 0.0.4
+ */
+@FunctionalInterface
+public interface JwtOAuth2TokenCustomizer {
```
jgrandja, line 33: I would prefer to introduce a general interface that would suit customizing the claims of a `Jwt` and opaque access token, e.g. `OAuth2TokenCustomizer`. We might need to introduce a hierarchy for customizing the `Jwt` header, something like `interface JwtTokenCustomizer extends OAuth2TokenCustomizer` and provides the additional operation for customizing the headers.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
```diff
@@ -0,0 +1,144 @@
...
+	 * @see OAuth2AccessToken
+	 * @see Jwt
+	 * @see <a href="https://tools.ietf.org/html/rfc7519#section-4">RFC7519 section 4</a>
+	 *
+	 * @param tokenRequest the request
+	 * @return a new OAuth2Token with metadata
+	 */
+	@Override
+	public OAuth2TokenResult<OAuth2AccessToken> issue(@NonNull OAuth2TokenRequest tokenRequest) {
+		Assert.notNull(tokenRequest.getPrincipalName(), "resourceOwner cannot be null");
```
jgrandja, line 94: These assertions belong in `OAuth2TokenRequest.Builder.build()`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
```diff
@@ -0,0 +1,144 @@
...
+		Assert.notNull(tokenRequest.getRegisteredClient(), "registeredClient cannot be null");
+		Assert.notNull(tokenRequest.getTimeToLive(), "timeToLive cannot be null");
+
+		Instant issuedAt = Instant.now(this.clock);
+		Instant expiresAt = issuedAt.plus(tokenRequest.getTimeToLive());
+
+		JoseHeader.Builder joseHeaderBuilder = JoseHeader.withAlgorithm(SignatureAlgorithm.RS256);
+		Set<String> scopes = tokenRequest.getScopes();
+		JwtClaimsSet.Builder jwtClaimsSetBuilder = JwtClaimsSet.withClaims()
+				.subject(tokenRequest.getPrincipalName())
```
jgrandja, line 104: The TODO for the `issuer` claim should be put back in here so we don' forget

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
```diff
@@ -0,0 +1,144 @@
...
+
+		JoseHeader joseHeader = joseHeaderBuilder.build();
+		JwtClaimsSet claimsSet = jwtClaimsSetBuilder.build();
+		this.validators.forEach(jwtOAuth2TokenValidator -> jwtOAuth2TokenValidator.validate(joseHeader, claimsSet, tokenRequest));
+
+		Jwt jwt = jwtEncoder.encode(joseHeaderBuilder.build(), jwtClaimsSetBuilder.build());
+		OAuth2AccessToken token = new OAuth2AccessToken(OAuth2AccessToken.TokenType.BEARER, jwt.getTokenValue(), issuedAt, expiresAt, scopes);
+		OAuth2TokenMetadata metadata = OAuth2TokenMetadata
+				.builder()
+				.metadata(OAuth2TokenMetadata.JWT, jwt)
```
jgrandja, line 121: Let's change `OAuth2TokenMetadata.JWT` to `CLAIMS` to keep things more generic. The type could simply be a `Map`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
```diff
@@ -0,0 +1,144 @@
...
+	}
+
+	private List<JwtOAuth2TokenValidator> getDefaultValidators() {
+		return Collections.singletonList(new TokenLifetimeValidator());
+	}
+
+	/**
+	 * Default token lifetime validator, ensures that token's expiresAt is set and after issuedAt
+	 */
+	private static class TokenLifetimeValidator implements JwtOAuth2TokenValidator {
```
jgrandja, line 135: Are you able to reuse `JwtTimestampValidator` and/or `JwtClaimValidator`?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
```diff
@@ -0,0 +1,144 @@
...
+	void setClock(Clock clock) {
+		this.clock = clock;
+	}
+
+	/**
+	 * Override default token validators with the provided collection. Validators are run after all customization applied.
+	 *
+	 * @param validators new validators collection
+	 */
+	public void setValidators(Iterable<JwtOAuth2TokenValidator> validators) {
```
jgrandja, line 68: We should try to reuse `OAuth2TokenValidator` instead of creating a new type.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OpaqueRefreshTokenOAuth2TokenIssuer.java
```diff
@@ -0,0 +1,65 @@
...
+/**
+ * An implementation of {@link OAuth2TokenIssuer} returning an opaque {@link OAuth2RefreshToken}.
+ *
+ * @see OAuth2RefreshToken
+ * @see OAuth2TokenIssuer
+ *
+ * @author Alexey Nesterov
+ * @since 0.0.4
+ */
+public class OpaqueRefreshTokenOAuth2TokenIssuer implements OAuth2TokenIssuer<OAuth2RefreshToken> {
```
jgrandja, line 38: Let's remove this. The only place we need to create an `OAuth2RefreshToken` is in `OAuth2AuthorizationCodeAuthenticationProvider` and `OAuth2RefreshTokenAuthenticationProvider`. The code is minimal so I'd rather not expose an API at this point. We could put the code in `OAuth2AuthenticationProviderUtils` to reuse if necessary.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
```diff
@@ -117,24 +124,36 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+			OAuth2TokenResult<OAuth2RefreshToken> refreshTokenResult = refreshTokenIssuer.issue(refreshTokenRequest);
+			refreshToken = refreshTokenResult.getToken();
+			tokensBuilder.refreshToken(refreshToken, refreshTokenResult.getMetadata());
 		}
 
 		OAuth2Tokens tokens = tokensBuilder.build();
 		authorization = OAuth2Authorization.from(authorization)
 				.tokens(tokens)
-				.attribute(OAuth2AuthorizationAttributeNames.ACCESS_TOKEN_ATTRIBUTES, jwt)
+				.attribute(OAuth2AuthorizationAttributeNames.ACCESS_TOKEN_METADATA, tokenResult.getMetadata())
```
jgrandja, line 156: The metadata is already associated in `OAuth2Tokens` so no need to store here.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenMetadata.java
```diff
@@ -33,13 +33,18 @@
...
+
 	private static final long serialVersionUID = Version.SERIAL_VERSION_UID;
 	protected static final String TOKEN_METADATA_BASE = "metadata.token.";
 
 	/**
 	 * The name of the metadata that indicates if the token has been invalidated.
 	 */
 	public static final String INVALIDATED = TOKEN_METADATA_BASE.concat("invalidated");
+	public static final String JWT = TOKEN_METADATA_BASE.concat("jwt");
+	public static final String AUTHORIZED_SCOPES = TOKEN_METADATA_BASE.concat("authorized-scopes");
```
jgrandja, line 45: `AUTHORIZED_SCOPES` should stay in `OAuth2Authorization.attributes`. That will likely change but it will remain in `OAuth2Authorization`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenMetadata.java
```diff
@@ -33,13 +33,18 @@
...
+
+	public static final OAuth2TokenMetadata EMPTY = OAuth2TokenMetadata.builder().build();
```
jgrandja, line 47: Please remove as each `OAuth2Token` should have it's own instance of `OAuth2TokenMetadata`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2AuthorizationAttributeNames.java
```diff
@@ -56,4 +56,9 @@
 	 */
 	String ACCESS_TOKEN_ATTRIBUTES = OAuth2Authorization.class.getName().concat(".ACCESS_TOKEN_ATTRIBUTES");
 
+	/**
+	 * The name of the attribute used for the metadata of the {@link OAuth2AccessToken}.
+	 */
+	String ACCESS_TOKEN_METADATA = OAuth2Authorization.class.getName().concat(".ACCESS_TOKEN_ATTRIBUTES");
```
jgrandja, line 62: Please remove as it's not needed. The `OAuth2TokenMetadata` is associated in the `OAuth2Tokens`

jgrandja (changes requested):
Thanks for the PR @alek-sys. Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenIssuer.java
alek-sys: I'll have to copy `OAuth2AccessToken` as well and make both refresh and access token implement this copied `OAuth2Token` interface. I guess it's fine for now, all of these classes will be removed after migration to 5.5.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
alek-sys: I made this switch but found that there isn't much to validate really - lifetime is already validated by `NumbusJwtEncoder`. I can't think of any other validations so I decided to remove it alltogether, let me me know if you want to bring it back without any defaults and purely as a customization point.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OpaqueRefreshTokenOAuth2TokenIssuer.java
alek-sys: The only reason I think this class is useful is that it can be overriden as a bean to customize `OAuth2AuthorizationCodeAuthenticationProvider` and `OAuth2RefreshTokenAuthenticationProvider` behaviour when generating refresh token. This seems to be a reasonable extension point, even though I'm not sure users will even change that. Are you sure you want to remove it?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
alek-sys: Changed to `TOKEN`. I'm not sure there is much value to store `OAuth2Token` there though, my idea was to store original object representing token, like `Jwt`, `String`, or class representing Paseo token in the future in the metadata. This has no immidate value but can be used later by Token Introspection endpoint. With that in mind, what about storing token like `.metadata(OAuth2TokenMetadata.TOKEN, jwt)`?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
alek-sys: Good point. Do you think anything should go to `OAuth2Authorization` attributes at all here? Token metadata now holds originally generated `Jwt` object, so there is no need to store it here.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
alek-sys: See below, I don't think these validators needed at all.

alek-sys:
Thanks for review @jgrandja. Most of your comments are now implemented. There are few things I'd like to discuss though, see unresolved comments above. Short summary is:
- Removing validators from `JwtOAuth2AccessTokenIssuer` as there is nothing really to validate
- Storing original `Jwt` object in the token metadata `TOKEN` property and removing it from `ACCESS_TOKEN_ATTRIBUTE` attribute on `OAuth2Authroziation` object
- Keeping `OpaqueRefreshTokenOAuth2TokenIssuer` purely as an extension point for `*Provider` classes
- Sticking to `AbstractOAuth2Token` for now rather than copying `OAuth2Token` from 5.5

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
alek-sys: After having a chat with Joe, we decided to keep the original `jwt` token in `TOKEN` attribute.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
alek-sys: After having a chat with Joe, we decided to remove validators for now.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
alek-sys: Going to remove this attribute completely since the the original token is now in `OAuth2TokenMetadata`.

alek-sys:
Everything is updated based on your feedback and ready for another review @jgrandja.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtAccessTokenOAuth2TokenIssuer.java
jgrandja: I meant to put back the original comment and the hardcoded `iss` claim:

> String issuer = "https://oauth2.provider.com";        // TODO Allow configuration for issuer claim

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2AuthorizationGrantContext.java
```diff
@@ -0,0 +1,122 @@
...
+ * the context.
+ *
+ * @author Alexey Nesterov
+ * @since 0.1.0
+ */
+public class OAuth2AuthorizationGrantContext {
+
+	private final String principalName;
+	private final RegisteredClient registeredClient;
+	private final Set<String> scopes;
```
jgrandja, line 38: `scopes` is optional as per spec so we should change this field to be more generic, e.g. `claims`? During user consent, the user may authorize more than just `scopes`, they may authorize specific claims, e.g. `email`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenCustomizer.java
```diff
@@ -0,0 +1,37 @@
...
+@FunctionalInterface
+public interface OAuth2TokenCustomizer {
+
+	/**
+	 * Apply token customization before it is issued by the issuer.
+	 *
+	 * @param claims original claims created by the issuer
+	 * @param authorizationGrantContext original authorization grant context
+	 */
+	void customizeClaims(Map<String, Object> claims, OAuth2AuthorizationGrantContext authorizationGrantContext);
```
jgrandja, line 35: I'm wondering if we can simplify this and remove the need for `JwtOAuth2TokenCustomizer`. Maybe we introduce a `OAuth2TokenBuilder` that contains all the data needed. The interface would then be `void customize (OAuth2TokenBuilder tokenBuilder)`. It would be nice to also remove the need for `OAuth2AuthorizationGrantContext` and instead put `RegisteredClient` and `principalName` in a  `Map` of `attributes` in `OAuth2TokenBuilder`. See what you can come up with based on this proposed design.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenMetadata.java
```diff
@@ -33,13 +33,15 @@
...
 public class OAuth2TokenMetadata implements Serializable {
+
 	private static final long serialVersionUID = Version.SERIAL_VERSION_UID;
 	protected static final String TOKEN_METADATA_BASE = "metadata.token.";
 
 	/**
 	 * The name of the metadata that indicates if the token has been invalidated.
 	 */
 	public static final String INVALIDATED = TOKEN_METADATA_BASE.concat("invalidated");
+	public static final String TOKEN = TOKEN_METADATA_BASE.concat("token");
```
jgrandja, line 44: Add javadoc

jgrandja (changes requested):
Thanks for the updates @alek-sys. Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenCustomizer.java
alek-sys: The main purpose was to abstract out notion of headers, which is very specific to JWT and irrelevant for other token types. My concern is that having `OAuth2TokenBuilder` won't solve that, it still will have to include `headers` property. I'll experiment with it and see how it comes together.

jgrandja: There are use cases where the headers will need to be customizable, e.g. applications may want to set the `jku` or `x5u` header instead of using `kid`. We'll need to provide this capability to customize headers.

Maybe this would work -> `OAuth2TokenCustomizer.customize(OAuth2TokenBuilder<T extends AbstractOAuth2Token> builder)`. The `Jwt` specific `OAuth2TokenBuilder` would provide a `headers(Consumer<Map<String, Object>> headersConsumer)` or maybe a `headers(Consumer<JoseHeader.Builder> joseHeaderConsumer)`?

alek-sys: Ok, thanks for clarifying on headers customization. So I updated the PR with another iteration, trying your idea of introducing a builder. I think it worked out quite nicely. However instead of binding a customizer to a token type by having `T extends AbstractOAuth2Token` I decided to link it to token identity, e.g. Jwt. So a customizers can be used for any token type (Id, Access) if that's represented by a Jwt token.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtOAuth2TokenBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+	 */
+	public static JwtOAuth2TokenBuilder withEncoder(JwtEncoder jwtEncoder) {
+		return new JwtOAuth2TokenBuilder(jwtEncoder);
+	}
+
+	/**
+	 * Claims set customizer. The given consumer will be invoked with pre-configured set of claims.
+	 * @param  claimsCustomizer a {@code JwtClaimsSet} customizer
+	 */
+	public void claims(Consumer<Map<String, Object>> claimsCustomizer) {
```
jgrandja, line 62: Let's change to `Consumer<JwtClaimsSet.Builder>`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtOAuth2TokenBuilder.java
```diff
@@ -0,0 +1,78 @@
...
+		this.jwtClaimsSetBuilder.claims(claimsCustomizer);
+	}
+
+	/**
+	 * Headers customizer. The given consumer will be invoked with pre-configured headers.
+	 * @param headersCustomizer a {@code JoseHeader} customizer
+	 */
+	public void headers(Consumer<Map<String, Object>> headersCustomizer) {
```
jgrandja, line 70: Let's change to `Consumer<JoseHeader.Builder>`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2AuthorizationGrantContext.java
```diff
@@ -0,0 +1,132 @@
...
+/**
+ * Context describing current status of OAuth2 authorization. User by a {@link OAuth2TokenIssuer} to issue a new OAuth2 token.
+ *
+ * Different implementations of {@link OAuth2TokenIssuer} may decide to use different properties of
+ * the context.
+ *
+ * @author Alexey Nesterov
+ * @since 0.1.0
+ */
+public class OAuth2AuthorizationGrantContext {
```
jgrandja, line 34: I'm wondering if we should rename this to `OAuth2TokenContext`? It might be more intuitive given the other names: `OAuth2TokenIssuer`, `OAuth2TokenResult`, `OAuth2TokenMetadata` and `AbstractOAuth2Token` (`OAuth2Token`)

jgrandja (changes requested):
Thanks for the updates @alek-sys. I left a couple more comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2AuthorizationGrantContext.java
alek-sys: What about renaming it back to `OAuth2TokenRequest`? I think of a context as something I can put data into and get data from. In this case this just a read-only object that a provider uses to pass data to token issuer. And the flow is quite logical then `OAuth2TokenRequest -> OAuth2TokenIssuer + OAuth2TokenCustomizer -> OAuth2TokenResult`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtOAuth2TokenBuilder.java
alek-sys: That's how I did it initially. But to be fair, now I like map as an argument more for a couple of reasons. Firstly, the given consumer goes directly into `JwtClaimsSet.Builder.claims` method, and there is no need to do something like `tokenBuilder.claims(builder -> builder.claims(claims -> claims.get("claim-name")))` to access existing claims. Secondly, map is a abstract enough and even if underlying `JwtClaimsSet` implementation changes, there'll be no need for users to change their code. `JwtBuilder` can even decide to switch implementation from `JwtClaimsSet` completely without affecting user's code.

If it's not convincing, I'm happy to change it to `JwtClaimsSet.Builder` but in this case there is no need in the consumer, `JwtTokenBuilder` can simply return claims set builder as `JwtClaimsSet.Builder getClaimsBuilder() { return this.jwtClaimsSetBuilder; }`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2TokenBuilder.java
```diff
@@ -0,0 +1,30 @@
...
+/**
+ * Basic interface for a token builder to share context between {@link OAuth2TokenIssuer} and {@link OAuth2TokenCustomizer}.
+ * A specific implementation can include relevant properties for a token type, e.g. headers and claims for a JWT token.
+ *
+ * @see JwtOAuth2TokenBuilder
+ *
+ * @author Alexey Nesterov
+ * @since 0.1.0
+ */
+public interface OAuth2TokenBuilder<T> {
```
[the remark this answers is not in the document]
alek-sys: `OAuth2TokenBuilder` is not building instances of `AbstractOAuth2Token` but instead instances of `Jwt` object. This object can then be converted into `OAuth2AccessToken` (or `OAuth2IdToken`) by the issuer. 

And I'm struggling to find a bound here, it is basically anything that represents a token - `Jwt` now, PASETO in the future, could even be a string, etc. Maybe this interface shouldn't exist, and issuer should declare what type of builder it expects? So `JwtBuilder` won't implement any interfaces but I think that's fine.

jgrandja:
@alek-sys After reviewing the PR in greater detail, I decided to hold off on introducing `OAuth2TokenIssuer` at this point.

The main reasons are:

1) The only implementation of `OAuth2TokenIssuer` is `JwtAccessTokenOAuth2TokenIssuer`, which doesn't look much different than what we currently have in `OAuth2TokenIssuerUtil`.
2) There is no `OAuth2TokenIssuer` implementation for refresh tokens - it's using `OAuth2AuthenticationProviderUtils`. So we don't know how this will integrate with refresh tokens.

It also feels like it might be premature to introduce `OAuth2TokenIssuer` before we implement OpenID Connect ID Token as well as opaque token support and the Introspection Endpoint.

Regarding the `OAuth2TokenCustomizer`, I originally suggested that it would be convenient to have a general abstraction for a token customizer. However, I'm not convinced that is the right path anymore.

For `Jwt` customization, it feels intuitive to have `customize(JoseHeader.Builder headers, JwtClaimsSet.Builder claims)`.

It also feels natural to associate a "token customizer" with a "token generator" (`JwtEncoder`).

Hence, I added `NimbusJwsEncoder.setJwtCustomizer(BiConsumer<JoseHeader.Builder, JwtClaimsSet.Builder> jwtCustomizer)` to allow for `Jwt` customization in  79f1cf5

I'm going to put this PR on hold until #53 and #52 is completed. At that point, we can circle back to `OAuth2TokenIssuer` and see if it makes sense to integrate.

Let me know if you have any questions.

Thanks again for your work on this @alek-sys.

And I apologize for initiating this enhancement pre-maturely.

alek-sys:
> After reviewing the PR in greater detail, I decided to hold off on introducing `OAuth2TokenIssuer` at this point.
> 

Thanks for feedback @jgrandja! No worries, I think the reasoning is very good, this abstraction at the moment looks a bit unnecessary, maybe that's why it was really hard to get it right. 

I like the idea of re-iterating on some sort of `TokenIssuer` interface later, when more consumers of this interface exist. Thanks for all your comments on this PR, I appreciate it. I'm going to close it for now, to avoid having umerged PR.

Issue #101, opened 2020-08-21
Add support for OAuth 2.0 Mutual-TLS Client Authentication and Certificate-Bound Access Tokens


tlodderstedt:
**Expected Behavior**
Support for RFC 8705 would allows clients to authenticate using TLS certificates and to bind access tokens to such certificates.

**Current Behavior**
Addition of client authentication with public client crypto increases the security capabilities of this project.

**Context**
There are several advantages:
* Certificate-based authentication is more secure than shared secrets since the secret is managed at the client only. This reduces the attack surface at the AS.
* Binding access tokens to certificates allows replay detection at the RS (recommended by the OAuth Security BCP). RFC 8705 provides a simple technical solution.
* Since RFC 8705 also supports self-signed certs, this mechanisms does not suffer from the typical headache a PKI has built in.
* Certificate-based client authentication is a great basis for using OAuth in micro service architecture as it allows 

Related gh-1558, gh-1559, gh-1560

pkostrzewa:
Has it already been assigned to someone? I'd like to give it a try.

jgrandja:
Thanks for the offer @pkostrzewa. I likely need to split this ticket up into 2 separate tasks. I'll get back to you sometime next week with a plan on how to implement this feature. Thanks!

jgrandja:
@pkostrzewa Apologies for the long delay in my response.

I'm going to be taking on this feature as part of a POC that I'll be starting next week. The POC involves integration with [SPIFFE / SPIRE](https://spiffe.io/).

If there are any other tasks you're interested in please let me know.

pkostrzewa:
@jgrandja Yes, I can take something else.

jgrandja:
See sample implementation:
https://github.com/jgrandja/spring-authorization-server/commits/springone-2021

aphalke:
Hi @jgrandja, in which release we can expect this feature?

jgrandja:
@aphalke We'll target this feature in the `1.1.0` release, which is approx. 7 months out.

jgrandja:
This feature has been pushed to the `1.2.0` release, as we don't have enough time to get it into `1.1.0`. It will be top priority for `1.2.0`.

ZuoJunLN:
@jgrandja Do you know which version of spring authorization server will support mtls?

jgrandja:
[...]

Issue #59, opened 2020-04-21
Support JWT Bearer Client Authentication


jgrandja:
This feature will deliver [Using JWTs for Client Authentication](https://datatracker.ietf.org/doc/html/rfc7523#section-2.2) defined in the JSON Web Token (JWT) Profile spec.

rlewczuk:
I've implemented client authentication part of this specification. If there are no objections, I'm ready to issue PR as soon as 0.1.2 cycle starts. 

This implementation contains significant amount of duplicated code (PKCE related) that needs to be factored out, I'm aware of that and plan to take care of it.

Branch implementing this feature is here: https://github.com/rlewczuk/spring-authorization-server/tree/jwt-client-auth

rlewczuk:
Here is another part of this epic - JWT assertion grant type: https://github.com/rlewczuk/spring-authorization-server/tree/jwt-assertion-grant

In both parts I had to make some design decisions that really need to be reviewed and discussed. Plus I have to review it once again for  compliance with RFC (especially error codes). If you have time some time next week, we can start working on this. Granted that design issues will be reviewed and fixed, its more-or-less complete and will be ready to land in master branch early in 0.1.2 (or even a bit earlier if we feel safe about it).

jgrandja:
Thanks @rlewczuk !

Let's focus on JWT client authentication for `0.1.2`. I'm not sure when JWT authorization grant will get scheduled at this point.

I'm pretty backlogged over the next 2 weeks. After `0.1.1` is out, please submit PR for JWT client authentication. Thanks!

larsw:
What's the status for this feature?

jgrandja:
@larsw This is scheduled for `0.2.1` as indicated in the Milestone.

PR #161, opened 2020-11-17
Implement Token Introspection Endpoint

rozagerardo:
Issue https://github.com/spring-projects-experimental/spring-authorization-server/issues/52
Added support to the Token Introspection Endpoint as per the description in the ticket above.


oauth2-authorization-server/spring-security-oauth2-authorization-server.gradle
```diff
@@ -7,6 +7,7 @@ dependencies {
 	compile 'org.springframework.security:spring-security-oauth2-jose'
 	compile springCoreDependency
 	compile 'com.nimbusds:nimbus-jose-jwt'
+	compile 'com.nimbusds:oauth2-oidc-sdk'
```
rozagerardo, line 10: This was added mainly to use [the _TokenIntrospectionSuccessResponse_ class](https://www.javadoc.io/doc/com.nimbusds/oauth2-oidc-sdk/5.0/com/nimbusds/oauth2/sdk/TokenIntrospectionSuccessResponse.html), but it might simply not be worth it. 
Of course, we might need it in the future (If I'm not mistaken the Spring Security Resource Server functionality uses this library as well), but note that it also forces us to handle some Nimbus specific classes (nothing too complicated though) that could be avoided by using our own response implementation.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,450 @@
...
+
+			oauthToken = parseJwt(oauthToken);
+
+			validateToken(oauthToken);
+
+			TokenIntrospectionSuccessResponse.Builder builder = new TokenIntrospectionSuccessResponse.Builder(true);
+			builder.clientID(new ClientID(clientAuthentication.getRegisteredClient().getClientId()));
+			Optional.ofNullable(oauthToken.getIssuedAt()).map(Date::from).ifPresent(builder::issueTime);
+			Optional.ofNullable(oauthToken.getExpiresAt()).map(Date::from).ifPresent(builder::expirationTime);
+			TokenToIntrospectionResponseFieldsMapper.extractFromToken(oauthToken, builder);
```
rozagerardo, line 216: I have added functionality to retrieve all the response fields, except for [the _username_](https://tools.ietf.org/html/rfc7662#section-2.2) ( _Human-readable identifier for the resource owner who authorized this token._).
For when we have a _Jwt_, we could use the "sub" field, but for the rest of the tokens, I think we're not storing that information anywhere.
Let me know what you think

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/TokenType.java
```diff
@@ -15,17 +15,18 @@
...
+import org.springframework.util.Assert;
+
 /**
  * @author Joe Grandja
  */
 public final class TokenType implements Serializable {
 	private static final long serialVersionUID = Version.SERIAL_VERSION_UID;
 	public static final TokenType ACCESS_TOKEN = new TokenType("access_token");
 	public static final TokenType REFRESH_TOKEN = new TokenType("refresh_token");
+	public static final TokenType ID_TOKEN = new TokenType("id_token");
```
rozagerardo, line 29: Maybe we want to create a task to add some additional tests after we add proper OIDC and _id_token_ support

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,450 @@
...
+		ServletServerHttpResponse httpResponse = new ServletServerHttpResponse(response);
+		this.tokenIntrospectionHttpResponseConverter.write(tokenIntrospectionResponse, null, httpResponse);
+	}
+
+	private AbstractOAuth2Token parseJwt(AbstractOAuth2Token token) {
+		for (JwtDecoder jwtDecoder : this.jwtDecoders) {
+			try {
+				return jwtDecoder.decode(token.getTokenValue());
+			} catch (JwtException ex) {
+				// token might not be a JWT, or can't be processed with this decoder.
```
rozagerardo, line 257: Maybe we should go further here, and throw an _InvalidTokenException_  (which simply translates into an _{"active": false}_ response) if we detect the Token has been signed, but the signature is not valid? here we would treat it as an opaque token...
On the other hand, I guess if a JWT was tampered, then it won't work as an opaque token either, as it shouldn't be found within our stored tokens anyway. Let me know what you think.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,450 @@
...
+				throw new InvalidTokenException("Token does not correspond to authenticated client");
+			}
+
+			// we obtained the authorization from the token, no need to validate Optional here
+			AbstractOAuth2Token oauthToken = authorization.getTokens().getToken(token).get();
+			if (authorization.getTokens().getTokenMetadata(oauthToken).isInvalidated()) {
+				throw new InvalidTokenException("Token has been invalidated");
+			}
+
+			oauthToken = parseJwt(oauthToken);
```
rozagerardo, line 208: I first thought of validating both the _AbstractOAuth2Token_ implementation (which could be an _OAuth2AccessToken_, an _OAuth2RefreshToken_, or an _OidcIdToken_) and the _Jwt_ separately, but according to the following specs:
https://tools.ietf.org/html/draft-ietf-oauth-access-token-jwt-10
https://openid.net/specs/openid-connect-core-1_0.html#IDToken
the jwt-formatted token counterparts **will** contain ( according to their required fields) all the base token fields (mainly _iat_, _exp_) so I decided to simply validate it as a _Jwt_ if possible, and avoid validating things twice. I couldn't find a spec for the JWT profile for a _refresh_token_.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,450 @@
...
+				: (OAuth2TokenValidator<T>) new TokenExpirationValidator();
+		OAuth2TokenValidatorResult result = tokenValidator.validate(token);
+		if (result.hasErrors()) {
+			String errorMessages = result.getErrors().stream().map(OAuth2Error::getErrorCode)
+					.collect(Collectors.joining(", ", "Invalid Token:", "."));
+			throw new InvalidTokenException(errorMessages);
+		}
+	}
+
+	private static boolean isPrincipalAuthenticated(Authentication principal) {
```
rozagerardo, line 276: Note I copied this method from the _OAuth2AuthorizationEndpointFilter_ class, maybe we should move it to a utils class?

rozagerardo:
Hey @jgrandja I prepared this PR for the Token Introspection Endpoint.
I've made a few comments in the PR to discuss a few aspects, but first I want to point out I have quite a few formatting changes even though I'm sticking to the `.editorconfig` file we have in the root of the project. Should I revert all these changes to make the PR more readable?

jgrandja:
Thanks for the PR @rozagerardo!

I'm a bit backlogged at the moment so I'll try to review this mid to end next week.

I did take a quick glance and noticed quite a bit of formatting changes. Can you please revert all formatting changes that are not directly related to the changes required for this PR. Thanks!

rozagerardo:
Hey @jgrandja sure, that was in fact my question just above: 
https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#issuecomment-729284018
Sounds good, I'll fix that in the next few days and resolve all the conflicts that arise too, thanks or the heads up!

jgrandja:
@rozagerardo Thanks for reverting the formatting changes.

I took a look at `OAuth2TokenIntrospectionEndpointFilter` and it doesn't follow the general pattern within Spring Security's `Filter` and `AuthenticationManager` / `AuthenticationProvider` architecture.

As a next step, I would encourage you to review the reference documentation to gain a deeper understanding:

- [Servlet Security: The Big Picture](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#servlet-architecture)
- [Authentication](https://docs.spring.io/spring-security/site/docs/current/reference/html5/#servlet-authentication)

After you review that, please review the [Token Revocation](https://github.com/spring-projects-experimental/spring-authorization-server/blob/061d9f8c1808ef5525de2550636201e6d78c8b90/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenRevocationEndpointFilter.java#L56) implementation in detail as it will follow the general pattern as Token Introspection.

This PR contains `OAuth2TokenIntrospectionEndpointFilter` but is missing `OAuth2TokenIntrospectionAuthenticationToken` and `OAuth2TokenIntrospectionAuthenticationProvider`.

Let me know if you have any questions.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
```diff
@@ -64,7 +69,7 @@ public OAuth2Authorization findByToken(String token, @Nullable TokenType tokenTy
 	}
 
 	private boolean hasToken(OAuth2Authorization authorization, String token, TokenType tokenType) {
-		if (OAuth2AuthorizationAttributeNames.STATE.equals(tokenType.getValue())) {
+		if (tokenType != null && OAuth2AuthorizationAttributeNames.STATE.equals(tokenType.getValue())) {
```
rozagerardo, line 72: I know we have an issue raised for this potential NPE (https://github.com/spring-projects-experimental/spring-authorization-server/issues/170), but this was generating issues for some Introspection scenarios as well, si I worked on a quick patch.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
```diff
@@ -108,4 +113,26 @@ public int hashCode() {
 			return Objects.hash(this.registeredClientId, this.principalName);
 		}
 	}
+
+	@Override
+	public Optional<OAuth2Authorization> findByTokenWithHint(String token,
```
rozagerardo, line 118: One quick note here. I realize the existing _findToken_ method will retrieve null if the token doesn't match the provided _tokenType_. However, here we want to make the search extensive to any supported token type, as per the specs:

> token_type_hint
>       OPTIONAL.  A hint about the type of the token submitted for
>       introspection.  The protected resource MAY pass this parameter to
>       help the authorization server optimize the token lookup.  **If the
>       server is unable to locate the token using the given hint, it MUST
>       extend its search across all of its supported token types.**  An
>       authorization server MAY ignore this parameter, particularly if it
>       is able to detect the token type automatically.  Values for this
>       field are defined in the "OAuth Token Type Hints" registry defined
>       in OAuth Token Revocation [RFC7009].

	  
I realized the Token Revocation code also uses this logic, but the specs indicate the same as above, so IMO that should be changed to use this method instead. If you agree, I'll create an issue for that :)
Finally, I know https://github.com/spring-projects-experimental/spring-authorization-server/pull/162 is already closed and I'm ok with that decision, but I caught this issue using those tests locally (in our current Integration tests we're mocking up these interactions), just to point out that they might be suitable for some edge cases :)

rozagerardo:
Hey @jgrandja sorry for the delay here, the last couple of weeks have been quite busy for me.

Also, thanks for the heads up with the last comment, naturally I had read these sections before on a couple of occasions, but it's certainly different from a "framework developer" perspective :) and yes, I now understand the approach we should follow.

All right, I now added the missing classes and adapted the code I had been working on. I have also added a couple of comments to the PR.

However, I see an error in the CI build (a 401 retrieving a dependency), but it's definitely not related to my changes:

> * What went wrong:
> Could not determine the dependencies of task ':spring-security-oauth2-authorization-server-docs-manual:unzipDocumentationResources'.
> > Could not resolve all dependencies for configuration ':spring-security-oauth2-authorization-server-docs-manual:detachedConfiguration2'.
>    > Could not resolve com.fasterxml.jackson:jackson-bom:2.+.
>      Required by:
>          project :spring-security-oauth2-authorization-server-docs-manual

The build is running ok locally, let me know if I can help with anything to fix this CI issue. I think it might be a temporary issue, but I can't trigger the CI plan manually.

Looking forward to the next comments Joe!

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
rozagerardo: Note: I now created https://github.com/spring-projects-experimental/spring-authorization-server/issues/175 to fix this in the Token Revokation logic.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2Tokens.java
```diff
@@ -87,19 +89,15 @@ public OAuth2RefreshToken getRefreshToken() {
...
-	@SuppressWarnings("unchecked")
-	public <T extends AbstractOAuth2Token> T getToken(String token) {
+	public Optional<AbstractOAuth2Token> getToken(String token) {
 		Assert.hasText(token, "token cannot be empty");
-		OAuth2TokenHolder tokenHolder = this.tokens.values().stream()
+		return this.tokens.values().stream()
 				.filter(holder -> holder.getToken().getTokenValue().equals(token))
-				.findFirst()
-				.orElse(null);
-		return tokenHolder != null ? (T) tokenHolder.getToken() : null;
```
rozagerardo, line 102: I couldn't find a clear approach to follow in the project (I've seen both allowing @Nullable / null values as well as using `Optional`s). I refactored this a little bit to avoid changing the approach in this method, I think it's cleaner, but I can easily revert it.

rozagerardo:
Just a quick note about the last comment I made: https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#issuecomment-737469164
The issue with the ci build seems to be solved, so this should be ready to be reviewed @jgrandja 👍

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
```diff
@@ -0,0 +1,178 @@
...
+			String tokenTypeHint = tokenIntrospectionAuthentication.getTokenTypeHint();
+			if (StringUtils.hasText(tokenTypeHint)) {
+				if (TokenType.REFRESH_TOKEN.getValue().equals(tokenTypeHint)) {
+					tokenType = TokenType.REFRESH_TOKEN;
+				} else if (TokenType.ACCESS_TOKEN.getValue().equals(tokenTypeHint)) {
+					tokenType = TokenType.ACCESS_TOKEN;
+				} else if (TokenType.ID_TOKEN.getValue().equals(tokenTypeHint)) {
+					tokenType = TokenType.ID_TOKEN;
+				} else {
+					throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes2.UNSUPPORTED_TOKEN_TYPE));
```
rozagerardo, line 92: I decided to leave this functionality as we have for the Token Revocation endpoint (as suggested), but just to be clear, I think this does not follow the specs correctly (more info in this issue I created: https://github.com/spring-projects-experimental/spring-authorization-server/issues/174) ,thus I think we'll have to update this logic :)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
```diff
@@ -0,0 +1,178 @@
...
+			throw new IntrospectionTokenException(errorMessages);
+		}
+	}
+
+	/**
+	 * Exception that can be triggered when a token is found invalid.
+	 *
+	 * @author Gerardo Roza
+	 */
+	private static class IntrospectionTokenException extends RuntimeException {
```
jgrandja, line 154: Remove this and instead use `OAuth2AuthenticationException`, e.g.

`throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_GRANT))`

Take a look at `OAuth2AuthorizationCodeAuthenticationProvider` and `OAuth2TokenRevocationAuthenticationProvider` for examples on how exceptions are thrown with associated OAuth 2.0 error codes.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
```diff
@@ -0,0 +1,178 @@
...
+			}
+
+			AbstractOAuth2Token token = authorization.getTokens().getToken(tokenIntrospectionAuthentication.getTokenValue())
+					.get();
+
+			if (authorization.getTokens().getTokenMetadata(token).isInvalidated()) {
+				throw new IntrospectionTokenException("Token has been invalidated");
+			}
+
+			token = parseJwt(token);
```
jgrandja, line 111: No need to `parseJwt()` or `validateToken()` as the original `Jwt` is stored in [OAuth2Authorization.attributes](https://github.com/spring-projects-experimental/spring-authorization-server/blob/8c71e5635089993be98997f3a21589b30f834056/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java#L138)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2Tokens.java
```diff
@@ -87,19 +89,15 @@ public OAuth2RefreshToken getRefreshToken() {
...
-	 * @param token the token
-	 * @param <T> the type of the token
-	 * @return the token, or {@code null} if not available
+	 * @param token the token value
+	 * @return an optional wrapped {@code AbstractOAuth2Token}
 	 */
-	@Nullable
-	@SuppressWarnings("unchecked")
-	public <T extends AbstractOAuth2Token> T getToken(String token) {
+	public Optional<AbstractOAuth2Token> getToken(String token) {
```
jgrandja, line 95: Please revert all changes in `OAuth2Tokens`.
FYI, this will be removed and code merged into `OAuth2Authorization` as part of a remodelling exercise.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/TokenExpirationValidator.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.time.Instant;
+
+/**
+ * An implementation of {@link OAuth2TokenValidator} to verify a token is not expired.
+ *
+ * @author Gerardo Roza
+ * @since 0.0.4
+ */
+public class TokenExpirationValidator implements OAuth2TokenValidator<AbstractOAuth2Token> {
```
jgrandja, line 36: Please remove this and instead use `JwtTimestampValidator`

jgrandja (changes requested):
Thanks for the updates @rozagerardo ! Please see review comments.

Just a heads up that I'll be on PTO from Dec 10 and returning Jan 4. I'll review the next updates when I return. Thanks and enjoy your holidays!

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
rozagerardo: Two things regarding this @jgrandja :
1- We're currently storing the `access_token` `Jwt` in the OAuth2Authorization.attributes. However, we can also introspect a `refresh_token`, or in the future, an `id_token` as well.
What is the plan here? Are we going to store **all** `Jwt` token versions in OAuth2Authorization.attributes (which would explain why the `refresh_token` is not stored; it is not jwt-formatted), or just the `access_token`?
If that is, in fact, the plan, then we can certainly avoid parsing the `Jwt`, the logic, in this case, would be "we try to find the `Jwt` version in the `OAuth2Authorization.attributes` storage if we can't find it we validate it as a regular opaque token".
However, if we still don't have a clear plan, or if we want to gain flexibility (avoid tying this endpoint functionality to what we store in the `OAuth2Authorization.attributes` then the current approach might be more suitable; we simply try to decode the tokenValue using the same configuration we used to encode it.
2- I think we do need the `validateToken()` in any case, to run any validations we consider suitable at the time of introspecting the token.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/TokenExpirationValidator.java
rozagerardo: I am using the `JwtTimestampValidator` when we can decode the token as a `Jwt` @jgrandja , but we also need to cover validating non-jwt-formatted tokens, for example, to validate a `refresh_token`.
Let me know if there is any plan for the future that might affect this (maybe issuing `Jwt` for all type of tokens including `refresh_token`s, or in the opposite direction, maybe we want to support issuing just opaque tokens, which would reinfoce the need of this validator)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
rozagerardo: I am aware of the `OAuth2AuthenticationException` (and understand how this `AuthenticationException` fits in the documentation you suggested me to read here https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#issuecomment-734426330), but note this is on purpose. Let me explain.

Simply put, **in most cases we don't want to retrieve an error response for this endpoint** (like for example an error response indicating the reason why the token is invalid) to avoid exposing any information about the server state, but instead, **we will simply retrieve an `{"active": false}` response**:

> If the protected resource uses OAuth 2.0 client credentials to
>   authenticate to the introspection endpoint and its credentials are
>   invalid, the authorization server responds with an HTTP 401
>   (Unauthorized) as described in Section 5.2 of OAuth 2.0 [RFC6749].
> ...
> **Note that a properly formed and authorized query** for an inactive or
>   otherwise invalid token (or a token the protected resource is not
>   allowed to know about) **is not considered an error response by this**
>   **specification.  In these cases, the authorization server MUST instead**
>   **respond with an introspection response with the "active" field set to**
>   **"false"** as described in Section 2.2.
>   ...
>   Note that **to avoid disclosing too**
>   **much of the authorization server's state** to a third party, the
>   authorization server **SHOULD NOT include any additional information**
>   **about an inactive token, including why the token is inactive.**

So, here:

- we're covering the first case thanks to the `getAuthenticatedClientElseThrowInvalidClient` functionality which does throw an `OAuth2AuthenticationException` when suitable
- and for the rest of the cases, I use this internal `IntrospectionTokenException`, **without leaving it to bubble up**; so we can **stop processing the token at any point of the operation** where we find out the token is not valid and t**he AuthenticationProvider will retrieve an authenticated `OAuth2TokenIntrospectionAuthenticationToken` (bc we're in fact authenticated), but in a way that will result in a `{"active": false }` response**.

If I used a `OAuth2AuthenticationException` for these cases, I would have to catch it, analyze somehow the reason (if it was in fact because the request was unauthorized, or if it was because the token is invalid), and as you can imagine this will be unnecessarily complex.

Of course, if you want to suggest a better approach, I'd love to hear it, as this would help me improve and use better solutions :)

rozagerardo:
I now addressed all the comments, or commented back to discuss a few of them @jgrandja .
I tried to create a commit for each comment, to analyze the change easier. I'll squash them before merging.
Looking forward to your next comments.
Cheers, and hope you had a great end of year!

jgrandja:
Happy New Year @rozagerardo !

Just a heads up that I have a couple of higher priority items that are targeted for the 0.1.0 release at end of month so I'll come back to review this in a couple of weeks.

rozagerardo:
no problem @jgrandja, take your time and let me know if I can help with anything 👍

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
rozagerardo: A quick update here regarding point 2 above.
I realized the `decode` operation we're using now also validates the `Jwt`, so there is no need to validate ourselves in that case.
However, if it is not a JWT-formatted token, then we still want to validate it (using the `AbstractOAuth2Token` fields) to make sure it is not expired for example.
BTW, the logic changed a little bit after we remove the CryptoKeySource (https://github.com/spring-projects-experimental/spring-authorization-server/issues/196) So have a look at the updated changes.

rozagerardo:
Hey @jgrandja just a quick update here, I now updated my code with the latest changes we have in master, this is ready to be reviewed again, and I'll keep an eye to keep the branch updated.
Also, a quick reminder for whenever you get back to this task, these discussions are still open:
https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#discussion_r538767537
https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#discussion_r538771618
https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#discussion_r538776479

jgrandja:
Thanks for the reminder @rozagerardo.
This PR is at the top of my priority list as soon as I release `0.1.0` on Feb 11.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
jgrandja: I would still remove `IntrospectionTokenException` as it's not used anyway other than catching it and returning a `OAuth2TokenIntrospectionAuthenticationToken`. Creating and throwing exceptions is costly so we should avoid it unless we use it.

Instead, perform a fail-fast and return `OAuth2TokenIntrospectionAuthenticationToken` at the point validation fails. You can see that [OAuth2TokenRevocationAuthenticationProvider](https://github.com/spring-projects-experimental/spring-authorization-server/blob/ee1b46b9a627504ed2d0fdaf7308b11fba23fb98/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java#L68) does this.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
jgrandja: There has been a lot of changes to the `OAuth2Authorization` model and it's associated tokens via #213 and subsequent commits.

The important thing to note here is that whatever is returned in `authorizationService.findByToken(tokenValue, null)` is the "source of truth" and the tokens contained within should be used on the code path forward.

@rozagerardo Before I continue with the review, please rebase off of master and integrate the changes. Also, I noticed you did a merge commit - please avoid this and only rebase. Thanks.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/TokenExpirationValidator.java
jgrandja: `TokenExpirationValidator` is not needed. Please see [comment](https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#discussion_r578581645)

jgrandja: Just as an FYI, I'd prefer not to expose a public API if not needed. As soon as we expose, then we need to support, which complicates things if we need to deprecate because of later improvements. So for initial implementations, we expose the minimal set of API's and only "open it up" when it makes sense.

jgrandja:
@rozagerardo I finally circled back to this PR and have replied to the 3 outstanding comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2TokenIntrospectionResponse.java
```diff
@@ -0,0 +1,270 @@
...
+		}
+
+		/**
+		 * Build the {@link OAuth2TokenIntrospectionResponse}
+		 *
+		 * @return The constructed {@link OAuth2TokenIntrospectionResponse}
+		 */
+		public OAuth2TokenIntrospectionResponse build() {
+			Map<String, Object> responseParams = this.params.entrySet().stream()
+					.filter(entry -> SUPPORTED_FIELDS.contains(entry.getKey()))
```
rozagerardo, line 265: The spec indicates:

> Specific implementations MAY extend this structure with their own
   service-specific response names as top-level members of this JSON
   object.

Thus we might want to remove this filter if we want to retrieve custom claims.
For the time being, I limited the response to avoid leaking entries we might have in the token metadata claims.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/introspection/http/converter/HttpMessageConverters.java
```diff
@@ -0,0 +1,67 @@
...
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.json.GsonHttpMessageConverter;
+import org.springframework.http.converter.json.JsonbHttpMessageConverter;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.util.ClassUtils;
+
+/**
+ * TODO
+ * This class is a straight copy from Spring Security.
+ * It should be consolidated when merging this codebase into Spring Security.
```
rozagerardo, line 28: Note I could use `org.springframework.security.oauth2.core.oidc.http.converter.org.springframework.security.oauth2.core.oidc.http.converter` and avoid adding a new class, but it has a protected scope.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
rozagerardo: Sounds good, thanks for the clarification Joe, this is now resolved 👍

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/TokenExpirationValidator.java
rozagerardo: That makes sense, this is now fixed

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
rozagerardo: Cool, I had a look at the new approach, I think this allows a cleaner approach for the Introspection Endpoint. I now adapted the code for these changes.

rozagerardo:
Cool, thanks for the clarifications @jgrandja, all the comments make sense 👍
I now updated the branch with the latest changes in `master` and addressed the comments. This is ready for a new review, looking forward to your comments :)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
```diff
@@ -0,0 +1,114 @@
...
+			return generateAuthenticationTokenForInvalidToken(clientPrincipal, registeredClient);
+		}
+
+		if (isExpired(tokenHolder.getToken())
+				|| (tokenHolder.getClaims() != null && hasInvalidClaims(tokenHolder.getClaims()))) {
+			return generateAuthenticationTokenForInvalidToken(clientPrincipal, registeredClient);
+		}
+
+		return new OAuth2TokenIntrospectionAuthenticationToken(
+				clientPrincipal, registeredClient.getClientId(), tokenHolder);
```
rozagerardo, line 87: We could also add the principal name (obtained from the `OAuth2Authorization`) in the `OAuth2TokenIntrospectionAuthenticationToken` and use it later on in the process to populate the `username` Token Introspection Response field. I just want to confirm this is worth it, as the response field is optional as per the specs 👍

jgrandja:
Thanks for the updates @rozagerardo !
Can you please squash all commits. I'm going to do a full review from fresh.
I realize this will likely remove references to the existing comments but I'm fine with that.
Thanks.

rozagerardo:
Sure @jgrandja the PR now contains only one commit, and I think you can see the comments in the files.
Let me know that's ok, or if I should change anything else :)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,240 @@
...
+	 */
+	private static final class TokenToIntrospectionResponseFieldsMapper {
+
+		private static final Map<Class<? extends AbstractOAuth2Token>, BiConsumer<AbstractOAuth2Token, OAuth2TokenIntrospectionResponse.Builder>> supportedTokens;
+		static {
+			Map<Class<? extends AbstractOAuth2Token>, BiConsumer<AbstractOAuth2Token, OAuth2TokenIntrospectionResponse.Builder>> tokenMap = new HashMap<>();
+			tokenMap.put(
+					OAuth2AccessToken.class,
+					(token, builder) -> extractFromOAuth2AccessToken((OAuth2AccessToken) token, builder));
+			supportedTokens = Collections.unmodifiableMap(tokenMap);
```
rozagerardo, line 177: I know this mapper doesn't make much sense now that we have only one entry (before it contained another entry for `Jwt`) but maybe we want to leave it as a flexible solution in case we can extract relevant information from Refresh Tokens or, more likely, from ID Tokens in the future.

rozagerardo:
Quick ping @jgrandja just to make sure you didn't miss my last comment:
https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#issuecomment-786202371
Although I know you probably are just busy with something else 👍

jgrandja:
@rozagerardo Apologies for the delay. I'm quite backlogged right now and trying to catch up with Spring Security issues and Spring Security OAuth (legacy) issues/PR's. I'm getting close to catching up and I'm planning on circling back to this PR this week. Then I will focus on getting it merged before end of month.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,240 @@
+/*
+ * Copyright 2020 the original author or authors.
```
jgrandja, line 2: Update to `2020-2021`. Please apply to all files.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,240 @@
...
+ */
+public class OAuth2TokenIntrospectionEndpointFilter extends OncePerRequestFilter {
+
+	/**
+	 * The default endpoint {@code URI} for token introspection requests.
+	 */
+	public static final String DEFAULT_TOKEN_INTROSPECTION_ENDPOINT_URI = "/oauth2/introspect";
+
+	private final AuthenticationManager authenticationManager;
+	private final RequestMatcher tokenEndpointMatcher;
```
jgrandja, line 73: Rename to `tokenIntrospectionEndpointMatcher`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,240 @@
...
+
+		if (!this.tokenEndpointMatcher.matches(request)) {
+			filterChain.doFilter(request, response);
+			return;
+		}
+
+		try {
+
+			OAuth2TokenIntrospectionAuthenticationToken introspectionTokenAuthentication = (OAuth2TokenIntrospectionAuthenticationToken) this.authenticationManager
+					.authenticate(this.tokenIntrospectionAuthenticationConverter.convert(request));
```
jgrandja, line 114: Let's assign to variable -> `this.tokenIntrospectionAuthenticationConverter.convert(request)` to allow for better readability

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,240 @@
...
+			throws ServletException, IOException {
+
+		if (!this.tokenEndpointMatcher.matches(request)) {
+			filterChain.doFilter(request, response);
+			return;
+		}
+
+		try {
+
+			OAuth2TokenIntrospectionAuthenticationToken introspectionTokenAuthentication = (OAuth2TokenIntrospectionAuthenticationToken) this.authenticationManager
```
jgrandja, line 113: Rename `introspectionTokenAuthentication` -> `tokenIntrospectionAuthentication`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
```diff
@@ -0,0 +1,240 @@
...
+					.authenticate(this.tokenIntrospectionAuthenticationConverter.convert(request));
+
+			OAuth2TokenIntrospectionResponse tokenIntrospectionResponse = introspectionTokenAuthentication
+					.isTokenActive()
+							? generateTokenIntrospectionResponse(
+									introspectionTokenAuthentication.getTokenHolder(),
+									introspectionTokenAuthentication.getClientId())
+							: OAuth2TokenIntrospectionResponse.builder(false).build();
+
+			this.sendTokenIntrospectionResponse(response, tokenIntrospectionResponse);
```
jgrandja, line 123: Remove `this` - only members are prefixed with `this` NOT instance methods.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
jgrandja: Let's remove this as it's not needed.

In `OAuth2TokenIntrospectionAuthenticationToken` remove `clientId` and `tokenHolder` members and add member `Map<String, Object> claims`. The `OAuth2TokenIntrospectionAuthenticationProvider` has all the context required to produce the `claims` that will be returned in the Token Introspection response via `OAuth2TokenIntrospectionEndpointFilter` as-is.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
```diff
@@ -0,0 +1,113 @@
...
+				tokenIntrospectionAuthentication);
+		RegisteredClient registeredClient = clientPrincipal.getRegisteredClient();
+
+		OAuth2Authorization authorization = this.authorizationService
+				.findByToken(tokenIntrospectionAuthentication.getTokenValue(), null);
+		if (authorization == null) {
+			return generateAuthenticationTokenForInvalidToken(clientPrincipal, registeredClient);
+		}
+
+		if (!registeredClient.getId().equals(authorization.getRegisteredClientId())) {
```
jgrandja, line 70: This check should be removed. Any authenticated client could introspect a token that was not originally granted to it.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2TokenIntrospectionResponse.java
```diff
@@ -0,0 +1,270 @@
...
+import java.util.Map;
+
+/**
+ * A representation of an OAuth 2.0 Introspection Token Response.
+ *
+ * @author Gerardo Roza
+ * @since 0.1.1
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc7662#section-2.2">Section 2.2 - Introspection Response</a>
+ */
+public class OAuth2TokenIntrospectionResponse {
```
jgrandja, line 51: I don't think we need this class at this point. Please try to remove it.

jgrandja (changes requested):
Thanks for the updates @rozagerardo. Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2TokenIntrospectionResponse.java
jgrandja: Actually, on second thought, we do need this. However, please rename to `OAuth2TokenIntrospectionClaims` and move to package `org.springframework.security.oauth2.core` - this will be used on client side as well.

Also, please follow the design of `OidcProviderConfiguration` so you should also add `OAuth2TokenIntrospectionClaimAccessor`.

jgrandja: Also please move `OAuth2TokenIntrospectionResponseHttpMessageConverter` to package `org.springframework.security.oauth2.core.http.converter`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenIntrospectionEndpointFilter.java
rozagerardo: Cool. My idea was to separate as much as possible the response representation model classes from the framework model classes, that is, in the provider handle objects that belong to the core of the domain, and then use the EndpointFilter to map that to the expected response (kind of a layer separation). Of course, the `claims` is actually a plain map, but the structure, the keys it contains are strictly related to the response that is finally serialized. 
Having clarified the reasons why I chose that approach (probably because I'm more used to enterprise solutions, some rules don't totally apply to frameworks/libraries), I understand the reasons for your suggestion, it makes sense. I now changes this (y)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
rozagerardo: I now removed the check @jgrandja but keep in mind that even though the RFC spec doesn't seem to suggest any approach regarding this, Okta seems to have a clear position towards this, which is not allowing clients to introspect others' tokens:
https://www.oauth.com/oauth2-servers/token-introspection-endpoint/

> Any other error is considered an “inactive” token.
>*  The token requested does not exist or is invalid
>*  The token expired
>*  **The token was issued to a different client than is making this request**
>
>In any of these cases, it is not considered an error response, and the endpoint returns simply an inactive flag.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2TokenIntrospectionResponse.java
rozagerardo: Cool, I now did all the changes @jgrandja . Just a heads up regarding the `OAuth2TokenIntrospectionClaimAccessor`, instead of defining and using a `OAuth2IntrospectionClaimNames` class as in the Resource Server version (`OAuth2IntrospectionClaimNames.SCOPE`, `OAuth2IntrospectionClaimNames.TOKEN_TYPE`, `OAuth2IntrospectionClaimNames.EXPIRES_AT`,...), I decided to use the definitions we already have (`OAuth2ParameterNames`, `JwtClaimNames`).

Regarding the JWT fields, remember I mentioned in the ticket description that the spec itself points to the JWT specs when mentioning these Token Introspection response fields.

rozagerardo:
Cool, this is now ready for a new round @jgrandja .
Regarding the comments you made, I now addressed them all, but please have a look at this discussion before marking it as resolved:
https://github.com/spring-projects-experimental/spring-authorization-server/pull/161#discussion_r592711912
(I also made a small comment on some of the other points, but nothing important.)

Also, let me know if you want me to squash the commits again before reviewing them.

Looking forward to the next comments Joe 👍

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
jgrandja: Thanks for being on top of this @rozagerardo ! Can you check to see if Keycloak, Curity and Microsoft Azure enforce this policy? Then we can decide what to do.

Right now I'm thinking we should re-apply the same policy. It's always best to start off more secure than less. We can always open things up later if it makes sense.

rozagerardo: No problem @jgrandja , IMO it's part of the commitment and the ownership I assume when I request working on a feature 👍 

BTW, sorry for the delay with this, I didn't have much free time these last weeks.

Ok, regarding my findings, it seems **all** providers (surprisingly, even Okta) do allow to introspect tokens issued to a different client.

Here are the demo videos for the three providers (Keycloak, Okta and Curity, as Azure AD doesn't yet provide support for the Introspection Endpoint - as indicated [here](https://feedback.azure.com/forums/169401-azure-active-directory/suggestions/18930289-introspection-endpoint-for-azure-active-directory)) showing how they enforce this policy, just in case you want to see this in action:
https://drive.google.com/drive/folders/1LC98BnxTWDl82hCp3RYBWq9KoZT5n2nz?usp=sharing

Also, just FYI I reported a question in the Okta support center just to clarify if they should update the documentation:
https://devforum.okta.com/t/introspection-endpoint-should-work-for-tokens-issued-by-different-client/13587

dfcoffin: Section 4. Security Considerations of [RFC 7662] OAuth 2.0 Token Introspection discusses the token validation steps required by the AS when processing the Introspection request:

```
   o  If the token can expire, the authorization server MUST determine
      whether or not the token has expired.
   o  If the token can be issued before it is able to be used, the
      authorization server MUST determine whether or not a token's valid
      period has started yet.
   o  If the token can be revoked after it was issued, the authorization
      server MUST determine whether or not such a revocation has taken
      place.
   o  If the token has been signed, the authorization server MUST
      validate the signature.
   o  If the token can be used only at certain resource servers, the
      authorization server MUST determine whether or not the token can
      be used at the resource server making the introspection call.
      
   If an authorization server fails to perform any applicable check, the
   resource server could make an erroneous security decision based on
   that response.  Note that not all of these checks will be applicable
   to all OAuth 2.0 deployments and it is up to the authorization server
   to determine which of these checks (and any other checks) apply.
   
   If left unprotected and un-throttled, the introspection endpoint
   could present a means for an attacker to poll a series of possible
   token values, fishing for a valid token.
```

Based on the last of the AS Token Endpoint validation requirements, IMO you've discovered potential security issues with Keycloak, Okta, and Curity AS software.

rozagerardo: Thanks for the comment @dfcoffin.

I interpreted the last validation requirement differently, IMO the _client_id_ doesn't relate strictly with the _resource servers_ in the sense of serving to "identify" a particular Resource Server (in fact, it would be used to identify a particular Client, but not the Resource Server, which is the one consuming this endpoint. Read the last section about Curity to know how they modeled this). In fact, the RFC gives a sort of example of how this Resource Server whitelist/blacklist would look like:

> The introspection endpoint MAY accept other OPTIONAL parameters to
   provide further context to the query.  For instance, an authorization
   server may desire to know the IP address of the client accessing the
   protected resource to determine if the correct client is likely to be
   presenting the token.

In my opinion, this is what the restriction refers to, a sort of IP-based firewall to restrict requests from just any origin.

Of course, I might be getting this wrong, let me know what you think.

Now, just to add some more information, in the case of Curity, it is definitely a conscious decision, on their platform you have to enable explicitly the introspection functionality granularly on an application/client, and they suggest not mixing this feature with other token-issuer clients; they separate this functionality as it is (commonly) used by Resource Servers instead of Clients:


> https://curity.io/resources/learn/introspect-with-phantom-token/#configure-client-for-introspection-gateway-client

rozagerardo: Awesome, thanks for the very interesting feedback @dfcoffin :)

Yes, the Client names used in the videos might not be totally representative, as my point was only to analyze if a Client (of any type) could introspect tokens issued by another Client (again, of any type, or with any enabled feature).

Anyway, IMO the feedback provided by Justin strengthens the idea that is ok to authenticate using a particular Client's credentials to introspect another Client's tokens, as a Client setup that is configured to issue tokens is not necessarily the same that will introspect the token.

This is especially clear for Public Clients; they would need a different, confidential Client to be able to Introspect its tokens.

jgrandja: @dfcoffin @rozagerardo Regarding the following reference in RFC 7662:

> If the token can be used only at certain resource servers, the
   authorization server MUST determine whether or not the token can
   be used at the resource server making the introspection call.

This seems very related to how [Resource Indicators for OAuth 2.0](https://tools.ietf.org/html/rfc8707) would be implemented.

As far as Curity goes:

> Introspection clients are rarely the same as clients that can request tokens. A gateway client should be setup to only be allowed to do introspection and no other OAuth flow.

This makes total sense to me and I really do like how Curity has implemented their product.

The 2 configuration options that I would favour for an introspection-specific client is Mutual TLS client authentication and Bearer Token client authentication. The bearer token would have introspection-specific scopes associated with it (e.g. `token.introspect`) and a client certificate claim (e.g. `x5t#S256` - X.509 Certificate SHA-256 Thumbprint).

rozagerardo: Thanks for the comments @jgrandja!

Let's see, regarding the first note about Resource Indicators for OAuth 2.0, that is indeed very interesting, and it seems it would be related, but I'm having some doubts.

I just had a quick look at RFC 8707 so I might be missing a few things, but basically, it seems this mechanism would impact mainly in the "audience" value the Authorization Server retrieves:

> The authorization server **SHOULD audience-restrict** issued access
   tokens to the resource(s) indicated by the "resource" parameter.
   Audience restrictions can be communicated in JSON Web Tokens
   [RFC7519] **with the "aud" claim and the top-level member of the same
   name provides the audience restriction information in a Token
   Introspection** [RFC7662] response.
   
So, going back to the original note in the Token Introspection spec:

>  o  If the token can be used only at certain resource servers, the
      authorization server MUST determine whether or not the token can
      be used at the resource server making the introspection call.
	  
I'm not sure a check based on the RFC 8707 feature corresponds to the one mentioned above, if the first impacts the audience, then IMO it should be the Resource Server the one deciding if the token represents authorized access to its resource...We could consider the following behavior though: 

- if the _resource_ is a "_https URL_" then it might be suitable to perform the validation right away in the AS Introspection Endpoint and consider the token invalid if introspected by a different Resource Server
- if the _resource_ is a "_more abstract identifier_", then simply use it as the "aud" field and consider the token "active" 

But this would be whole new discussion and we shouldn't loose focus :)

Finally, regarding the two configuration options you mention, yes, I think that makes sense; an introspection-specific client would mandatorily be a confidential client thus it would be inherently suitable for these authentication methods... I don't totally understand the use of forcing the client to add a particular scope in the Bearer token for this, but I haven't used this mechanism much, so I'd love to see/participate in the development of that feature.

Now, regarding the next steps here, we don't yet have support for these authentication methods, right? So, maybe we need to create a few tickets (maybe to allow setting introspection as a feature on client configurations, analyze implementing Resource Indicators, and to support different methods of client authentication), but is there anything that should be part of this particular ticket?

Looking forward to your comments regarding this Joe (and @dfcoffin, of course), and let me know if you want me to create tickets for some of the features discussed here 👍 

PS: I found very appealing the solution used by Curity too :)

dfcoffin: @jgrandja @rozagerardo 

> Now, regarding the next steps here, we don't yet have support for these authentication methods, right? So, maybe we need to create a few tickets (maybe to allow setting introspection as a feature on client configurations, analyze implementing Resource Indicators, and to support different methods of client authentication), but is there anything that should be part of this particular ticket?

RFC 8707 only references "aud" once and JWT three times in a single paragraph. On the other hand, it references RFC 6749 twelve times.  Based on the occurrences of RFC6749 and the inclusion in the introduction of the following:

> Bearer tokens, currently the most commonly utilized type of OAuth access token, allow any party in possession of a token to get access to the associated resources.  To prevent misuse, several important security assumptions must hold, one of which is that an access token must only be valid for use at a specific protected resource and for a specific scope of access.  Section 5.2 of [RFC6750], for example, prescribes including the token's intended recipients within the token to prevent token redirect.  When the authorization server is informed of the resource that will process the access token, it can restrict the intended audience of that token to the given resource such that the token cannot be used successfully
> at other resources.

IMO we need to list RFC 8707 as an EPIC since its implementation will require modification of several existing filters and endpoints to determine if the "resource" parameter is present in the Authorization Request and how to handle it.  Adding a "TODO" flag to review when RFC 8707 Resouce Indicators for OAuth 2.0 is implemented.

rozagerardo: Right, that was kind of where I was heading too, IMO these topics have an impact that goes beyond this current feature, and I don't think anything in the code involved here should be changed (at this point).

Let me know if you want me to create a ticket for any of the issues discussed here, and just to be clear regarding this @dfcoffin:

> Adding a "TODO" flag to review when RFC 8707 Resouce Indicators for OAuth 2.0 is implemented.

I'm not totally sure what you mean, is that referring just to a personal reminder you have? I mean, that is not something we should add to the code right? (I've seen a just couple of TODO flags, but I think this shouldn't be the case). Sorry about this doubt, I'm not totally familiarized with the tickets/features management process you guys follow :)

cc @jgrandja 

PS Donald, note that the section you quoted about Bearer tokens also mentions that the _Resource Indicators_ feature is carried out by restricting the audience of the token :)

jgrandja: @rozagerardo @dfcoffin Thanks for all your feedback.

I just logged a ticket for Resource Indicators gh-279.

To recap on how I view Resource Indicators will be implemented:

1) The client sends the `resource` parameter in the authorization or token request. Its value can be 1 or more absolute URI's.
2) The authorization server will use the `aud` claim to represent the value(s) in the `resource` parameter. The authorization server MAY also apply some policy decisions around the resource and further enhance the token.
3) The resource server will use the `aud` claim to apply authorization decisioning on whether to allow the request to proceed. This is where it checks if it's the "intended audience" of the incoming token.

@rozagerardo Resource indicators is a separate feature so there is no additional work for this PR. Apologies for the delay on this PR but I've been quite busy on the OAuth Client side of things as we're releasing Spring Security 5.5.0-RC1 today. My time has now opened up and I'm refocusing back on Authorization Server starting today. I'm circling back to this PR with the plan on merging it before end of week.

Regarding client authentication method support, this will be a major theme for `0.1.2`. See gh-59 and gh-101

rozagerardo: Cool, everything sounds good to me @jgrandja thanks for the update!

Don't worry for the times BTW, I know you have a lot on your plate and I also understand that new projects are not totally stable in the early stages; keeping the PR updated does require some extra effort, but that's totally expected and is part of my commitment towards the project :)

Of course, let me know if I can help with anything else, either with this ticket or with any other task; I'll probably pick up a new ticket at some point, but some pressure to grab a new task works for me too :) cheers!

jgrandja:
@rozagerardo Thanks for all your work ! This is now merged to master.

FYI, I applied some updates with a follow-up polish commit. Please review and let me know if you have any questions.

PR #162, opened 2020-11-18
Add end-to-end Integration test to sample project

rozagerardo:
Hey @jgrandja as I was working on my Token Introspection Endpoint ticket I created an Integration Test for the sample project.
I used it to debug the whole process and get familiarized with the code, and naturally to have a good look at what I was really obtaining in the Endpoint I was developing.
So, I decided to create a separate PR for this, just in case you think it might be useful for other devs as well. Of course, this is currently limited to the functionality I was working on, but it can be improved to make it more general-purpose.
I also have a Postman collection which simulates the same process I execute in the Integration test here, but that's probably less useful.

### DIFF
--- a/samples/boot/oauth2-integration/authorizationserver/src/test/java/sample/OAuth2AuthorizationServerIntegrationTests.java
+++ b/samples/boot/oauth2-integration/authorizationserver/src/test/java/sample/OAuth2AuthorizationServerIntegrationTests.java
@@ -1,0 +1,166 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package sample;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.hamcrest.Matchers.allOf;
+import static org.hamcrest.Matchers.contains;
+import static org.hamcrest.Matchers.containsString;
+import static org.hamcrest.Matchers.greaterThan;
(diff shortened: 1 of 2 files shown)


samples/boot/oauth2-integration/authorizationserver/src/test/java/sample/OAuth2AuthorizationServerIntegrationTests.java
```diff
@@ -0,0 +1,166 @@
...
+import java.time.Instant;
+
+/**
+ * End-to-end Integration Tests for the application using Spring Authorization Server features.
+ *
+ * @author Gerardo Roza
+ */
+@SpringBootTest
+@AutoConfigureMockMvc
+public class OAuth2AuthorizationServerIntegrationTests {
```
jgrandja, line 58: Did you see the integration tests under package `org.springframework.security.config.annotation.web.configurers.oauth2.server.authorization` ?

`OAuth2AuthorizationCodeGrantTests`, `OAuth2TokenRevocationTests`, etc.

This is where you can put integration tests for the PR.

rozagerardo: Hi @jgrandja yes I saw that and I, in fact, provided the corresponding IntegrationTest for the ticket I was working on:
https://github.com/spring-projects-experimental/spring-authorization-server/pull/161/files#diff-48c0ff0541896c7f642329cadb8791995de250a61fec9d6a1ce224eba04a9d72R75
However, these tests are slightly different, they are a little bit broader since they execute the whole flow (login, authorize consent, request a token, introspect the access and the refresh tokens).
This helped me validate that everything was indeed working as expected between all these different processes (we're mocking up these things in the Integration Tests mentioned above), and it allowed me to explore the real status of the application at the time of calling the Token Introspection endpoint I was working on, for example, to check that the JwtDecoder I configured was able to decode the JWT that our application created.
I understand these tests are not critical, but they helped me in the process, so I wanted to offer them in case it's of use to anyone else.

jgrandja: Thanks for the offer @rozagerardo. Let's keep the integration tests in the core module as you currently have with `OAuth2TokenIntrospectionTests`. I realize these tests are a bit different but we already have the other flows covered in their existing integration tests so I'd rather not duplicate. Thank you for your understanding.

PR #167, opened 2020-11-26
Implement OAuth 2.0 Server Metadata (RFC 8414)

### LINKED ISSUE #54: Epic: OAuth 2.0 Authorization Server Metadata

Kehrlann:
See https://tools.ietf.org/html/rfc8414

Closes #54

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
@@ -49,6 +49,7 @@ import org.springframework.security.oauth2.server.authorization.config.ProviderS
 import org.springframework.security.oauth2.server.authorization.oidc.web.OidcProviderConfigurationEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.NimbusJwkSetEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2AuthorizationEndpointFilter;
+import org.springframework.security.oauth2.server.authorization.web.OAuth2AuthorizationServerConfigurationEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2ClientAuthenticationFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2TokenEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2TokenIntrospectionEndpointFilter;
@@ -89,13 +90,15 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 	private RequestMatcher tokenRevocationEndpointMatcher;
 	private RequestMatcher jwkSetEndpointMatcher;
 	private RequestMatcher oidcProviderConfigurationEndpointMatcher;
+	private RequestMatcher oauth2ServerConfigurationEndpointMatcher;
 	private final RequestMatcher endpointsMatcher = (request) ->
 			this.authorizationEndpointMatcher.matches(request) ||
 			this.tokenEndpointMatcher.matches(request) ||
 			this.tokenIntrospectionEndpointMatcher.matches(request) ||
 			this.tokenRevocationEndpointMatcher.matches(request) ||
 			this.jwkSetEndpointMatcher.matches(request) ||
-			this.oidcProviderConfigurationEndpointMatcher.matches(request);
+			this.oidcProviderConfigurationEndpointMatcher.matches(request) ||
+			this.oauth2ServerConfigurationEndpointMatcher.matches(request);
 
 	/**
 	 * Sets the repository of registered clients.
@@ -214,6 +217,10 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 			OidcProviderConfigurationEndpointFilter oidcProviderConfigurationEndpointFilter =
 					new OidcProviderConfigurationEndpointFilter(providerSettings);
 			builder.addFilterBefore(postProcess(oidcProviderConfigurationEndpointFilter), AbstractPreAuthenticatedProcessingFilter.class);
+
+			OAuth2AuthorizationServerConfigurationEndpointFilter authorizationServerConfigurationFilter
+					= new OAuth2AuthorizationServerConfigurationEndpointFilter(providerSettings);
+			builder.addFilterBefore(postProcess(authorizationServerConfigurationFilter), AbstractPreAuthenticatedProcessingFilter.class);
 		}
 
 		JWKSource<SecurityContext> jwkSource = getJwkSource(builder);
@@ -277,6 +284,8 @@ public final class OAuth2AuthorizationServerConfigurer<B extends HttpSecurityBui
 				providerSettings.jwkSetEndpoint(), HttpMethod.GET.name());
 		this.oidcProviderConfigurationEndpointMatcher = new AntPathRequestMatcher(
 				OidcProviderConfigurationEndpointFilter.DEFAULT_OIDC_PROVIDER_CONFIGURATION_ENDPOINT_URI, HttpMethod.GET.name());
+		this.oauth2ServerConfigurationEndpointMatcher = new AntPathRequestMatcher(
+				OAuth2AuthorizationServerConfigurationEndpointFilter.DEFAULT_OAUTH2_AUTHORIZATION_SERVER_CONFIGURATION_ENDPOINT_URI, HttpMethod.GET.name());
 	}
 
 	private static void validateProviderSettings(ProviderSettings providerSettings) {
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2AuthorizationServerConfiguration.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2AuthorizationServerConfiguration.java
@@ -1,0 +1,370 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.core;
+
+import org.springframework.util.Assert;
+
+import java.io.Serializable;
+import java.net.URI;
+import java.net.URL;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+import java.util.Set;
+import java.util.function.Consumer;
+import java.util.stream.Collectors;
+
+
+/**
+ * A base representation of a Provider Configuration response, returned by an endpoint defined
+ * either in OpenID Connect Discovery 1.0 or OAuth 2.0 Authorization Server Metadata.
+ * It contains a set of claims about the Provider's configuration.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @see OAuth2AuthorizationServerMetadataClaimAccessor
+ * @since 0.1.1
+ * @see <a target="_blank" href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse">4.2. OpenID Provider Configuration Response</a>
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc8414#section-3.2">3.2. Authorization Server Metadata Response</a>
+ */
+public abstract class AbstractOAuth2AuthorizationServerConfiguration implements OAuth2AuthorizationServerMetadataClaimAccessor, Serializable {
+	private static final long serialVersionUID = Version.SERIAL_VERSION_UID;
+
+	protected final Map<String, Object> claims;
+
+	protected AbstractOAuth2AuthorizationServerConfiguration(Map<String, Object> claims) {
+		Assert.notEmpty(claims, "claims cannot be empty");
+		this.claims = Collections.unmodifiableMap(new LinkedHashMap<>(claims));
+	}
+
+	/**
+	 * Returns the Authorization Server metadata.
+	 *
+	 * @return a {@code Map} of the metadata values
+	 */
+	@Override
+	public Map<String, Object> getClaims() {
+		return this.claims;
+	}
+
+	/**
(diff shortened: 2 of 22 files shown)


Kehrlann:
This is still a draft. Remaining todos are:
- Integration tests
- General polishing
- Make sure test coverage is appropriate
- Ensuring consistency with ab591d (polish commit for OIDC Provider Configuration)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractAuthorizationServerMetadata.java
```diff
@@ -0,0 +1,338 @@
...
+	/**
+	 * An abstract builder for subclasses of {@link AbstractAuthorizationServerMetadata}.
+	 */
+	public static abstract class AbstractBuilder<T extends AbstractAuthorizationServerMetadata, B extends AbstractBuilder<T, B>> {
+
+		protected final Map<String, Object> claims = new LinkedHashMap<>();
+
+		protected AbstractBuilder() { }
+
+		protected abstract B getThis();	// avoid unchecked casts in builder when doing "return (T) this;"
```
Kehrlann, line 107: See https://stackoverflow.com/questions/17164375/subclassing-a-java-builder-class/34741836#34741836

oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerMetadataTests.java
```diff
@@ -0,0 +1,97 @@
...
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.jsonPath;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
+ * Integration tests for OAuth 2.0 Authorization Server Metadata.
+ *
+ * @author Daniel Garnier-Moiroux
+ */
+public class OAuth2AuthorizationServerMetadataTests {
```
Kehrlann, line 44: I am not convinced by the name. These are the integration tests; there's a name conflict with those of the domain object, which are in `org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationServerMetadataTests` .

oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerMetadataTests.java
```diff
@@ -0,0 +1,97 @@
...
+
+	@EnableWebSecurity
+	@Import(OAuth2AuthorizationServerConfiguration.class)
+	static class AuthorizationServerConfigurationWithIssuer extends AuthorizationServerConfiguration {
+
+		@Bean
+		ProviderSettings providerSettings() {
+			return new ProviderSettings().issuer(issuerUrl);
+		}
+	}
```
Kehrlann, line 96: I didn't replicate the "loadContext" tests found in `OidcTests` as they would basically exercise the exact same failure paths (e.g. configuration error with an issuer that's an invalid URL)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractAuthorizationServerMetadata.java
```diff
@@ -0,0 +1,338 @@
...
+					.filter(entry -> entry.getValue() != null)
+					.filter(entry -> Collection.class.isAssignableFrom(entry.getValue().getClass()))
+					.filter(entry -> ((Collection<?>) entry.getValue()).size() == 0)
+					.map(Map.Entry::getKey)
+					.collect(Collectors.toSet());
+
+			for (String claimToRemove : claimsToRemove) {
+				this.claims.remove(claimToRemove);
+			}
+		}
```
Kehrlann, line 307: This mutates the `claims` field ; I dislike it a bit. Subclasses will implement their `build` methods like this:

```java
public XXX build() {
	validateCommonClaims();
	validateXXXClaims();
	removeEmptyClaims();
	return new XXX(this.claims);
}

```

The alternative I considered was a more functional `protected static Map<String, Object> omitEmptyClaims(Map<String, Object> claims)`, that subclasses could use like so:

```java
public XXX build() {
	validateCommonClaims();
	validateXXXClaims();
	return new XXX(omitEmptyClaims(this.claims));
}
```

However the `validateClaims` throw runtime exceptions, so I stuck `removeEmptyClaims`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractAuthorizationServerMetadata.java
```diff
@@ -0,0 +1,338 @@
...
+			claimsConsumer.accept(this.claims);
+			return getThis();
+		}
+
+		/**
+		 * Creates the {@link AbstractAuthorizationServerMetadata}.
+		 *
+		 * @return the {@link AbstractAuthorizationServerMetadata}
+		 */
+		public abstract T build();
```
Kehrlann, line 272: I prefer leaving this one abstract, the implementation is small enough that it feels ok to have it in the subclasses.

For reference both subclasses follow the same implementation:
- validate common claims
- validate subclass-specific claims (oidc or oauth 2.0 server metadata)
- remove empty lists from claims
- build

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/endpoint/OAuth2AuthorizationServerMetadata.java
```diff
@@ -0,0 +1,180 @@
...
+ * The claims are defined by the OAuth 2.0 Authorization Server Metadata
+ * specification (RFC 8414).
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see AbstractAuthorizationServerMetadata
+ * @see OAuth2AuthorizationServerMetadataClaimAccessor
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc8414#section-3.2">3.2. Authorization Server Metadata Response</a>
+ */
+public final class OAuth2AuthorizationServerMetadata extends AbstractAuthorizationServerMetadata
```
jgrandja, line 42: Rename to `OAuth2AuthorizationServerConfiguration`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractAuthorizationServerMetadata.java
```diff
@@ -0,0 +1,337 @@
...
+ * either in OpenID Connect Discovery 1.0 or OAuth 2.0 Authorization Server Metadata.
+ * It contains a set of claims about the Provider's configuration.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see AuthorizationServerMetadataClaimAccessor
+ * @see <a target="_blank" href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfigurationResponse">4.2. OpenID Provider Configuration Response</a>
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc8414#section-3.2">3.2. Authorization Server Metadata Response</a>
+ */
+public abstract class AbstractAuthorizationServerMetadata implements AuthorizationServerMetadataClaimAccessor, Serializable {
```
jgrandja, line 49: Rename to `AbstractAuthorizationServerConfiguration`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/http/converter/OAuth2AuthorizationServerMetadataHttpMessageConverter.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * A {@link HttpMessageConverter} for an {@link OAuth2AuthorizationServerMetadata OAuth 2.0 Authorization Server Metadata Response}.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see AbstractHttpMessageConverter
+ * @see OAuth2AuthorizationServerMetadata
+ */
+public class OAuth2AuthorizationServerMetadataHttpMessageConverter
```
jgrandja, line 49: Rename to `OAuth2AuthorizationServerConfigurationHttpMessageConverter`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationServerMetadataEndpointFilter.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * A {@code Filter} that processes OAuth 2.0 Authorization Server Metadata Requests.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see ProviderSettings
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc8414#section-3">3. Obtaining Authorization Server Metadata</a>
+ */
+public class OAuth2AuthorizationServerMetadataEndpointFilter extends OncePerRequestFilter {
```
jgrandja, line 44: Rename to `OAuth2AuthorizationServerConfigurationEndpointFilter`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AuthorizationServerMetadataClaimAccessor.java
```diff
@@ -0,0 +1,107 @@
...
+ * its configuration, used either in OpenID Connect Discovery 1.0 or OAuth 2.0 Authorization
+ * Server Metadata.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 0.1.0
+ * @see ClaimAccessor
+ * @see <a target="_blank" href="https://tools.ietf.org/html/rfc8414#section-2">2. Authorization Server Metadata</a>
+ * @see <a target="_blank" href="https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata">3. OpenID Provider Metadata</a>
+ */
+public interface AuthorizationServerMetadataClaimAccessor extends ClaimAccessor {
```
jgrandja, line 33: I feel like we should merge this into `OAuth2AuthorizationServerMetadataClaimAccessor`. Then we can remove `AuthorizationServerMetadataClaimAccessor`, `AuthorizationServerMetadataClaimNames` and `AbstractAuthorizationServerMetadata`. This change will simplify the hierarchy.

Although `revocation_endpoint`, `revocation_endpoint_auth_methods_supported` and `code_challenge_methods_supported` are not defined in OIDC Provider Metadata, I believe an OIDC provider will support these configurations either way.

What are your thoughts on this proposed change?

jgrandja (changes requested):
Thanks for the PR @Kehrlann ! Please see review comments.

In addition to the comments, please ensure copyright year is `2020-2021` and `@since` is `0.1.1` for all files.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AuthorizationServerMetadataClaimAccessor.java
Kehrlann: 👍  definitely agree with the proposed change

Quick research:
- Okta supports those three claims in their OIDC configuration
- [Google](https://accounts.google.com/.well-known/openid-configuration) supports `revocation_endpoint` and `code_challenge_methods_supported`
- [Microsoft](https://login.microsoftonline.com/common/v2.0/.well-known/openid-configuration) supports none of them

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationServerMetadataEndpointFilter.java
Kehrlann: Should we also rename
`DEFAULT_OAUTH2_AUTHORIZATION_SERVER_METADATA_ENDPOINT_URI` 
to
`DEFAULT_OAUTH2_AUTHORIZATION_SERVER_CONFIGURATION_ENDPOINT_URI`

in that class ?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AuthorizationServerMetadataClaimAccessor.java
Kehrlann: One caveat, removing `AbstractAuthorizationServerMetadata`, using `OAuth2AuthorizationServerConfiguration` instead in `OidcProviderConfiguration extends OAuth2AuthorizationServerConfiguration` 

Is only possible if we make `OAuth2AuthorizationServerConfiguration` not final. And I believe we want our model classes to be final?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationServerMetadataEndpointFilter.java
jgrandja: Yes please

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AuthorizationServerMetadataClaimAccessor.java
jgrandja: Our typical design process is to make things final (or restrict access) at the start. However, if there is need to open things up than we do only if it makes sense. This is one of those instances so it's ok to make `OAuth2AuthorizationServerConfiguration` non-final BUT make the constructor `protected`.

Kehrlann: So I've tried this, unfortunately the type hierarchy with generic builders does not allow to have this kind of hierarchy:

```
- OAuth2AuthorizationServerConfiguration
  - Builder (inner class)
  - static builder() method, returning a Builder
  - OidcProviderConfiguration extends OAuth2AuthorizationServerConfiguration
    - OidcBuilder (inner class, extends Builder) 
    - static builder() method, returning an OidcBuilder
```

Problems are:
- If `OidcBuilder` extends the concrete class `OAuth2AuthorizationServerConfiguration.Builder`, the builder pattern won't work as, for example, `builder.issuer("some string")` would always return the parent `Builder` type. I don't think we can get away without a base abstract builder, see [this Stackoverflow thread for reference](https://stackoverflow.com/questions/21086417/builder-pattern-and-inheritance) for reference.
- If both Builder and OidcBuilder inherit a common, generic, AbstractBuilder, then OidcProviderConfiguration.builder() return type becomes incompatible with that of `OAuth2AuthorizationServerConfiguration.builder()`.

If we really want to get rid of the `AbstractAuthorizationServerConfiguration`, we could have the abstract builder and default claims in the `AuthorizationServerMetadataClaimAccessor` interface.

jgrandja:
Thanks for all the work @Kehrlann ! This is now merged.

I decided to rename `OAuth2AuthorizationServerConfiguration` back to the original `OAuth2AuthorizationServerMetadata`. It made more sense.

I also made a few other changes in a follow-up polish commit. Please let me know if you have any questions in regards to the changes. Thanks!

Issue #155, opened 2020-11-15
Refresh token should not be issued if client is not configured with refresh_token grant type


lspil:
This behavior is strange from my point of view. If I enable the refresh token

.tokenSettings(t -> {
          t.enableRefreshTokens(true);
 })

Which is also the default out of my observations. Even if the client doesn't support the REFRESH_TOKEN grant type, I get a refresh token in the response. Now, of course, I can never use that value because the client doesn't support this grant type. So why do I need to get the value at all then?

Maybe it's just my opinion, but I find this behavior strange. Shouldn't these two configurations be somehow coordinated?

lspil:
I forgot to mention I reproduce this in version 0.0.3

jgrandja:
Good catch @lspil. Would you be interested in submitting a fix for this?

lspil:
Hi. Sorry for the late answer. If nobody took it yet, yes. I could. Let me know if you assign it to me. Thanks :)

jgrandja:
Thanks @lspil. The issue is yours!

PR #168, opened 2020-11-28
Refresh token should not be issued if client is not configured with refresh_token grant type#155

lspil:
A refresh token was generated even if the client didn't have the refresh token grant type assigned.
The refresh token is now generated only for a client registered for the refresh token grant type.

Closes gh-155

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
@@ -18,11 +18,7 @@ package org.springframework.security.oauth2.server.authorization.authentication;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
-import org.springframework.security.oauth2.core.OAuth2AccessToken;
-import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
-import org.springframework.security.oauth2.core.OAuth2Error;
-import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
-import org.springframework.security.oauth2.core.OAuth2RefreshToken;
+import org.springframework.security.oauth2.core.*;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
 import org.springframework.security.oauth2.jwt.Jwt;
 import org.springframework.security.oauth2.jwt.JwtEncoder;
@@ -126,7 +122,7 @@ public class OAuth2AuthorizationCodeAuthenticationProvider implements Authentica
 				.accessToken(accessToken);
 
 		OAuth2RefreshToken refreshToken = null;
-		if (registeredClient.getTokenSettings().enableRefreshTokens()) {
+		if (isConfiguredForRefreshToken(registeredClient)) {
 			refreshToken = OAuth2TokenIssuerUtil.issueRefreshToken(registeredClient.getTokenSettings().refreshTokenTimeToLive());
 			tokensBuilder.refreshToken(refreshToken);
 		}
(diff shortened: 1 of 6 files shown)


lspil:
Hi, This issue seems to be a fault of mine not working often with GitHub. My pull request that solves #155 created this as a new issue. Sorry for this problem.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
```diff
@@ -149,4 +145,9 @@ public Authentication authenticate(Authentication authentication) throws Authent
 	public boolean supports(Class<?> authentication) {
 		return OAuth2AuthorizationCodeAuthenticationToken.class.isAssignableFrom(authentication);
 	}
+
+	private boolean isConfiguredForRefreshToken(RegisteredClient registeredClient) {
+		return registeredClient.getTokenSettings().enableRefreshTokens() &&
+				registeredClient.getAuthorizationGrantTypes().contains(AuthorizationGrantType.REFRESH_TOKEN);
```
alek-sys, line 151: Just my 2 cents here since I was working on the original PR to support refresh token grant.  

I somehow missed this supported authorization grant types. And now I'm not even sure now how these two settings can be separated. It almost feels like `enableRefreshToken` flag can be removed completely and if the grant type is supported then refresh tokens should be issued.

lspil: Yes. I agree. I don't see a reason for that flag. Should I then remove it completely? I think I need some second opinion here before I do that change.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
```diff
@@ -18,11 +18,7 @@
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
-import org.springframework.security.oauth2.core.OAuth2AccessToken;
-import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
-import org.springframework.security.oauth2.core.OAuth2Error;
-import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
-import org.springframework.security.oauth2.core.OAuth2RefreshToken;
+import org.springframework.security.oauth2.core.*;
```
jgrandja, line 21: Please revert this as we don't use wildcard for package imports.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
jgrandja: I agree as well. Let's remove `enableRefreshToken()` and simply check `registeredClient.getAuthorizationGrantTypes().contains(AuthorizationGrantType.REFRESH_TOKEN)`.

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/TestRegisteredClients.java
```diff
@@ -53,6 +53,18 @@
 				.scope("scope2");
 	}
 
+	public static RegisteredClient.Builder registeredClient3() {
```
jgrandja, line 56: Let's remove this and reuse an existing `RegisteredClient` template. You can easily adjust the configuration via the returned `Builder`.

jgrandja (changes requested):
Thanks for the PR @lspil !

Just so you are aware of the PR process, after you submit a PR it does create a new type of issue and therefore resulting in 2 issues - the original issue and the PR issue. This is the normal process. After we merge the PR, we will close it and label as a duplicate.

jgrandja:
Thanks for the updates @lspil ! This is now in master.

FYI, I added a polish commit via 8c71e56.

Issue #170, opened 2020-11-28
NullPointerException when not specifying the token_type_hint for the revoke token endpoint


lspil:
Hi. I got a new issue. If I omit to specify the token_type_hint request parameter for the revocation token I get a NullPointerException.
In my opinion, I should get a 4XX HTTP status with a nice message as it happens if I omit the token for example (or for the other endpoints like /token). Please let me know if you consider this an issue as well and if yes, you can assign it to me. I can take care of it.

**Steps to reproduce:** Call the /oauth2/revoke endpoint without specifying the token_type_hint request parameter.

**Result:**

java.lang.NullPointerException: Cannot invoke "org.springframework.security.oauth2.server.authorization.TokenType.getValue()" because "tokenType" is null
	at org.springframework.security.oauth2.server.authorization.InMemoryOAuth2AuthorizationService.hasToken(InMemoryOAuth2AuthorizationService.java:67) ~[main/:na]
	at org.springframework.security.oauth2.server.authorization.InMemoryOAuth2AuthorizationService.lambda$findByToken$0(InMemoryOAuth2AuthorizationService.java:61) ~[main/:na]
	at java.base/java.util.stream.ReferencePipeline$2$1.accept(ReferencePipeline.java:176) ~[na:na]
	at java.base/java.util.concurrent.ConcurrentHashMap$ValueSpliterator.tryAdvance(ConcurrentHashMap.java:3620) ~[na:na]
	at java.base/java.util.stream.ReferencePipeline.forEachWithCancel(ReferencePipeline.java:127) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.copyIntoWithCancel(AbstractPipeline.java:502) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:488) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474) ~[na:na]
	at java.base/java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:150) ~[na:na]
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234) ~[na:na]

lspil:
So, I actually see this paramters is OPTIONAL in the RFC. So I guess if it's not given, the authorization server should be able to find the token without the hint. Still, it shouldn't throw a NullPointerException

[...]

PR #171, opened 2020-12-06
NullPointerException when not specifying the token_type_hint for the revoke token endpoint#170

lspil:
Solving issue #170

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
@@ -64,6 +64,10 @@ public final class InMemoryOAuth2AuthorizationService implements OAuth2Authoriza
 	}
 
 	private boolean hasToken(OAuth2Authorization authorization, String token, TokenType tokenType) {
+		if (tokenType == null) {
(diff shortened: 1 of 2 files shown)


lspil:
I wanted to ask before changing a lot of things. Should we make that TokenType an Optional counting that we assume it might get null values. I'm not really a fan of using null. Let me know your opinion please :)

lspil:
I also see now some other maybe unwanted behavior reported with issues #174 and #175. Maybe we can merge them as well with this solution if you agree. They seem all to be very related.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/InMemoryOAuth2AuthorizationService.java
```diff
@@ -64,6 +64,10 @@ public OAuth2Authorization findByToken(String token, @Nullable TokenType tokenTy
 	}
 
 	private boolean hasToken(OAuth2Authorization authorization, String token, TokenType tokenType) {
+		if (tokenType == null) {
+			return authorization.getTokens().getAccessToken() != null ||
```
jgrandja, line 68: This logic is not correct since it doesn't compare the `token` value.

jgrandja:
@lspil I agree that this issue is related to #174.

I just pushed a fix for #174 via 7f8aff7.

I'll close this as a duplicate.

PR #177, opened 2020-12-10
Add actuator/health endpoints to sample resource and client apps

joshuatcasey:
- This helps with liveness/readiness K8s probes


joshuatcasey:
Since there aren't automated `tests` for this, I tested just by running all three samples and ensured that they still worked, and I was able to load `localhost:8080/actuator/health` and `localhost:8090/actuator/health` and receive the following without any auth.

`{"status":"UP"}`

jgrandja:
Thanks for the PR @joshuatcasey.

The main goal of the sample is to demonstrate the integration between the authorization server, client and resource server. As well, to demonstrate how to configure the features that are currently implemented.

I don't feel adding Actuator to the sample makes sense as I'd like to keep the sample as minimal (and easy) as possible and to only demonstrate OAuth / OIDC specific features / flows.

Based on this, I'm going to close this PR. I appreciate your understanding.

PR #217, opened 2021-02-05
Implement an audit event framework

fhanik:
Please review one suggestion for implementing audit events in SAS.

Currently the only event being transmitted is a _token issued event_ during a client_credentials grant

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientCredentialsAuthenticationProvider.java
@@ -20,6 +20,8 @@ import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
@@ -39,6 +41,7 @@ import org.springframework.security.oauth2.server.authorization.client.Registere
 import org.springframework.security.oauth2.server.authorization.config.ProviderSettings;
 import org.springframework.security.oauth2.server.authorization.JwtEncodingContext;
 import org.springframework.security.oauth2.server.authorization.OAuth2TokenCustomizer;
+import org.springframework.security.oauth2.server.authorization.event.OAuth2ClientCredentialsTokenIssuedEvent;
 import org.springframework.util.Assert;
 import org.springframework.util.CollectionUtils;
 
@@ -59,9 +62,11 @@ import static org.springframework.security.oauth2.server.authorization.authentic
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.4">Section 4.4 Client Credentials Grant</a>
  * @see <a target="_blank" href="https://tools.ietf.org/html/rfc6749#section-4.4.2">Section 4.4.2 Access Token Request</a>
  */
-public class OAuth2ClientCredentialsAuthenticationProvider implements AuthenticationProvider {
+public class OAuth2ClientCredentialsAuthenticationProvider implements AuthenticationProvider,
+		ApplicationEventPublisherAware {
 	private final OAuth2AuthorizationService authorizationService;
 	private final JwtEncoder jwtEncoder;
+	private ApplicationEventPublisher publisher = event -> {};
 	private OAuth2TokenCustomizer<JwtEncodingContext> jwtCustomizer = (context) -> {};
 	private ProviderSettings providerSettings;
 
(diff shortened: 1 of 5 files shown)


jgrandja:
Thanks @fhanik ! I will look at this next week. I'm releasing `0.1.0` this Thursday and need to get a couple of items merged beforehand.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/event/OAuth2AuthorizationEventPublisher.java
```diff
@@ -0,0 +1,40 @@
...
+package org.springframework.security.oauth2.server.authorization.event;
+
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.server.authorization.OAuth2Authorization;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AccessTokenAuthenticationToken;
+
+public interface OAuth2AuthorizationEventPublisher {
+
+	static OAuth2AuthorizationEventPublisher nullPublisher() {
+		return new OAuth2AuthorizationEventPublisher() {
```
AbstractCoderX, line 25: extract this to constant

AbstractCoderX (changes requested):

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/event/DefaultOAuth2AuthorizationEventPublisher.java
```diff
@@ -0,0 +1,43 @@
...
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.event;
+
+import org.springframework.context.ApplicationEventPublisher;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.server.authorization.OAuth2Authorization;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AccessTokenAuthenticationToken;
+
+public final class DefaultOAuth2AuthorizationEventPublisher implements OAuth2AuthorizationEventPublisher {
```
jgrandja, line 23: Please remove this and instead use Spring Security's `DefaultAuthenticationEventPublisher`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/event/OAuth2AuthorizationEvent.java
```diff
@@ -0,0 +1,33 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.event;
+
+import org.springframework.security.access.event.AbstractAuthorizationEvent;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
+
+public abstract class OAuth2AuthorizationEvent extends AbstractAuthorizationEvent {
```
jgrandja, line 21: `AbstractAuthorizationEvent` are events that are fired by `FilterSecurityInterceptor` when authorizing the current request. Instead, we should extend from `AbstractAuthenticationEvent` for authentication related events.

Take a look at the hierarchy of `AbstractAuthenticationEvent` in order to determine where to extend from. The common extensions will be either `AuthenticationSuccessEvent` and `AbstractAuthenticationFailureEvent`.

At this point, I'm not sure if we'll introduce an OAuth2-specific event hiearchy, so for now let's just prefix the event classes with OAuth2.

jgrandja (changes requested):
Thanks for the PR @fhanik. Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/event/DefaultOAuth2AuthorizationEventPublisher.java
fhanik: Removed, there isn't a need to use the Spring Security publisher in this case. Spring Framework creates a publisher that is sufficient, and delegation isn't needed.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/event/OAuth2AuthorizationEvent.java
fhanik: When a token is issued, it isn't an authentication event. As such this feels a bit forced, but I have made the change.

fhanik:
I just noticed that SAS publishes an authentication success event when a token is issued. 

https://github.com/spring-projects-experimental/spring-authorization-server/blob/7652d0ebbe2205aeb9b851b7a41580c1b36c9a00/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java#L157

We can probably just trap this event and rewrite it, rather than publishing a new one.

fhanik:
There is an `AuthenticationSuccess` event thrown by an `AuthenticationManager` when a token is issued. This is sufficient for now as we can translate that event into a token issued event on our side. So from a feature standpoint, we don't need this PR, but the overloading of Spring Security concepts may cause some confusion (or not :) )

Issue #139, opened 2020-10-23
Provide capability for custom configuration


jgrandja:
The goal of this ticket is to capture all the possible configuration options that could require customization based on the requirements of an authorization server deployment.

**NOTE:** This ticket will be on hold until we reach a candidate MVP release, along with stability in the public API's.

The following is the _on-going_ list of possible configuration options:

- Provide plugin mechanism for extension authorization grant
- Configure a custom User Consent page

Clem69B:
Hello ! 

An other proposition : 
> Provide plugin mechanism for Access Token customization (add custom claims in JWT token)

Thanks for this job !

jgrandja:
@Clem69B JWT customization is already available via #173

anoopgarlapati:
@jgrandja I have a couple more customization suggestions:
1. Provide ability for custom `redirect_uri` matching: Currently (release `0.1.0`) the `redirect_uri` from authorization request is matched strictly against the registered values from `RegisteredClient`. However, there is a provision in [OAuth 2.1 draft spec](https://tools.ietf.org/html/draft-ietf-oauth-v2-1-01#section-10.3.3) which allows any port to be specified for redirection on loopback address.
   > The authorization server MUST allow any port to be specified at the time of the request for loopback IP redirect URIs, to accommodate clients that obtain an available ephemeral port from the operating system at the time of the request.
   
   The above is more of a fix in the current in the current redirection matching than a new customization as we are attempting to conform to OAuth 2.1. 
   Also, there are use cases for authorization servers to relax exact string matching for `redirect_uri` in non-production or development environments. For this use case it would be better to allow custom matching of the redirect endpoint.
2. Provide ability to add custom token response parameters: Currently (release `0.1.0`) the token response is constructed from `OAuth2AccessTokenAuthenticationToken` which does have `additionalParameters` which would be written in the response. However, there is no convenient way to add parameters to this `Map` without completely overriding the `AuthenticationProvider` classes. Similar to customizing `Jwt` claims we should be able to add custom `additionalParameters` to the token.

metacubed:
#### Provide a generic mechanism to reduce<sup>++</sup> the granted scopes present in a token

> Moving over #218 as recommended:

For either client- or user-based grants, the list of granted scopes is initially populated with those available to the client. This list is then filtered by other constraints, such as:

- The resource-owner consenting to only a subset of scopes
- The `scope` claim in the token request (#212)
[...]

PR #240, opened 2021-02-15
Introduce RedirectValidator

anoopgarlapati:
- Introduced `RedirectValidator` interface for validation of `redirect_uri`
- Added `DefaultRedirectValidator` - a default implementation for validating
`redirect_uri` as per OAuth 2.1

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/DefaultRedirectValidator.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/DefaultRedirectValidator.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2020-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


anoopgarlapati:
Hi @jgrandja 
I created this draft PR to discuss the redirection issue for loopback addresses I mentioned in https://github.com/spring-projects-experimental/spring-authorization-server/issues/139#issuecomment-779188444 
I introduced the `RedirectValidator` interface so that it would also allow custom configuration later when the instance in `OAuth2AuthorizationEndpointFilter` can be configured.
I am yet to work on tests for this but I would like a review of this approach before proceeding.

jgrandja:
Thanks for the PR @anoopgarlapati !

I'd rather not introduce `RedirectValidator` as it's specific only to one type of validation.

There are a number of validations that occur in the `Filter`'s (`OAuth2AuthorizationEndpointFilter`, `OAuth2TokenEndpointFilter`) the `AuthenticationProvider`'s (`OAuth2AuthorizationCodeAuthenticationProvider`, `OAuth2ClientCredentialsAuthenticationProvider`, etc.) and other areas in the code.

What would be ideal is to come up with an abstraction that would work for all validations.

I'm wondering if we can come up with a design that is similar to `OAuth2TokenValidator` but a higher level of abstraction. This could be the start of the "Policy Enforcement" abstraction #139.

What are your thoughts?

Also, would you be able to submit another PR (or update this one) that fixes (patches) the `redirect_uri` validation so we can get it merged. The design of the Policy Enforcer could take some time.

anoopgarlapati:
Thanks @jgrandja !

Yes, that sounds good to me with incorporating these (redirect validations, etc) in a single place rather than having specific ones. I will try to come up with some suggestions there and weigh in on the discussion.

I am closing this PR and will create another one that fixes the `redirect_uri` validation in the `OAuth2AuthorizationEndpointFilter` itself.

PR #250, opened 2021-03-03
Ensure that the access token response "scope" shows the scopes in the JWT

joshuatcasey:
Since the `scope` claim can be customized, we have to make sure that the access token response `scope` field shows the same scopes as the customized claim.

TODO: If this implementation makes sense I'll work on an accompanying test. This didn't feel like it naturally fit into another test.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
@@ -167,9 +167,11 @@ public class OAuth2AuthorizationCodeAuthenticationProvider implements Authentica
 		JwtClaimsSet claims = context.getClaims().build();
 		Jwt jwtAccessToken = this.jwtEncoder.encode(headers, claims);
 
+		Set<String> customizedScopes = new HashSet<>(claims.getClaimAsStringList(OAuth2ParameterNames.SCOPE));
+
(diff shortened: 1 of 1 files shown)


anoopgarlapati:
@joshuatcasey Please check my comment here https://github.com/spring-projects-experimental/spring-authorization-server/issues/199#issuecomment-776805577 
I think picking the `scope` field for access token response from the `scope` claim in JWT is not a good idea for servers that do not use the `scope` claim. I think it should be in a way that the customized scopes for the claim in the JWT and the access token response are picked up from a common location.

joshuatcasey:
Thanks for the context @anoopgarlapati 

Looks like one of these ways could work:
- The `OAuth2TokenCustomizer<JwtEncodingContext>` `@Bean` could modify the context's `getAuthorizedScopes` but this feels awkward... something like this:

jgrandja:
@joshuatcasey The existing implementation is correct.

The main references in the [spec](https://tools.ietf.org/html/draft-ietf-oauth-v2-1-01#section-3.3) are:

> The authorization and token endpoints allow the client to specify the
scope of the access request using the "scope" request parameter.  In
turn, the authorization server uses the "scope" response parameter to
inform the client of the scope of the access token issued.

> If the issued access token scope
is different from the one requested by the client, the authorization
server MUST include the "scope" response parameter to inform the
client of the actual scope granted.

When the client sends the `scope` parameter in the Authorization Request, it may be granted all, partial or none of the requested scopes depending on the Authorization Consent phase. The accepted/declined or (auto-approved) scopes in the 
Authorization Consent phase become the "Authorized Scopes" and this is what should be returned in the `AccessTokenResponse.scope` (if different from the "requested" `scope).

I hope this make sense?

I'm going to close this PR as existing implementation is correct.

PR #252, opened 2021-03-04
Scope "openid" should be in access token scopes

joshuatcasey:
- Still does not require user consent


jgrandja:
Thanks for the PR @joshuatcasey. This is now in master.

PR #253, opened 2021-03-04
Introduce an OAuth2TokenType for ID_TOKEN

joshuatcasey:


jgrandja:
@joshuatcasey `id_token` is not defined as a token type in [4.1.2 OAuth Token Type Hints Registry](https://tools.ietf.org/html/rfc7009#section-4.1.2).

I'd rather leave this out and only define the standard token types in `OAuth2TokenType`

PR #254, opened 2021-03-05
WIP for feedback: Prototype to remember user consent decisions

joshuatcasey:
A user scope consent decisions can have a lifecycle outside of the context of an authorization code grant. For example, if a user `user` grants scope `message.read` to client `message-client`, this consent decision could live for 7 days, or 30 days, or be revokable by the user.

This is only a prototype to indicate how this might fit in with the current flow of `OAuth2AuthorizationEndpointFilter`.

TODO:
- Remove any extraneous lines (such as `imports` that were moved accidentally)
- Provide a sample `HttpSession`-based implementation of `UserConsentRepository`
- Needs tests!
- `UserConsentRepository` should be a bean that's injected via context, not created inside of `OAuth2AuthorizationEndpointFilter`
- Note that the consent lookups use `subject` and `clientId`. Right now `subject` is the name of the principal, but this is really not a long-term solution. Authentication could be delegated to multiple upstream identity providers, each of which returns the same name for the principal. `clientId` also has problems, for example if that client is deleted, all of the consent decisions related to that client should be revoked else a new client with the same `clientId` will inherit those consent decisions.

Helpful feedback appreciated!

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/consent/InMemoryUserConsentRepository.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/consent/InMemoryUserConsentRepository.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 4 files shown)


jgrandja:
@joshuatcasey As discussed with the team, I'll close this in favour of #280

Issue #267, opened 2021-03-26
Login page should not be configured in OAuth2AuthorizationServerConfigurer


jgrandja:
The login page should not be defaulted in `OAuth2AuthorizationServerConfigurer`.

The custom application configured with `OAuth2AuthorizationServerConfigurer` is responsible for configuring the `AuthenticationEntryPoint`'s.

PR #275, opened 2021-04-01
Fix samples: use loopback IP, merge filter chains

Kehrlann:
Fixes regressions introduced by latest commits:
- `localhost` is disallowed in redirect_uris as per OAuth 2.1, we are now using 127.0.0.1 instead
- Merged the filter chain for spring-authorization-server and the sample app.
- Added simple "integration" tests to avoid regressions.

This is not merge-ready yet.

Notes:
- Trying to navigate to `http://auth-server:9000/oauth2/authorize?response_type=code&client_id=messaging-client&scope=openid&state=SOME_STATE=&redirect_uri=http://localhost:8080/login/oauth2/code/messaging-client-oidc` and logging in gives an error HTTP 999 (??) instead of 400
- `.antMatcher("/error").permitAll()` gives the 400 error back, not sure why...


jgrandja:
Thanks for the PR @Kehrlann. There were a couple of changes required to fix the sample as a result of gh-267.

I applied the fix via 2d8d568.

Closing this as a duplicate.

Kehrlann:
Perfect, thanks!

I ported the tests to their own PR, if you're interested. See #277

PR #276, opened 2021-04-05
Include the remote address in authentication details

joshuatcasey:
We'd like to log the IP address whenever an authN/authZ event happens in Spring Auth Server, such as:

* Client Credentials token issued
* Client authN failure
* ID token issued
* etc

For non-OAuth2/OIDC scenarios such as user login, we rely on the `Authentication::getDetails()` method to return an instance of `WebAuthenticationDetails` that includes the user's remote IP address.

The changes below modifies each `AuthenticationConverter` to add `WebAuthenticationDetails` to the generated `Authentication` object, and modifies each `AuthenticationProvider` to pass through the `WebAuthenticationDetails` to the new `Authentication` object.

If this approach seems reasonable, I'll update this PR with tests.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientSecretBasicAuthenticationConverter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientSecretBasicAuthenticationConverter.java
@@ -16,6 +16,7 @@
 package org.springframework.security.oauth2.server.authorization.web;
 
 import org.springframework.http.HttpHeaders;
+import org.springframework.security.authentication.AuthenticationDetailsSource;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
@@ -23,6 +24,7 @@ import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientAuthenticationToken;
 import org.springframework.security.web.authentication.AuthenticationConverter;
+import org.springframework.security.web.authentication.WebAuthenticationDetailsSource;
 import org.springframework.util.StringUtils;
 
 import javax.servlet.http.HttpServletRequest;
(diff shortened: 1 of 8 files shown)


joshuatcasey:
PR updated with tests

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientSecretBasicAuthenticationConverter.java
```diff
@@ -86,8 +90,10 @@ public Authentication convert(HttpServletRequest request) {
 			throw new OAuth2AuthenticationException(new OAuth2Error(OAuth2ErrorCodes.INVALID_REQUEST), ex);
 		}
 
-		return new OAuth2ClientAuthenticationToken(clientID, clientSecret, ClientAuthenticationMethod.BASIC,
+		OAuth2ClientAuthenticationToken oAuth2ClientAuthenticationToken = new OAuth2ClientAuthenticationToken(clientID, clientSecret, ClientAuthenticationMethod.BASIC,
 				extractAdditionalParameters(request));
+		oAuth2ClientAuthenticationToken.setDetails(authenticationDetailsSource.buildDetails(request));
```
jgrandja, line 95: Instead of updating `ClientSecretBasicAuthenticationConverter`, `ClientSecretPostAuthenticationConverter`, etc. we should update `OAuth2ClientAuthenticationFilter.doFilterInternal()` as such:

```java
Authentication authenticationRequest = this.authenticationConverter.convert(request);
if (authenticationRequest instanceof AbstractAuthenticationToken) {
	((AbstractAuthenticationToken) authenticationRequest).setDetails(
			authenticationDetailsSource.buildDetails(request));
}
```

Please apply the same to `OAuth2TokenEndpointFilter.doFilterInternal()`.

Ultimately, the changes should only be applied in the `Filter`'s.

jgrandja (changes requested):
Thanks for the PR! Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientSecretBasicAuthenticationConverter.java
mamachanko: Thanks for recommending these changes. I've addressed them, hopefully according to your comment.

jgrandja:
Closing in favour of gh-322

PR #277, opened 2021-04-06
Introduce integration tests for the sample oauth server

Kehrlann:


jgrandja:
Thanks for the PR @Kehrlann. This is now in main.

Issue #288, opened 2021-05-10
The authorize request should fail if scope request parameter is not provided


anoopgarlapati:
Currently a client performing authorize request can omit the `scope` parameter and receive an access token without any scopes. This is neither useful nor the expected behavior for many authorization servers. 
As per [RFC 6749 section 3.3](https://datatracker.ietf.org/doc/html/rfc6749#section-3.3 ):
> If the client omits the scope parameter when requesting
   authorization, the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope.  The authorization server SHOULD
   document its scope requirements and default value (if defined).
   
As per the above criterion, the default behavior provided by the project should be changed to reject the request if `scope` parameter is not provided. Ability to customize this default behavior for using a pre-defined value as an alternative behavior suggested in the spec would be provided by custom configuration that would be delivered in #139.

anoopgarlapati:
@jgrandja Yes, I'm interested in submitting a PR for this.

jgrandja:
@anoopgarlapati I've been giving this some further thought and I now feel that we should not add this `scope` validation.

There are use cases where the `scope` parameter will not be used at all and the authorities assigned to the access token will be derived from other sources, e.g. authorities from user, resource indicators #279, custom data source, etc.

If we add this `scope` validation then the above use cases will not work out-of-the-box, which will be very limiting. Those applications will need to customize/disable the `scope` validation. This won't be a good user experience since the validation is quite restrictive.

How can we address these parts in the spec?

> If the client omits the scope parameter when requesting
   authorization, **the authorization server MUST either process the
   request using a pre-defined default value or fail the request
   indicating an invalid scope**.  The authorization server SHOULD
   document its scope requirements and default value (if defined).

Part 1: 

> process the request using a pre-defined default value

This responsibility really belongs in the "product"  and not the "framework". The provider product that leverages the framework will define these kinds of "policies" and document in their reference. Default values are not necessarily statically defined but could be advanced and derived from multiple sources at runtime. Based on this alone, it makes sense that the provider product owns this responsibility. The framework should only provide extension points to allow the product to derive the default scopes, which is already possible via the `OAuth2TokenCustomizer`.

Part 2: 

> fail the request indicating an invalid scope

[...]

Issue #199, opened 2021-01-19
Customizing Jwt claims and headers needs to be more flexible


jgrandja:
The `NimbusJwsEncoder.jwtCustomizer` (#173) needs to be re-designed as it does not easily provide all the context required for customization. For example, it is complicated to obtain the associated `OAuth2Authorization` and/or `RegisteredClient` in order to provide context for token customization.

joshuatcasey:
Glad to see this, @jgrandja ! The `sso4k8s` team hopes to add some of the standard claims to the `id_token` pretty soon.

jgrandja:
@joshuatcasey I'm getting close to a flexible design so hoping to have this merged next week. I'll be looking for your teams feedback on whether the 2nd iteration of the design provides the flexibility you need.

nickmelis:
Excellent work @jgrandja, I too am looking forward to this. While we're here, can I raise a couple of issues I stumbled upon today? 
I noticed that [OAuth2TokenIssuerUtil](https://github.com/spring-projects-experimental/spring-authorization-server/blob/12f4001c9dea449683b86c2a1274e29659f7db9c/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIssuerUtil.java#L69) has the `issuer` and `timeToLive` values hardcoded (with a big TODO next to it), which means there's currently no way to host my auth server on anything different than `http://auth-server:9000`. Is it something you're planning to tackle as part of this big refactoring?

jgrandja:
@nickmelis 

> Is it something you're planning to tackle as part of this big refactoring?

I'm not sure if it will be part of the commit for this ticket. But it will definitely be part of the 0.1.0 (end of month).

joshuatcasey:
@nickmelis in the meantime it's possible to use the [`setJwtCustomizer`](https://github.com/spring-projects-experimental/spring-authorization-server/blob/12f4001c9dea449683b86c2a1274e29659f7db9c/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/jwt/NimbusJwsEncoder.java#L103-L112) method to modify claims after the default claim values have been added.

Something like this works for us:

```java
    @Bean
    public JwtEncoder jwtEncoder(final JWKSource jwkSource) {
        final NimbusJwsEncoder nimbusJwsEncoder = new NimbusJwsEncoder(jwkSource);
[...]

PR #289, opened 2021-05-11
Authorize request should fail if scope request parameter is not provided

anoopgarlapati:
Closes gh-288

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
@@ -349,10 +349,16 @@ public class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilter {
 			return;
 		}
 
-		// scope (OPTIONAL)
+		// scope (REQUIRED)
+		// If the client omits the scope parameter when requesting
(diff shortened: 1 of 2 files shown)


anoopgarlapati:
@wsaca As noted in the issue description #288, authorization server must either reject a request without `scope` parameter or accept the request with a pre-defined set of scopes. The current behavior provided in the project is neither. As a result this PR will change the default behavior to reject such a request which is what providers such as Okta do. 
Also as noted
> Ability to customize this default behavior for using a pre-defined value as an alternative behavior suggested in the spec would be provided by custom configuration that would be delivered in #139.


So, the ability to switch this behavior to use a pre-defined value (based on user profile, client profile, etc) would be provided as customization hook in #139.
@jgrandja Can you please chime in here?

jgrandja:
@wsaca 

> How could we customize the access token to add default scopes? For example, based on the user who was logged in.

The existing `OAuth2TokenCustomizer` (gh-199) would allow you to customize the scopes based on the user who was logged in.

See example configuration:

https://github.com/spring-projects-experimental/spring-authorization-server/blob/93d16d4419440be4b89e0ff92ec267bf26db1a02/oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationCodeGrantTests.java#L381

anoopgarlapati:
Closing as this is not needed anymore. See https://github.com/spring-projects-experimental/spring-authorization-server/issues/288#issuecomment-859567170 in the issue.

PR #291, opened 2021-05-12
JDBC implementation of RegisteredClientRepository

### LINKED ISSUE #265: Provide JDBC implementation of RegisteredClientRepository
This would be useful in real-world workloads where clients are registered and maintained at runtime, without restarting server.

In fact, I've implemented it as I'm evaluating new spring authorization server in my (not-yet-production) projects. It follows  `userdetails.jdbc.JdbcDaoImpl` implementation and only uses Spring `JdbcTemplate`. 

If there are no objections, I would be happy to submit pull request when right time comes.

Current implementation: https://github.com/rlewczuk/spring-authorization-server/tree/gh-265-jdbc-client-repository

Related gh-57 gh-245

rlewczuk:
Code seems to be more-or-less ready for preliminary review, yet I still have some questions in comments below.

Closes gh-265

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
@@ -1,0 +1,335 @@
+/*
+ * Copyright 2020-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.client;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import org.springframework.jdbc.core.*;
+import org.springframework.jdbc.support.lob.DefaultLobHandler;
+import org.springframework.jdbc.support.lob.LobCreator;
+import org.springframework.jdbc.support.lob.LobHandler;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
+import org.springframework.security.oauth2.server.authorization.config.ClientSettings;
+import org.springframework.security.oauth2.server.authorization.config.TokenSettings;
+import org.springframework.util.Assert;
+
+import java.nio.charset.StandardCharsets;
+import java.sql.*;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.*;
+import java.util.function.Function;
+import java.util.stream.Collectors;
+
+/**
+ * JDBC-backed registered client repository
+ *
+ * @author Rafal Lewczuk
+ * @since 0.1.2
+ */
+public class JdbcRegisteredClientRepository implements RegisteredClientRepository {
+
+	private static final Map<String, AuthorizationGrantType> AUTHORIZATION_GRANT_TYPE_MAP;
+	private static final Map<String, ClientAuthenticationMethod> CLIENT_AUTHENTICATION_METHOD_MAP;
+
+	private static final String COLUMN_NAMES = "id, "
+			+ "client_id, "
+			+ "client_id_issued_at, "
+			+ "client_secret, "
+			+ "client_secret_expires_at, "
+			+ "client_name, "
+			+ "client_authentication_methods, "
+			+ "authorization_grant_types, "
+			+ "redirect_uris, "
+			+ "scopes, "
+			+ "client_settings,"
+			+ "token_settings";
+
+	private static final String TABLE_NAME = "oauth2_registered_client";
+
+	private static final String LOAD_REGISTERED_CLIENT_SQL = "SELECT " + COLUMN_NAMES + " FROM " + TABLE_NAME + " WHERE ";
+
+	private static final String INSERT_REGISTERED_CLIENT_SQL = "INSERT INTO " + TABLE_NAME
+			+ "(" + COLUMN_NAMES + ") values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";
+
+	private RowMapper<RegisteredClient> registeredClientRowMapper;
+
+	private Function<RegisteredClient, List<SqlParameterValue>> registeredClientParametersMapper;
+
+	private final JdbcOperations jdbcOperations;
+
+	private final LobHandler lobHandler = new DefaultLobHandler();
+
+	private final ObjectMapper objectMapper;
+
+	public JdbcRegisteredClientRepository(JdbcOperations jdbcOperations, ObjectMapper objectMapper) {
+		Assert.notNull(jdbcOperations, "jdbcOperations cannot be null");
+		Assert.notNull(objectMapper, "objectMapper cannot be null");
+		this.jdbcOperations = jdbcOperations;
+		this.objectMapper = objectMapper;
+		this.registeredClientRowMapper = new DefaultRegisteredClientRowMapper();
+		this.registeredClientParametersMapper = new DefaultRegisteredClientParametersMapper();
+	}
+
+	/**
+	 * Allows changing of {@link RegisteredClient} row mapper implementation
+	 *
+	 * @param registeredClientRowMapper mapper implementation
+	 */
+	public void setRegisteredClientRowMapper(RowMapper<RegisteredClient> registeredClientRowMapper) {
+		Assert.notNull(registeredClientRowMapper, "registeredClientRowMapper cannot be null");
+		this.registeredClientRowMapper = registeredClientRowMapper;
+	}
+
+	/**
+	 * Allows changing of SQL parameter mapper for {@link RegisteredClient}
+	 *
+	 * @param registeredClientParametersMapper mapper implementation
+	 */
+	public void setRegisteredClientParametersMapper(Function<RegisteredClient, List<SqlParameterValue>> registeredClientParametersMapper) {
+		Assert.notNull(registeredClientParametersMapper, "registeredClientParameterMapper cannot be null");
+		this.registeredClientParametersMapper = registeredClientParametersMapper;
+	}
+
+	@Override
+	public void save(RegisteredClient registeredClient) {
+		Assert.notNull(registeredClient, "registeredClient cannot be null");
+		RegisteredClient foundClient = this.findBy("id = ? OR client_id = ? OR client_secret = ?",
+				registeredClient.getId(), registeredClient.getClientId(),
+				registeredClient.getClientSecret().getBytes(StandardCharsets.UTF_8));
+
+		if (null != foundClient) {
+			Assert.isTrue(!foundClient.getId().equals(registeredClient.getId()),
+					"Registered client must be unique. Found duplicate identifier: " + registeredClient.getId());
+			Assert.isTrue(!foundClient.getClientId().equals(registeredClient.getClientId()),
+					"Registered client must be unique. Found duplicate client identifier: " + registeredClient.getClientId());
+			Assert.isTrue(!foundClient.getClientSecret().equals(registeredClient.getClientSecret()),
+					"Registered client must be unique. Found duplicate client secret for identifier: " + registeredClient.getId());
+		}
+
+		List<SqlParameterValue> parameters = this.registeredClientParametersMapper.apply(registeredClient);
+
+		try (LobCreator lobCreator = this.lobHandler.getLobCreator()) {
+			PreparedStatementSetter pss = new LobCreatorArgumentPreparedStatementSetter(lobCreator, parameters.toArray());
+			jdbcOperations.update(INSERT_REGISTERED_CLIENT_SQL, pss);
+		}
+	}
+
+	@Override
+	public RegisteredClient findById(String id) {
+		Assert.hasText(id, "id cannot be empty");
+		return findBy("id = ?", id);
+	}
+
+	@Override
+	public RegisteredClient findByClientId(String clientId) {
+		Assert.hasText(clientId, "clientId cannot be empty");
+		return findBy("client_id = ?", clientId);
+	}
+
+	private RegisteredClient findBy(String condStr, Object...args) {
+		List<RegisteredClient> lst = jdbcOperations.query(
+				LOAD_REGISTERED_CLIENT_SQL + condStr,
+				registeredClientRowMapper, args);
+		return !lst.isEmpty() ? lst.get(0) : null;
+	}
+
+	private class DefaultRegisteredClientRowMapper implements RowMapper<RegisteredClient> {
+
+		private final LobHandler lobHandler = new DefaultLobHandler();
+
+		private Collection<String> parseList(String s) {
+			return s != null ? Arrays.asList(s.split("\\|")) : Collections.emptyList();
+		}
+
+		@Override
+		@SuppressWarnings("unchecked")
+		public RegisteredClient mapRow(ResultSet rs, int rowNum) throws SQLException {
+			Collection<String> scopes = parseList(rs.getString("scopes"));
+			List<AuthorizationGrantType> authGrantTypes = parseList(rs.getString("authorization_grant_types"))
+					.stream().map(AUTHORIZATION_GRANT_TYPE_MAP::get).collect(Collectors.toList());
+			List<ClientAuthenticationMethod> clientAuthMethods = parseList(rs.getString("client_authentication_methods"))
+					.stream().map(CLIENT_AUTHENTICATION_METHOD_MAP::get).collect(Collectors.toList());
+			Collection<String> redirectUris = parseList(rs.getString("redirect_uris"));
+			Timestamp clientIssuedAt = rs.getTimestamp("client_id_issued_at");
+			Timestamp clientSecretExpiresAt = rs.getTimestamp("client_secret_expires_at");
+			byte[] clientSecretBytes = this.lobHandler.getBlobAsBytes(rs, "client_secret");
+			String clientSecret = clientSecretBytes != null ? new String(clientSecretBytes, StandardCharsets.UTF_8) : null;
+			RegisteredClient.Builder builder = RegisteredClient
+					.withId(rs.getString("id"))
+					.clientId(rs.getString("client_id"))
+					.clientIdIssuedAt(clientIssuedAt != null ? clientIssuedAt.toInstant() : null)
+					.clientSecret(clientSecret)
+					.clientSecretExpiresAt(clientSecretExpiresAt != null ? clientSecretExpiresAt.toInstant() : null)
+					.clientName(rs.getString("client_name"))
+					.clientAuthenticationMethods(coll -> coll.addAll(clientAuthMethods))
+					.authorizationGrantTypes(coll -> coll.addAll(authGrantTypes))
+					.redirectUris(coll -> coll.addAll(redirectUris))
+					.scopes(coll -> coll.addAll(scopes));
+
+			RegisteredClient rc = builder.build();
+
+			TokenSettings ts = rc.getTokenSettings();
+			ClientSettings cs = rc.getClientSettings();
+
+			try {
+				String tokenSettingsJson = rs.getString("token_settings");
+				if (tokenSettingsJson != null) {
+
+					Map<String, Object> m = JdbcRegisteredClientRepository.this.objectMapper.readValue(tokenSettingsJson, Map.class);
+
+					Number accessTokenTTL = (Number) m.get("access_token_ttl");
+					if (accessTokenTTL != null) {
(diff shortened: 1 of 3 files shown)


rlewczuk:
Q (testing): provided set of tests mostly copies those from `InMemoryRegisteredClientRepositoryTests`. This leads to considerable amount of code duplication, grouping those tests into common superclass would not only reduce duplication but also create baseline for testing future implementations of `RegisteredClientRepository` (eg. Redis variant). For now I've decided against it, so I don't touch any unrelated code. Should I refactor those into common base ? If so, should it be done as part of this change or as separate one ?

rlewczuk:
Question to maintainers: as RegisteredClient is immutable, there us no update semantics implemented in JdbcRegisteredClientRepository. Is it valid approach ?

rlewczuk:
Question: JdbcClientRepository checks for duplicated client secret. Does it make sens when password encoders are used ?

lanmingle:
你好！

我曾经在使用 mongo 进行适配，我建议使用转换器，这样更灵活。

---

Hello!

I used to use mongo for adaptation, and I recommend using a converter, which is more flexible.

---

## Reference Code

model

``` java

@Getter
@Setter
@ToString
@EqualsAndHashCode(of = {"key"})
public class OidcPersistentKeyValue {

    private String key;

    private Object value;

    public OidcPersistentKeyValue() {
    }

    public OidcPersistentKeyValue(String key, Object value) {
        this.key = key;
        this.value = value;
    }

}

/**
 * @see org.springframework.security.oauth2.server.authorization.config.ClientSettings
 */
@Getter
@Setter
@ToString(callSuper = true)
public class OidcClientSettings extends OidcPersistentKeyValue {

    public static final String CLIENT_SETTING_BASE = "setting.client.";

    public static final String REQUIRE_PROOF_KEY = CLIENT_SETTING_BASE.concat("require-proof-key");

    public static final String REQUIRE_USER_CONSENT = CLIENT_SETTING_BASE.concat("require-user-consent");

    public OidcClientSettings() {
    }

    public OidcClientSettings(String key, Object value) {
        super(key, value);
    }

}

/**
 * @see org.springframework.security.oauth2.server.authorization.config.TokenSettings
 */
@Getter
@Setter
@ToString(callSuper = true)
public class OidcTokenSettings extends OidcPersistentKeyValue {

    public static final String TOKEN_SETTING_BASE = "setting.token.";

    public static final String ACCESS_TOKEN_TIME_TO_LIVE = TOKEN_SETTING_BASE.concat("access-token-time-to-live");

    public static final String REUSE_REFRESH_TOKENS = TOKEN_SETTING_BASE.concat("reuse-refresh-tokens");

    public static final String REFRESH_TOKEN_TIME_TO_LIVE = TOKEN_SETTING_BASE.concat("refresh-token-time-to-live");

    public OidcTokenSettings() {
    }

    public OidcTokenSettings(String key, Object value) {
        super(key, value);
    }

}

/**
 * @see org.springframework.security.oauth2.core.ClientAuthenticationMethod
 */
public enum OidcClientAuthenticationMethod {

    BASIC,

    POST,

    NONE,

}

/**
 * @see org.springframework.security.oauth2.core.AuthorizationGrantType
 */
public enum OidcAuthorizationGrantType {

    AUTHORIZATION_CODE,

//    @Deprecated
//    IMPLICIT,

    REFRESH_TOKEN,

    CLIENT_CREDENTIALS,

    PASSWORD,

}

```

domain

``` java

@Getter
@Setter
@ToString
@EqualsAndHashCode(of = {"id"})
@Document(collection = "OidcRegisteredClient")
@CompoundIndexes({
        @CompoundIndex(def = "{'clientId': 1}", unique = true)
})
public class OidcRegisteredClient {

    @Id
    private String id;

    private String clientId;

    private String clientSecret;

    private Set<OidcClientAuthenticationMethod> clientAuthenticationMethods;

    private Set<OidcAuthorizationGrantType> authorizationGrantTypes;

    private Set<String> redirectUris;

    private Set<String> scopes;

    private Set<OidcClientSettings> clientSettings;

    private Set<OidcTokenSettings> tokenSettings;

}

```

repository

``` java

public interface OidcRegisteredClientRepository extends MongoRepository<OidcRegisteredClient, String> {

    @Query("{ 'clientId': :#{#clientId} }")
    Optional<OidcRegisteredClient> getByClientId(@Param("clientId") String clientId);

}

```

converter

``` java

@Slf4j
@Component
public class RegisteredClientConverter implements Converter<OidcRegisteredClient, RegisteredClient> {

    @Override
    public RegisteredClient convert(OidcRegisteredClient source) {
        return this.toConvert(OidcRegisteredClientBuilder.create(source).build());
    }

    private RegisteredClient toConvert(OidcRegisteredClient merge) {
        return RegisteredClient.withId(merge.getId())
                .clientId(merge.getClientId())
                .clientSecret(merge.getClientSecret())
                .clientAuthenticationMethods((consumer) -> this.toClientAuthenticationMethods(consumer, merge.getClientAuthenticationMethods()))
                .authorizationGrantTypes((consumer) -> this.toAuthorizationGrantTypes(consumer, merge.getAuthorizationGrantTypes()))
                .redirectUris((consumer) -> this.toRedirectUris(consumer, merge.getRedirectUris()))
                .scopes((consumer) -> this.toScopes(consumer, merge.getScopes()))
                .clientSettings((consumer) -> this.toClientSettings(consumer, merge.getClientSettings()))
                .tokenSettings((consumer) -> this.toTokenSettings(consumer, merge.getTokenSettings()))
                .build();
    }

    private void toClientAuthenticationMethods(Set<ClientAuthenticationMethod> sources, Set<OidcClientAuthenticationMethod> others) {
        if (CollUtil.isNotEmpty(others)) {
            others.forEach(source -> sources.add(new ClientAuthenticationMethod(source.name().toLowerCase())));
        }
    }

    private void toAuthorizationGrantTypes(Set<AuthorizationGrantType> sources, Set<OidcAuthorizationGrantType> others) {
        if (CollUtil.isNotEmpty(others)) {
            others.forEach(source -> sources.add(new AuthorizationGrantType(source.name().toLowerCase())));
        }
    }

    private void toRedirectUris(Set<String> sources, Set<String> others) {
        if (CollUtil.isNotEmpty(others)) {
            sources.addAll(others.stream().filter(StrUtil::isNotEmpty).collect(Collectors.toSet()));
        }
    }

    private void toScopes(Set<String> sources, Set<String> others) {
        if (CollUtil.isNotEmpty(others)) {
            sources.addAll(others.stream().filter(StrUtil::isNotEmpty).collect(Collectors.toSet()));
        }
    }

    private void toClientSettings(ClientSettings source, Set<OidcClientSettings> others) {
        if (CollUtil.isNotEmpty(others)) {
            others.forEach(other -> {
                String key = other.getKey();
                Object val = other.getValue();
                if (ClientSettings.REQUIRE_PROOF_KEY.equals(key)) {
                    source.setting(key, Convert.convert(Boolean.class, val, OidcRegisteredClientBuilder.REQUIRE_PROOF_KEY));
                } else if (ClientSettings.REQUIRE_USER_CONSENT.equals(key)) {
                    source.setting(key, Convert.convert(Boolean.class, val, OidcRegisteredClientBuilder.REQUIRE_USER_CONSENT));
                }
            });
        }
    }

    private void toTokenSettings(TokenSettings source, Set<OidcTokenSettings> others) {
        if (CollUtil.isNotEmpty(others)) {
            others.forEach(other -> {
                String key = other.getKey();
                Object val = other.getValue();
                if (TokenSettings.ACCESS_TOKEN_TIME_TO_LIVE.equals(key)) {
                    source.setting(key, Convert.convert(Duration.class, val, OidcRegisteredClientBuilder.ACCESS_TOKEN_TIME_TO_LIVE));
                } else if (TokenSettings.REFRESH_TOKEN_TIME_TO_LIVE.equals(key)) {
                    source.setting(key, Convert.convert(Duration.class, val, OidcRegisteredClientBuilder.REFRESH_TOKEN_TIME_TO_LIVE));
                } else if (TokenSettings.REUSE_REFRESH_TOKENS.equals(key)) {
                    source.setting(key, Convert.convert(Boolean.class, val, OidcRegisteredClientBuilder.REUSE_REFRESH_TOKENS));
                }
            });
        }
    }

}

```

RegisteredClientRepository

``` java

@Slf4j
@Repository
public class DelegateRegisteredClientRepository implements RegisteredClientRepository {

    @Autowired
    private OidcRegisteredClientService registeredClientService;

    @Autowired
    private RegisteredClientConverter registeredClientConverter;

    @Override
    public RegisteredClient findById(String id) {
        return this.registeredClientService.getById(id).map(registeredClient -> this.registeredClientConverter.convert(registeredClient)).orElse(null);
    }

    @Override
    public RegisteredClient findByClientId(String clientId) {
        return this.registeredClientService.getByClientId(clientId).map(registeredClient -> this.registeredClientConverter.convert(registeredClient)).orElse(null);
    }

}

```

jgrandja:
@rlewczuk

FYI, @sjohnr is a new member on our team and he will be handling the review of this PR. We are still targeting this for the 0.1.2 release (Jul 8) so @sjohnr will be providing initial feedback soon.

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/client/oauth2_registered_client_h2.ddl
```diff
@@ -0,0 +1,16 @@
...
+    client_name varchar(512),
+    client_authentication_methods varchar(512) not null,
+    authorization_grant_types varchar(512) not null,
+    redirect_uris varchar(4096) not null,
+    scopes varchar(1024) not null,
+    require_proof_key boolean not null,
+    require_user_consent boolean not null,
+    access_token_ttl integer default 300000 not null,
+    reuse_refresh_tokens boolean default true not null,
+    refresh_token_ttl integer default 600000 not null);
```
sjohnr, line 16: For the settings, please consider using a JSON field, `client_settings varchar(1000) DEFAULT NULL` and `token_settings varchar(1000) DEFAULT NULL`. See #304 for an example of this (e.g. `attributes`).

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+	private static final String LOAD_REGISTERED_CLIENT_SQL = "SELECT " + COLUMN_NAMES + " FROM " + TABLE_NAME + " WHERE ";
+
+	private static final String INSERT_REGISTERED_CLIENT_SQL = "INSERT INTO " + TABLE_NAME
+			+ "(" + COLUMN_NAMES + ") values (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)";
+
+	protected RowMapper<RegisteredClient> registeredClientRowMapper;
+
+	protected Function<RegisteredClient, List<SqlParameterValue>> registeredClientParameterMapper;
+
+	protected Function<String, Collection<String>> listParser;
```
sjohnr, line 72: I think we'd prefer to remove this, or at least hide it as `private` and remove setter for now. This can be fairly easily customized already using the row mapper.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+
+	protected Function<Collection<String>, String> listFormatter;
```
sjohnr, line 74: I think we'd prefer to remove this, or at least hide it as `private` and remove setter for now. This can be fairly easily customized already using the parameters mapper.

sjohnr (changes requested):
Hello @rlewczuk, thanks for your work on this! Just so you know, I'm about to head out of town for the week, but wanted to give a bit of feedback sooner than later. This feedback focuses on the schema and two additional items. I'll do a more thorough review of the code early next week hopefully.

See feedback inline.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+
+	protected JdbcOperations jdbcOperations;
+
+	public JdbcRegisteredClientRepository(JdbcOperations jdbcOperations) {
+		Assert.notNull(jdbcOperations, "jdbcOperations cannot be null");
+		this.jdbcOperations = jdbcOperations;
+		this.registeredClientRowMapper = this::defaultRegisteredClientRowMapper;
+		this.registeredClientParameterMapper = this::defaultRegisteredClientParameterMapper;
```
sjohnr, line 82: I suggest moving this to an inner class that implements `Function<RegisteredClient, List<SqlParameterValue>>`, or alternatively a `private static` method that returns a lambda.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	/**
+	 * Allows changing of SQL parameter mapper for {@link RegisteredClient}
+	 *
+	 * @param registeredClientParameterMapper mapper implementation
+	 */
+	public void setRegisteredClientParameterMapper(Function<RegisteredClient, List<SqlParameterValue>> registeredClientParameterMapper) {
+		Assert.notNull(registeredClientParameterMapper, "registeredClientParameterMapper cannot be null");
+		this.registeredClientParameterMapper = registeredClientParameterMapper;
```
sjohnr, line 104: Please rename field/parameter/setter as `registeredClientParametersMapper` (with an "s").

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+			Assert.isTrue(!foundClient.getId().equals(registeredClient.getId()),
+					"Registered client must be unique. Found duplicate identifier: " + registeredClient.getId());
+			Assert.isTrue(!foundClient.getClientId().equals(registeredClient.getClientId()),
+					"Registered client must be unique. Found duplicate client identifier: " + registeredClient.getClientId());
+			Assert.isTrue(!foundClient.getClientSecret().equals(registeredClient.getClientSecret()),
+					"Registered client must be unique. Found duplicate client secret for identifier: " + registeredClient.getId());
+		}
+
+		List<SqlParameterValue> parameters = this.registeredClientParameterMapper.apply(registeredClient);
+		PreparedStatementSetter pss = new ArgumentPreparedStatementSetter(parameters.toArray());
```
sjohnr, line 143: In conjunction with adding a `lob` value for `client_secret`, this would need to be switched to using a `LobHandler` and `LobCreatorArgumentPreparedStatementSetter`. See [JdbcOAuth2AuthorizedClientService](https://github.com/spring-projects/spring-security/blob/9d2db8983807cbe29b2efb5f6ee3c17e5b05364d/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java#L191-L195) for an example.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+		List<RegisteredClient> lst = jdbcOperations.query(
+				LOAD_REGISTERED_CLIENT_SQL + condStr,
+				registeredClientRowMapper, args);
+		return lst.size() == 1 ? lst.get(0) : null;
+	}
+
+	private RegisteredClient defaultRegisteredClientRowMapper(ResultSet rs, int rownum) throws SQLException {
+		Collection<String> scopes = listParser.apply(rs.getString("scopes"));
+		List<AuthorizationGrantType> authGrantTypes = listParser.apply(rs.getString("authorization_grant_types"))
+				.stream().map(AUTHORIZATION_GRANT_TYPE_MAP::get).collect(Collectors.toList());
```
sjohnr, line 169: Replace with a for-loop. See spring-projects/spring-security#7154.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+		List<ClientAuthenticationMethod> clientAuthMethods = listParser.apply(rs.getString("client_authentication_methods"))
+				.stream().map(CLIENT_AUTHENTICATION_METHOD_MAP::get).collect(Collectors.toList());
```
sjohnr, line 171: Replace with a for-loop. See spring-projects/spring-security#7154.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -0,0 +1,257 @@
...
+	public RegisteredClient findByClientId(String clientId) {
+		Assert.hasText(clientId, "clientId cannot be empty");
+		return findBy("client_id = ?", clientId);
+	}
+
+	private RegisteredClient findBy(String condStr, Object...args) {
+		List<RegisteredClient> lst = jdbcOperations.query(
+				LOAD_REGISTERED_CLIENT_SQL + condStr,
+				registeredClientRowMapper, args);
+		return lst.size() == 1 ? lst.get(0) : null;
```
sjohnr, line 163: Given the validation code that relies on this method, I think this should be:

```
return !lst.isEmpty() ? lst.get(0) : null;
```

sjohnr (changes requested):
@rlewczuk I've added some additional comments and requested a few more changes. Let me know if you have questions or comments. Please see my feedback inline.

rlewczuk:
I'll take care of it next week or so as I'm currently bogged with other project I'm finishing right now. Thanks for review.

oauth2-authorization-server/spring-security-oauth2-authorization-server.gradle
```diff
@@ -10,13 +10,18 @@ dependencies {
 	compile 'com.nimbusds:nimbus-jose-jwt'
 	compile 'com.fasterxml.jackson.core:jackson-databind'
 
+	optional 'org.springframework:spring-jdbc'
+	optional 'org.springframework:spring-tx'
```
sjohnr, line 14: Is this used? I think this should be removed.

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/client/oauth2_registered_client_h2.ddl
```diff
@@ -0,0 +1,16 @@
+create table oauth2_registered_client (
+    id varchar(128) primary key not null,
+    client_id varchar(128) not null,
+    client_id_issued_at datetime,
```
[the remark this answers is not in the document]
rlewczuk: Unfortunately HSQLDB does not support it properly, it ignores default value. So in addition to default value in database, there is additional check in parameters mapper that sets `Instant.now()` timestamp if it is null. It works but it does not look good to me, yet I don't see solution without further complicating things. Let me know if you think it should be further improved.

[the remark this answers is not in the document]
sjohnr: If it's an issue of hsqldb ignoring it, I think that's fine. We still want to include it so we are supporting full database systems that use ANSI standard SQL. Adding code as a fallback is fine, so that it also works with hsqldb. (But hsqldb isn't our target database.)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
rlewczuk: Reimplemented it as described above, yet I don't like it at all. It seems that just for having `client_secret` field of arbitrary length, considerable bunch of complexity was added. In addition local class LobCreatoArgumentPreparedStatement is now duplicated (plus another copy in spring-security proper), so it is probably a candidate to factor out as independent class. Yet it does not solve main problem: do we need all that LOB handling complexity just to skip declaring limit of `client_secret` field ? Change has been pushed, so please decide if I should simplify it or leave as it is.

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/client/oauth2_registered_client_h2.ddl
rlewczuk: Implemented but I have the same remarks as with client_secret LOB. It is more complex both in code and in day-to-day maintenance if application administrator wants to make manual changes in database.

rlewczuk:
Ok, all fixes are here and we're ready for second round.

sjohnr:
Thanks @rlewczuk! This is now in main as 769cf8fac72aa5ed40fddfae82f6cf4f834237b2.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
sjohnr: Feedback received, @rlewczuk, and thanks for being flexible on this one. I will log an issue to reduce code duplication for the setter implementation.

Issue #245, opened 2021-02-25
Provide JDBC implementation of OAuth2AuthorizationService


jgrandja:
We need to provide a JDBC implementation of `OAuth2AuthorizationService` that may be used for production deployments.

The `JdbcOAuth2AuthorizationService` should follow a similar implementation as [JdbcOAuth2AuthorizedClientService](https://github.com/spring-projects/spring-security/blob/42a787d1f6f76d670f0d43797a4321e4563443d4/oauth2/oauth2-client/src/main/java/org/springframework/security/oauth2/client/JdbcOAuth2AuthorizedClientService.java#L70) in the `oauth2-client` side.

lanmingle:
very good !

ovidiupopa91:
Hi @jgrandja . If  this issue is available I would like to work on it.

jgrandja:
Thanks @ovidiupopa91. I'll assign this task to you.

Issue #283, opened 2021-04-27
Provide configuration for custom Authorization Consent page


jgrandja:
We need to provide the capability for an application to configure a custom Authorization Consent page.

PR #304, opened 2021-05-24
Provide JDBC implementation of OAuth2AuthorizationService

ovidiupopa91:
Add new JDBC implementation of the OAuth2AuthorizationService

Closes gh-245


rlewczuk:
Regarding build errors on windows-latest 11/12, make sure you don't use .equals() when comparing Instant objects, instead check for some range, eg. +/0 few milliseconds or so.

ovidiupopa91:
> Regarding build errors on windows-latest 11/12, make sure you don't use .equals() when comparing Instant objects, instead check for some range, eg. +/0 few milliseconds or so.

Hi @rlewczuk . As I'm using the `equals` method from Ouath2Authorization I can't modify it.
 The fix that I implemented was to truncate the Instant when creating the tokens : `Instant.now().truncatedTo(ChronoUnit.MILLIS))`. That's because as of Java 9, `Instant.now()` has nanosecond resolution and the DB Timestamp has millisecond resolution.

ovidiupopa91:
Hi @jgrandja . The implementation is ready for the first round of review.

jgrandja:
@ovidiupopa91 

FYI, @sjohnr is a new member on our team and he will be handling the review of this PR. We are targeting this for the `0.1.2` release (Jul 8) so @sjohnr will be providing initial feedback soon.

ovidiupopa91:
hi @jgrandja / @sjohnr awesome!

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/oauth2-authorization-schema.sql
```diff
@@ -0,0 +1,27 @@
+CREATE TABLE oauth2_authorization (
+    id varchar(100) NOT NULL,
+    registered_client_id varchar(100) NOT NULL,
+    principal_name varchar(100) NOT NULL,
+    authorization_grant_type varchar(100) NOT NULL,
+    attributes varchar(1000) DEFAULT NULL,
+    state blob DEFAULT NULL,
```
sjohnr, line 7: For `state` and `attributes`: I'm wondering if it makes sense for `state` to be a `blob`, but all other attributes to be a `varchar(1000)`. Which flow would require a non-string state that is large enough to overflow 1000 chars? Would all other attributes fit comfortably in JSON format in 1000 characters?

If you have thought through that, awesome! Just thinking through this one myself, I don't have specific recommendations, but I'm wondering what your thoughts are.

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/oauth2-authorization-schema.sql
```diff
@@ -0,0 +1,27 @@
...
+    access_token_type varchar(100) DEFAULT NULL,
+    access_token_scopes varchar(1000) DEFAULT NULL,
+    oidc_id_token_value blob DEFAULT NULL,
+    oidc_id_token_issued_at timestamp DEFAULT NULL,
+    oidc_id_token_expires_at timestamp DEFAULT NULL,
+    oidc_id_token_metadata varchar(1000) DEFAULT NULL,
+    refresh_token_value blob DEFAULT NULL,
+    refresh_token_issued_at timestamp DEFAULT NULL,
+    refresh_token_expires_at timestamp DEFAULT NULL,
+    refresh_token_metadata varchar(1000) DEFAULT NULL,
```
sjohnr, line 25: For all code and token-related fields, this seems like a nice approach, given we want to focus on simplicity. The concern here would be if we ever need to add new token types, or realize we missed an attribute. Given that token types don't change often, and have a fairly fixed format, I think we can afford to stay flat here.

But I wanted to mention that it might be worth considering if there's any risk leaving it flat with all optional values. Pursuing a single-column JSON serialization scheme for tokens would be ideal for flexibility, but adds complexity and a more complicated API for this class. If you feel this ought to hold up well enough, I'll support leaving these columns as is.

sjohnr (changes requested):
Hello @ovidiupopa91, thanks for your work on this! Just so you know, I'm about to head out of town for the week, but wanted to give a bit of feedback sooner than later. This feedback focuses on the schema. I'll do a more thorough review of the code early next week hopefully.

See feedback inline.

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/oauth2-authorization-schema.sql
ovidiupopa91: I think I left the state as blob because it is a base64 encoded string. Not sure if this could change in the future or not, but I see no issue with defining it as a `varchar(1000)`. For the attributes I think we should increase the size to `2000` or even more.  Let me know if this is ok and I will do the change in no time 👍

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/oauth2-authorization-schema.sql
ovidiupopa91: I was think about the JSON serialization solution as well in the beginning. But I decided to follow the same approach as in the `oauth2-client` module from `spring-security`. 

The other reason why I decided to use the flat format here is because we may need to search for the token without knowing the `tokenType`. And with this approach I can search for the token value in DB 
```
"state = ? OR authorization_code_value = ? OR access_token_value = ? OR refresh_token_value = ?"
```

If we serialize the token object as JSON, I think it will make it harder to search for a token value as above.

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/oauth2-authorization-schema.sql
sjohnr: Hi @ovidiupopa91, I think the `attributes` is fine, unless we know for sure we'll easily move past the 1000 limit. I suggest leaving it. Users can always increased the size suggested by this template. But I am wondering if we want to leave `state` as a blob, given it's a string (even base64 encoded). Perhaps switch it to `varchar(1000)` as you mentioned?

sjohnr (approved):
Great job!

ovidiupopa91:
@sjohnr awesome, thank you! 

Let me know when there will be any planned feature/task that I could start looking at (for the next release)

jgrandja:
Thank you @ovidiupopa91 !

Would you be interested in adding the JDBC implementation for the recently added `OAuth2AuthorizationConsentService` in gh-283 ?

jgrandja:
@ovidiupopa91 I just logged gh-313. Can you ask for it on there so I can assign to you. Thanks!

PR #315, opened 2021-06-17
Reduce visibility of mapper implementations

sjohnr:
Polish gh-304


sjohnr:
@jgrandja @ovidiupopa91 - Looking at @rlewczuk's PR #291, those mappers are private, and I was thinking it seems like a better starting point than `public`. Any thoughts?

ovidiupopa91:
Hi @sjohnr . For me making the mappers `private` makes sense. 
The main reason why I left them public is that some time ago when I submitted a PR in spring-security for `R2dbcReactiveOAuth2AuthorizedClientService ` I received a comment saying that I should declare the mappers as `public` not `private`. But I don't remember the exact reason. I'm not sure if the reason was to follow the approach from `JdbcOauth2AuthorizedClientService` (where the mappers are declared as public as well) or something else.

sjohnr:
Thanks for the clarification @ovidiupopa91. Just trying to figure out which way we want to go with the initial release of these components, which is why this is just a draft (idea). Will keep you posted as we discuss more.

Issue #281, opened 2021-04-20
Oauth2 Client expects "expires_in" to be a number


roll57:
**Describe the bug**
The field expires_in from the POST /oauth2/token is a string, whereas some system expect a number (example Webswing oidc module) which is using [google-oauth-java-client](https://github.com/googleapis/google-oauth-java-client/blob/267b6f462722b5b610857259855036c182e3cb23/google-oauth-client/src/main/java/com/google/api/client/auth/oauth2/TokenResponse.java)

**To Reproduce**
```
public final class OAuth2AccessTokenResponseMapConverter ... {
	@Override
	public Map<String, String> convert(OAuth2AccessTokenResponse tokenResponse) {
		...
		parameters.put(OAuth2ParameterNames.EXPIRES_IN, String.valueOf(getExpiresIn(tokenResponse)));
		...
	}
}

```

**Expected behavior**
The field should be a number, in the [rfc6749](https://tools.ietf.org/html/rfc6749#section-5.1) we can see that the example is using a number instead of a string

```
   {
     "access_token":"2YotnFZFEjr1zCsicMWpAA",
     "token_type":"example",
     "expires_in":3600,
     "refresh_token":"tGzv3JOkF0XG5Qx2TlKWIA",
     "example_parameter":"example_value"
   }
```

jgrandja:
Thanks for the report @roll57. I logged an issue in Spring Security as this needs to be fixed there. See [spring-security#9685](https://github.com/spring-projects/spring-security/issues/9685).

We might be able to get a temporary fix in here until it's fixed there.

PR #321, opened 2021-06-23
Temporarily fix expires_in for access token response

sjohnr:
**Note:** This can be reverted when Spring Security 5.6 is released.

Closes gh-281


jgrandja:
Thanks for the PR @sjohnr ! This is now in main.

PR #322, opened 2021-06-23
Include WebAuthenticationDetails in token requests

Kehrlann:
Supersedes #276 , it has been reworked according to PR feedback.


jgrandja:
Thanks for the PR @Kehrlann ! This is now in main.

Issue #326, opened 2021-06-25
JdbcRegisteredClientRepository throws NPE when saving public client


sjohnr:
Saving a `RegisteredClient` with null `clientSecret` (e.g. a public client using PKCE) throws `NullPointerException`.

sjohnr:
Related gh-324

Issue #324, opened 2021-06-25
OAuth2AuthorizationCodeAuthenticationProvider does not properly deserialize OAuth2Authorization object attributes


ddubson:
**Describe the bug**

When using the JDBC implementation for `OAuth2AuthorizationService`, the `OAuth2AuthorizationCodeAuthenticationProvider` is not able to properly deserialize the stored attributes within `OAuth2Authorization` object upon authentication. Specifically, the attributes within OAuth2Authorization object are stored as `LinkedHashMap`s, and when it comes time for `OAuth2AuthorizationCodeAuthenticationProvider.authenticate` method to use the attributes, a `ClassCastException` is thrown due to the fact that a LinkedHashMap is not castable to a specific object, such as an `OAuth2AuthorizationRequest` (one of the attributes). All attributes suffer from this casting issue.

**To Reproduce**

- Enable JDBC implementation of at least `OAuth2AuthorizationService`. Ideally, all services should be JDBC-based. 
- Set up H2 In-memory database, and load schema as defined in `src/main/resources`.
- Launch the authorization server, sample client, and sample resource server
- Navigate to sample client (http://localhost:8080)
- Get redirected to auth server, sign in with `user1`/`password`, you will be redirected back.
- Observe error: `too many redirects` in browser. `ClassCastException` on server console.

Should be similar to the following:

```java
java.lang.ClassCastException: java.util.LinkedHashMap incompatible with org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest
	at org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeAuthenticationProvider.authenticate(OAuth2AuthorizationCodeAuthenticationProvider.java:121) ~[main/:na]
```

**Expected behavior**

- Enable JDBC implementation of at least `OAuth2AuthorizationService`. Ideally, all services should be JDBC-based.
- Launch the authorization server, sample client, and sample resource server
- Navigate to sample client
- Get redirected to auth server, sign in with `user1`/`password`, you will be redirected back.
- Observe resources fetched from resource server (no errors)

**Sample**

[Commit that replicates the bug](https://github.com/pivotal/spring-authorization-server/commit/137da46cd02840f3cc504fdb7464224951c56748)

jgrandja:
Thanks for the report @ddubson. We discovered the bug as well and are currently working on a fix.

harrycjy1:
[...]

PR #327, opened 2021-06-25
Fix NPE saving public client

sjohnr:
Closes gh-326
Related gh-324

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
@@ -15,8 +15,6 @@
  */
 package org.springframework.security.oauth2.server.authorization.client;
 
-import java.nio.charset.StandardCharsets;
-import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Timestamp;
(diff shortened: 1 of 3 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
```diff
@@ -150,9 +150,15 @@ public void setRegisteredClientParametersMapper(Function<RegisteredClient, List<
 	@Override
 	public void save(RegisteredClient registeredClient) {
 		Assert.notNull(registeredClient, "registeredClient cannot be null");
-		RegisteredClient foundClient = this.findBy("id = ? OR client_id = ? OR client_secret = ?",
-				registeredClient.getId(), registeredClient.getClientId(),
-				registeredClient.getClientSecret().getBytes(StandardCharsets.UTF_8));
+		RegisteredClient foundClient;
+		if (registeredClient.getClientSecret() != null) {
+			foundClient = this.findBy("id = ? OR client_id = ? OR client_secret = ?",
```
jgrandja, line 155: `RegisteredClient.id` must be unique so the query could simply use `id`.

Also, `RegisteredClient.clientId` must also be unique so you can use either of the two.

oauth2-authorization-server/src/main/resources/org/springframework/security/oauth2/server/authorization/client/oauth2-registered-client.sql
```diff
@@ -2,7 +2,7 @@ CREATE TABLE oauth2_registered_client (
     id varchar(100) NOT NULL,
     client_id varchar(100) NOT NULL,
     client_id_issued_at timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL,
-    client_secret blob NOT NULL,
+    client_secret blob DEFAULT NULL,
```
jgrandja, line 5: Let's change `client_secret` to varchar(200)

jgrandja (changes requested):
Thanks for the PR @sjohnr.

Please apply the requested changes and then go ahead and merge.

PR #341, opened 2021-07-06
support password grant

Leegecho:
support password grant


Leegecho:
the scope is not accurate

jgrandja:
@Leegecho See [comment](https://github.com/spring-projects-experimental/spring-authorization-server/pull/115#issuecomment-696009503)

Issue #346, opened 2021-07-13
Cannot request access token for client with CLIENT_SECRET_BASIC


jrmcdonald:
**Describe the bug**
The [ClientSecretBasicAuthenticationConverter#L89](https://github.com/spring-projects-experimental/spring-authorization-server/blob/77208f05d91210b156c2ce4475793c7d0a0f2c33/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/ClientSecretBasicAuthenticationConverter.java#L89) uses the deprecated ClientAuthenticationMethod.BASIC. This means that if you have a registered client with the ClientAuthenticationMethod.CLIENT_SECRET_BASIC, you cannot request an access token as the check in [OAuth2ClientAuthenticationProvider.java#L103](https://github.com/spring-projects-experimental/spring-authorization-server/blob/93d16d4419440be4b89e0ff92ec267bf26db1a02/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationProvider.java#L103) fails.

**To Reproduce**
1. Run the sample authorization server in this repository
2. Request an access token:

```
curl --location --request POST 'http://localhost:9000/oauth2/token' \
--header 'Content-Type: application/x-www-form-urlencoded' \
--header 'Authorization: Basic bWVzc2FnaW5nLWNsaWVudDpzZWNyZXQ=' \
--data-urlencode 'grant_type=client_credentials' \
--data-urlencode 'scope=message.read'
```

3. Observe successful response
4. Modify the registered client authentication method to ClientAuthenticationMethod.CLIENT_SECRET_BASIC
5. Repeat the above request
6. Observe 401 response

**Expected behavior**
Response should be 200 with a valid access token.

sjohnr:
@jrmcdonald, thanks for the bug report! As you can see, a community member is already working on this.

jrmcdonald:
Hi @sjohnr, thanks for that, do you know which milestone this is likely to be included in?

jgrandja:
@jrmcdonald This is now merged and ready for `0.2.0`.

PR #350, opened 2021-07-14
Remove use of deprecated ClientAuthenticationMethod's

bibibiu2017:
Fixes gh-346

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java
@@ -486,6 +486,7 @@ public class RegisteredClient implements Serializable {
 			}
 			validateScopes();
 			validateRedirectUris();
+			upgradeClientAuthenticationMethods();
 			return create();
 		}
 
@@ -541,6 +542,17 @@ public class RegisteredClient implements Serializable {
 			for (String redirectUri : redirectUris) {
 				Assert.isTrue(validateRedirectUri(redirectUri),
 						"redirect_uri \"" + redirectUri + "\" is not a valid redirect URI or contains fragment");
+			}
+		}
+
+		private void upgradeClientAuthenticationMethods() {
+			if (this.clientAuthenticationMethods.contains(ClientAuthenticationMethod.BASIC)) {
+				this.clientAuthenticationMethods.remove(ClientAuthenticationMethod.BASIC);
+				this.clientAuthenticationMethods.add(ClientAuthenticationMethod.CLIENT_SECRET_BASIC);
+			}
+			if (this.clientAuthenticationMethods.contains(ClientAuthenticationMethod.POST)) {
+				this.clientAuthenticationMethods.remove(ClientAuthenticationMethod.POST);
+				this.clientAuthenticationMethods.add(ClientAuthenticationMethod.CLIENT_SECRET_POST);
 			}
 		}
 
--- a/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClientTests.java
+++ b/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClientTests.java
@@ -43,7 +43,7 @@ public class RegisteredClientTests {
 	private static final Set<String> SCOPES = Collections.unmodifiableSet(
 			Stream.of("openid", "profile", "email").collect(Collectors.toSet()));
 	private static final Set<ClientAuthenticationMethod> CLIENT_AUTHENTICATION_METHODS =
-			Collections.singleton(ClientAuthenticationMethod.BASIC);
+			Collections.singleton(ClientAuthenticationMethod.CLIENT_SECRET_BASIC);
 
 	@Test
 	public void buildWhenAuthorizationGrantTypesNotSetThenThrowIllegalArgumentException() {
@@ -146,7 +146,7 @@ public class RegisteredClientTests {
 				.build();
 
 		assertThat(registration.getClientAuthenticationMethods())
-				.isEqualTo(Collections.singleton(ClientAuthenticationMethod.BASIC));
+				.isEqualTo(Collections.singleton(ClientAuthenticationMethod.CLIENT_SECRET_BASIC));
 	}
 
 	@Test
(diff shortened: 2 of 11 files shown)


sjohnr:
Hi @bibibiu2017, thanks for your interest in the project!

Unfortunately, this change has broken quite a lot of tests. Any chance you can take a look at those? Looking at the tests may highlight whether this alone is the correct change to make or not. I'm still catching up on the issue myself.

Also, please refer to [contributing guidelines](https://github.com/spring-projects-experimental/spring-authorization-server/blob/main/CONTRIBUTING.adoc) when submitting pull requests. In particular, you will want to update the header of files updated with the year "2021" in the copyright, squash your commits, and include "Fixes gh-346" on a separate line in your commit comment.

bibibiu2017:
I have taken a look at the tests, the configured test registered clients, and some assertions are using the deprecated POST and BASIC Client authentication methods.

sjohnr:
Thanks for looking at the tests, @bibibiu2017! I've tested on your branch, and while the immediate issue is resolved with this change, the sample project no longer works. This is because we've flipped the problem so now the deprecated client authentication methods are not allowed.

This is a difficult issue to solve for without too many changes, due to the fact that the `AuthenticationConverter`s being used aren't aware of the `RegisteredClient` which requests what method(s) should be supported, as that later runs into a conflict in [OAuth2ClientAuthenticationProvider.java#L103](https://github.com/spring-projects-experimental/spring-authorization-server/blob/687f03f047b693ee7ae6d885f46b63acefad5b86/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2ClientAuthenticationProvider.java#L103).

Of the available options, I think we can either:
1. Automatically add `CLIENT_SECRET_BASIC` or `CLIENT_SECRET_POST` when a client is registered with `BASIC` or `POST` respectively, or
2. Automatically upgrade `BASIC` or `POST` to `CLIENT_SECRET_BASIC` or `CLIENT_SECRET_POST` respectively when a client is registered.

Both options could have an impact. Would you be able to attempt option 2? One approach could be limiting changes to `RegisteredClient.Builder#clientAuthenticationMethod`, `RegisteredClient.Builder#clientAuthenticationMethods` and tests.

bibibiu2017:
I have attempted option 2. Though I could not localize impact to the recommended test files due to assertions that check equality with deprecated `BASIC` and `POST`. One recommendation would be to change overridden  `equals()` method of `ClientAuthenticationMethod`  making `BASIC` or `POST` equivalent to `CLIENT_SECRET_BASIC ` or `CLIENT_SECRET_POST ` respectively.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java
```diff
@@ -498,8 +490,9 @@ private RegisteredClient create() {
 			registeredClient.clientSecret = this.clientSecret;
 			registeredClient.clientSecretExpiresAt = this.clientSecretExpiresAt;
 			registeredClient.clientName = this.clientName;
-			registeredClient.clientAuthenticationMethods = Collections.unmodifiableSet(
-					new HashSet<>(this.clientAuthenticationMethods));
+			registeredClient.clientAuthenticationMethods = this.clientAuthenticationMethods.stream()
+					.map(this::upgradeDeprecatedAuthenticationMethod)
+					.collect(Collectors.collectingAndThen(Collectors.toCollection(HashSet::new), Collections::unmodifiableSet));
```
sjohnr, line 495: This is really nice! Unfortunately, we will want to [avoid the use of streams](https://github.com/spring-projects/spring-security/issues/9642).

Perhaps a new private method called from inside `build()` similar to `validateScopes()` and `validateRedirectUris()`:

```java
		private void upgradeClientAuthenticationMethods() {
			if (this.clientAuthenticationMethods.contains(ClientAuthenticationMethod.BASIC)) {
				this.clientAuthenticationMethods.remove(ClientAuthenticationMethod.BASIC);
				this.clientAuthenticationMethods.add(ClientAuthenticationMethod.CLIENT_SECRET_BASIC);
			}
			if (this.clientAuthenticationMethods.contains(ClientAuthenticationMethod.POST)) {
				this.clientAuthenticationMethods.remove(ClientAuthenticationMethod.POST);
				this.clientAuthenticationMethods.add(ClientAuthenticationMethod.CLIENT_SECRET_POST);
			}
		}
```

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java
```diff
@@ -15,25 +15,17 @@
...
 
+import org.springframework.lang.Nullable;
+import org.springframework.security.oauth2.core.*;
+import org.springframework.security.oauth2.server.authorization.config.*;
+import org.springframework.util.*;
+
 import java.io.Serializable;
-import java.net.URI;
-import java.net.URISyntaxException;
+import java.net.*;
```
sjohnr, line 24: Your IDE may be too aggressively condensing imports. Wildcards shouldn't be used.

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClientTests.java
```diff
@@ -291,7 +291,7 @@ public void buildWhenTwoClientAuthenticationMethodsAreProvidedThenBothAreRegiste
 				.build();
 
 		assertThat(registration.getClientAuthenticationMethods())
-				.containsExactlyInAnyOrder(ClientAuthenticationMethod.BASIC, ClientAuthenticationMethod.POST);
+				.containsExactlyInAnyOrder(ClientAuthenticationMethod.CLIENT_SECRET_BASIC, ClientAuthenticationMethod.CLIENT_SECRET_POST);
```
sjohnr, line 294: I think we should add an additional test here that uses the non-deprecated methods, but retain this one to test the upgrade process. Perhaps this one could be renamed to indicate it's testing the upgrade. `buildWhenBothDeprecatedClientAuthenticationMethodsAreProvidedThenBothNonDeprecatedAreRegistered`

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClientTests.java
```diff
@@ -309,7 +309,7 @@ public void buildWhenClientAuthenticationMethodsConsumerIsProvidedThenConsumerAc
...
```
sjohnr, line 312: I think we should add an additional test here that uses the non-deprecated methods, but retain this one to test the upgrade process. Perhaps this one could be renamed to indicate it's testing the upgrade. `buildWhenConsumerAddsDeprecatedClientAuthenticationMethodsThenNonDeprecatedAreRegistered`.

sjohnr (commented):
@bibibiu2017, sorry I should have been more clear, I meant "limiting **additional** changes outside of tests" and I think you have done that! I was just thinking of a slightly different approach than yours, but they work out to the same thing. So this seems like a great option!

See inline feedback below. Just a note I'm not going to merge this right away, as I'd like to have a discussion with our community maintainer first, but if we can get it ready for merge, then we'll be in really good shape.

sjohnr:
Also, I forgot to mention, make sure your commit message contains on a separate line:

"Closes gh-346"

bibibiu2017:
@sjohnr thanks for the feedback will be working on this shortly.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java
bibibiu2017: I have removed the stream api

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClient.java
bibibiu2017: I have removed world card imports

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClientTests.java
bibibiu2017: I have added the extra test `buildWhenConsumerAddsDeprecatedClientAuthenticationMethodsThenNonDeprecatedAreRegistered `

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/RegisteredClientTests.java
bibibiu2017: I have added the extra test `buildWhenBothDeprecatedClientAuthenticationMethodsAreProvidedThenBothNonDeprecatedAreRegistered `

jgrandja:
Thanks for all the updates @bibibiu2017 ! This is now in main.

I also added a polish commit that ensures all usages of `ClientAuthenticationMethod.BASIC` and `ClientAuthenticationMethod.POST` are no longer used.

Issue #356, opened 2021-07-21
Add update support in JdbcRegisteredClientRepository 


ovidiupopa91:
`JdbcRegisteredClientRepository.save()` it currently supports only inserts. With this feature, `RegisteredClient` updates should also be possible

ovidiupopa91:
hi @jgrandja . I can work on this task.

PR #365, opened 2021-07-22
Add update support in JdbcRegisteredClientRepository

ovidiupopa91:
Closes gh-356

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
@@ -56,6 +56,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Rafal Lewczuk
  * @author Joe Grandja
+ * @author Ovidiu Popa
  * @since 0.1.2
  * @see RegisteredClientRepository
  * @see RegisteredClient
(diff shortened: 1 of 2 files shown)


ovidiupopa91:
let me know if any of the `RegisteredClient` properties should be excluded from the update sql script.

jgrandja (changes requested):
Thanks for the PR @ovidiupopa91 ! Please see review comments.

jgrandja:
@ovidiupopa91 I just realized that we need to hash `client_secret` on `save()`. You will notice that `OAuth2ClientAuthenticationProvider` uses a `PasswordEncoder` when matching on `client_secret`.

Do you have time / interest in adding `JdbcRegisteredClientRepository.RegisteredClientParametersMapper.setPasswordEncoder()` in a separate PR ?

ovidiupopa91:
> @ovidiupopa91 I just realized that we need to hash `client_secret` on `save()`. You will notice that `OAuth2ClientAuthenticationProvider` uses a `PasswordEncoder` when matching on `client_secret`.
> 
> Do you have time / interest in adding `JdbcRegisteredClientRepository.RegisteredClientParametersMapper.setPasswordEncoder()` in a separate PR ?

Hi @jgrandja . Sure, I will do it.

ovidiupopa91:
@jgrandja  I've raised gh-378 for the other task.

jgrandja:
Thanks for the updates @ovidiupopa91 ! This is now in main.

Issue #378, opened 2021-07-28
Hash RegisteredClient client_secret on save


ovidiupopa91:
Set `PasswordEncoder` in `JdbcRegisteredClientRepository.RegisteredClientParametersMapper` and use it to hash the `client_secret` before saving it.

Issue #355, opened 2021-07-21
Implement Client Configuration Endpoint


ovidiupopa91:
This issue will deliver [ClientConfigurationEndpoint](https://openid.net/specs/openid-connect-registration-1_0.html#ClientConfigurationEndpoint)

ovidiupopa91:
hi @jgrandja . I can work on this feature.

PR #381, opened 2021-07-29
JdbcRegisteredClientRepository hashes client secret on save

ovidiupopa91:
Closes gh-378

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
@@ -36,6 +36,8 @@ import org.springframework.jdbc.core.JdbcOperations;
 import org.springframework.jdbc.core.PreparedStatementSetter;
 import org.springframework.jdbc.core.RowMapper;
 import org.springframework.jdbc.core.SqlParameterValue;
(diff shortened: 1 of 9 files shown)


ovidiupopa91:
still in draft as I will rebase it on `main` after the changes done in gh-356 are merged.

jgrandja:
Thanks for the PR @ovidiupopa91 ! This is now in main.

jgrandja:
@ovidiupopa91 I think there may be a double-encoding bug with this update. I decided to merge anyway but can you confirm if there is a bug or not.

In the scenario where an existing client is updated via `save()`, the `client-secret` might be double encoded depending on how the backing `PasswordEncoder` is implemented. This would result in a failed client authentication on a subsequent grant flow. Can you please confirm if this is an issue?

FYI, I'm on PTO all of next week and returning Aug 9 so I'll follow up with you then.

Thanks again for all your help !

ovidiupopa91:
Hi @jgrandja . I was thinking about this as well. I will let you know when I have a definitive answer.
Enhoy your time off 😀

jgrandja:
Thanks for confirming @ovidiupopa91.

Would you mind creating a new issue for this with the details?

We'll need to have this fixed before we implement gh-355.

ovidiupopa91:
welcome back @jgrandja . Issue created: gh-389

jgrandja:
Thanks @ovidiupopa91 ! I assigned it to you but this can wait until `0.2.1`.

PR #388, opened 2021-08-09
Sample Authorization Server Switch to in memory db - fixes #385 

### LINKED ISSUE #385: The JDBC-based sample code does not work properly
**Describe the bug**
When I run `'authorizationServer'`, `'resourceserver'` and `'client'` in the sample code, I found that I still jump back to the login page even after I log in with the correct account and password. But when I run the `'authorizationserver-custom-consent-page'`, the authorization process works correctly.

mikee:
Use in memorydb for client repository to match the custom consent form sample - fixes failure to call to token endpoint that is returning a 302 to the login form.


mikee:
#385 - replace sample jdbc code with in memory.  If this isn't preferred I'm happy to work out what is wrong with the jdbc sample itself

jgrandja:
Thanks for the PR @mikee.

However, we want the sample to use the JDBC-backed repository.

The main issue has been fixed via ea7c689

PR #396, opened 2021-08-13
Polish loopback address validation in DefaultRedirectUriOAuth2AuthenticationValidator

anoopgarlapati:
Changed loopback address validation from regex to explicit validation using IPv4 loopback address range and IPv6 address.


jgrandja:
Thanks for the update @anoopgarlapati ! This is now in main.

PR #421, opened 2021-08-29
Removed an empty statement

kiratkumar47:


jgrandja:
Thanks for the update @kiratkumar47. This is now in main.

kiratkumar47:
> Thanks for the update @kiratkumar47. This is now in main.

Thanks @jgrandja

Issue #389, opened 2021-08-09
Client secret double encoding issue when updating an existing registered client


ovidiupopa91:
When an existing client is updated via `save()`, the client-secret is double encoded. 

The `client_secret` should be checked if its already encoded or not. If its already encoded, it should not be encoded again.

ovidiupopa91:
hi @jgrandja . I was on PTO last 2 weeks. I am back now, ready for work. 

I have one question: is the `upgradeEncoding` method from the `PasswordEncoder` enough to check if the presented password should be encoded or not? 
As far as I am aware, there's no straightforward way to check if a value is encoded or not.

jgrandja:
Hey @ovidiupopa91. I'll get back to you later this week. I'm just in the middle of preparing for our SpringOne presentation. Thanks.

ovidiupopa91:
Good luck with your presentation @jgrandja . I am sure it will be great 😃

jgrandja:
@ovidiupopa91 

> is the `upgradeEncoding` method from the `PasswordEncoder` enough to check if the presented password should be encoded or not?

No, this wouldn't be the method to call to determine if it's encoded.

Take a look at `DelegatingPasswordEncoder.encode()` as it formats the encoding with a prefix as such:

`{bcrypt}$2a$10$x27xSzWVIdmj9bOTHk2STeQCGIiPZ0o2kOX1ewr.aFUOe2ohbnfxa`

This indicates that it's bcrypt encoded.

I think we could simply check for `{ }` to determine if it's already encoded.

FYI, I'm on PTO starting tomorrow and returning Sep 15.

Talk to you when I'm back.

PR #433, opened 2021-09-07
Client secret double encoding issue when updating an existing registered client

ovidiupopa91:
Avoid client secret double encoding when updating a registered client

Closes gh-389

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/client/JdbcRegisteredClientRepository.java
@@ -95,8 +95,7 @@ public class JdbcRegisteredClientRepository implements RegisteredClientRepositor
 
 	// @formatter:off
 	private static final String UPDATE_REGISTERED_CLIENT_SQL = "UPDATE " + TABLE_NAME
-			+ " SET client_secret = ?, client_secret_expires_at = ?,"
-			+ " client_name = ?, client_authentication_methods = ?, authorization_grant_types = ?,"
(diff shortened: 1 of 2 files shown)


ovidiupopa91:
hi @jgrandja. When [10273](https://github.com/spring-projects/spring-security/issues/10273) will be released in the spring-security project the implementation will no longer be valid as the `prefix` and `suffix` can be changed =>  checking for `{ }` to determine if it's already encoded will no longer be valid.

jgrandja:
Hi @ovidiupopa91. I realized shortly after I proposed the fix that it wasn't a robust solution and that we needed an alternative solution. And yes, 10273 just validated that.

I'm catching up with things this week after returning from PTO so please give me until next week to propose a more robust solution. I'll also review your other PR later next week. Thanks.

jgrandja:
Hi @ovidiupopa91. This fix is a bit tricky. However, there is one easy fix that can be applied in order to avoid the double encoding.

We currently don't allow updates for `id`, `client_id` and `client_id_issued_at`. We could also disable updates for `client_secret` and `client_secret_expires_at`. This would at least resolve the double encoding, which is the main issue we need to address.

I realize this limits the update capabilities, but applications can still customize based on their requirements. We might have to revisit this at a later point but I would say let's go with disabling updates for `client_secret` and `client_secret_expires_at`.

ovidiupopa91:
Hi @jgrandja . Ok, I will revert everything that I did until now and I will remove `client_secret` and `client_secret_expires_at` from the update statement.

jgrandja:
Thanks for the updates @ovidiupopa91  ! This is now in main.
FYI, I added minor polish to the test.
Thanks!

Shabin:
Hi, is there any plan for allowing the update of `client_secret_expires_at`? I have a requirement to extend the client validity.

PR #452, opened 2021-10-05
Simplify OAuth2ClientAuthenticationProvider PKCE verifications

Kehrlann:
In `OAuth2ClientAuthenticationProvider`, if no authentication is available, codeVerifier is REQUIRED, and so is codeChallenge. The value of `requireProofKey` has no influence. Currently, if you set it to "false" in the `authenticateWhenPkceAndRequireProofKeyAndMissingCodeChallengeThenThrowOAuth2AuthenticationException` test, the result will be the same.

Simplified the code to match the current expectations.

`requireProofKey` is still a key part of 
`OAuth2AuthorizationCodeRequestAuthenticationProvider#authenticateAuthorizationRequest`, it requires the presence of a `codeChallenge`.


Kehrlann:
Closed in favor of discussion in #453

Issue #456, opened 2021-10-08
OAuth2AuthorizationCodeAuthenticationProvider Failed to inject ProviderSettings


jymot:
**Describe the bug**
`OAuth2AuthorizationCodeAuthenticationProvider` cannot inject `ProviderSettings`
```java
	@Autowired(required = false)
	protected void setProviderSettings(ProviderSettings providerSettings) {
		this.providerSettings = providerSettings;
	}
```
`OAuth2AuthorizationCodeAuthenticationProvider` is instanced in `OAuth2TokenEndpointConfigurer` and is not injected into the container.

**Expected behavior**
Hope that after `OAuth2TokenEndpointConfigurer` instance `OAuth2AuthorizationCodeAuthenticationProvider`, manually call setProviderSettings to set ProviderSettings.
Because I want to set an issuer for the token.

```java
		String issuer = this.providerSettings != null ? this.providerSettings.getIssuer() : null;
		Set<String> authorizedScopes = authorization.getAttribute(
				OAuth2Authorization.AUTHORIZED_SCOPE_ATTRIBUTE_NAME);
```

jymot:
Of course, I have submitted PR.

PR #458, opened 2021-10-09
#456 Set ProviderSettings for the default authentication providers in…

jymot:
Set `ProviderSettings` for the default authentication providers in `OAuth2TokenEndpointConfigurer.createDefaultAuthenticationProviders`

Closes gh-456

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2TokenEndpointConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2TokenEndpointConfigurer.java
@@ -156,6 +156,8 @@ public final class OAuth2TokenEndpointConfigurer extends AbstractOAuth2Configure
 	}
 
 	private <B extends HttpSecurityBuilder<B>> List<AuthenticationProvider> createDefaultAuthenticationProviders(B builder) {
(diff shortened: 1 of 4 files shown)


jymot:
Hi @sjohnr , i have made relevant changes.

jymot:
Hi @sjohnr , i have finished the modification. Thank you for your correction.

jymot:
Hi @sjohnr , thanks for the feedback. I have squashed all the commits.

> First, a question which we could also discuss on the ticket itself, but I'll ask here first: Are you currently exposing an `@Bean` of `ProviderSettings` with an `issuer` set? If not, can you try that and see if it causes the `issuer` setting to propagate into the `AuthenticationProvider`s without this change?

At present I have exposed a `@Bean` of `ProviderSettings`, like this:
```
    @Bean
    public ProviderSettings providerSettings() {
        return ProviderSettings.builder().issuer("http://xxxx.com").build();
    }
```
Because `AuthenticationProvider`'s is not an externally exposed `@Bean`, so `AutowiredAnnotationBeanPostProcessor` does not inject `ProviderSettings` into `AuthenticationProvider`'s.

> Second, for now I think we'll want to undo the change to remove `@Autowired(required = false)`. Though we could change it to simply `@Autowired`. For context, this is necessary at the current time for exactly the reason you opened #456, specifically so we could get the `issuer` without making the `ProviderSettings` a constructor argument. We will be looking into ways to make this better closer to 0.3.0. At that time, we can look into exactly how we want to solve this long term, which may include merging this PR, or even doing something else.

As I said above, even if it is changed to simple `@Autowired`, `ProviderSettings` will not be injected.

Issue #453, opened 2021-10-05
Require code_verifier if code_challenge provided


Kehrlann:
Current "requireProofKey" setting for clients is slightly unclear in the javadoc, it says:

> Set to {@code true} if the client is required to provide a proof key challenge and verifier when performing the Authorization Code Grant flow.

### When requireProofKey = true:

- in the authorization code request
  - a `code_challenge` is REQUIRED 
- in the token request
  - if the client is authenticated (e.g. through `client_secret`), then do not check the `code_verifier` (so the name "requireProofKey" is a bit misleading, it's not required here)
  - if the client is not authenticated (e.g. no `client_secret` was present), then check the `code_verifier` 

### When requireProofKey = false

- in the authorization code request
  - a `code_challenge` is not required but MAY be present
- in the token request
  - if the client is authenticated (e.g. through `client_secret`), then do not check the `code_verifier`
  - if the client is not authenticated (e.g. no `client_secret` was present), then check the `code_verifier`, it must be present and valid wrt/ `code_challenge`. If no `code_challenge` was supplied, it fails (but it fails by validating PKCE, it maybe should fail earlier?)

### In every case

- if a `code_challenge` is supplied in the authorization code request
  - a valid `code_verifier` MUST be present in the token request (differs from current behavior)

### When requireProofKey = true:

- in the authorization code request
  - a `code_challenge` is REQUIRED (same as current behavior)
- in the  token request
  - a valid `code_verifier` MUST be present (see above, differs from the current behavior)
  - a `client_secret` or any other authentication, MAY be present, and, in that case, MUST be validated (same as current)

### When requireProofKey = false

- in the authorization code request
  - a `code_challenge` is not required but MAY be present (same as current)
- in the token request
  -  if a `code_challenge` was sent in the authorization code request
     - a valid `code_verifier` MUST be present (see above, this differs from the current behavior)
     - UNSURE: should we also check the authentication? I believe if a client_secret or other authentication is provided, it should be validated, but if no authentication was sent, the request is still valid.
  - if a `code_challenge` was not sent, a valid authentication is REQUIRED (same as current)

## Alternatives

An alternate path could be to change "allowPublicClients" which only does verification on the token requests, but I personally prefer requireProofKey.

PR #465, opened 2021-10-22
Require code_verifier if code_challenge provided

Kehrlann:
Closes gh-453


jgrandja:
Thanks for the fix @Kehrlann ! This is now in main.

Issue #488, opened 2021-11-11
Authorization server metadata is missing userinfo_endpoint


enlightenedmouse:
**Describe the bug**
Now that the user info endpoint has been implemented, Issue #176; I believe the authorization server metadata (OidcProviderConfiguration) should also include the userinfo_endpoint.

bjornharvold:
🍻

PR #489, opened 2021-11-11
Adds userinfo_endpoint to authorization server metadata

enlightenedmouse:
Fixes gh-488

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2AuthorizationServerMetadata.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2AuthorizationServerMetadata.java
@@ -139,6 +139,16 @@ public abstract class AbstractOAuth2AuthorizationServerMetadata implements OAuth
 		}
 
 		/**
+		 * Use this {@code userinfo_endpoint} in the resulting {@link AbstractOAuth2AuthorizationServerMetadata}, OPTIONAL.
+		 *
+		 * @param userInfoEndpoint the {@code URL} of the OAuth 2.0 UserInfo Endpoint
+		 * @return the {@link AbstractBuilder} for further configuration
+		 */
+		public B userInfoEndpoint(String userInfoEndpoint) {
+			return claim(OAuth2AuthorizationServerMetadataClaimNames.USER_INFO_ENDPOINT, userInfoEndpoint);
+		}
+
+		/**
 		 * Add this OAuth 2.0 {@code scope} to the collection of {@code scopes_supported}
 		 * in the resulting {@link AbstractOAuth2AuthorizationServerMetadata}, RECOMMENDED.
 		 *
@@ -342,6 +352,9 @@ public abstract class AbstractOAuth2AuthorizationServerMetadata implements OAuth
 			}
 			if (getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.JWKS_URI) != null) {
 				validateURL(getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.JWKS_URI), "jwksUri must be a valid URL");
+			}
+			if (getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.USER_INFO_ENDPOINT) != null) {
+				validateURL(getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.USER_INFO_ENDPOINT), "userInfoEndpoint must be a valid URL");
 			}
 			if (getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.SCOPES_SUPPORTED) != null) {
 				Assert.isInstanceOf(List.class, getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.SCOPES_SUPPORTED), "scopes must be of type List");
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/OAuth2AuthorizationServerMetadataClaimNames.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/OAuth2AuthorizationServerMetadataClaimNames.java
@@ -52,6 +52,11 @@ public interface OAuth2AuthorizationServerMetadataClaimNames {
 	String JWKS_URI = "jwks_uri";
 
 	/**
+	 * {@code userinfo_endpoint} - the {@code URL} of the OAuth 2.0 UserInfo Endpoint
+	 */
+	String USER_INFO_ENDPOINT = "userinfo_endpoint";
+
+	/**
 	 * {@code scopes_supported} - the OAuth 2.0 {@code scope} values supported
 	 */
 	String SCOPES_SUPPORTED = "scopes_supported";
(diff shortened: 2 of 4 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationServerMetadataEndpointFilter.java
```diff
@@ -83,6 +83,7 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
 				.tokenEndpoint(asUrl(this.providerSettings.getIssuer(), this.providerSettings.getTokenEndpoint()))
 				.tokenEndpointAuthenticationMethods(clientAuthenticationMethods())
 				.jwkSetUrl(asUrl(this.providerSettings.getIssuer(), this.providerSettings.getJwkSetEndpoint()))
+				.userInfoEndpoint(asUrl(this.providerSettings.getIssuer(), this.providerSettings.getOidcUserInfoEndpoint()))
```
jgrandja, line 86: The UserInfo endpoint is not defined in RFC 8414 [2.  Authorization Server Metadata](https://datatracker.ietf.org/doc/html/rfc8414#section-2) and instead defined only in [3. OpenID Provider Metadata](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderMetadata).

Therefore, the update should only be accessible in `OidcProviderConfiguration` not in `OAuth2AuthorizationServerMetadata`

jgrandja (changes requested):
Thanks for the PR @enlightenedmouse. Please see review comment.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationServerMetadataEndpointFilter.java
enlightenedmouse: Hi @jgrandja ... Got it thanks : ) I added it because of the comment at the end of Section [2.](https://datatracker.ietf.org/doc/html/rfc8414#section-2) "Additional authorization server metadata parameters MAY also be used" and the sample in section [3.2. Authorization Server Metadata Response](https://datatracker.ietf.org/doc/html/rfc8414#section-3.2) shows userinfo_endpoint. But more so, I added it because I wasn't sure if the ClientRegistration within oauth2-client would be configured using the openid configuration endpoint or an authorization server metadata endpoint. After looking into the docs, it appears it tries both. So your call ; ) Let me know.

sjohnr, line 86: I think the preference at the moment would be to remove this from `OAuth2AuthorizationServerMetadataEndpointFilter` at this time, as we want to align as closely to the spec as possible (and not support optional features/items such as this).

Since the `OAuth2AuthorizationServerMetadataEndpointFilter` is not very customizable at the moment, we should see if anyone requires this, and open an enhancement for customization options instead if needed.

Would you be able to provide an update to this PR to address this?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2AuthorizationServerMetadata.java
```diff
@@ -343,6 +353,9 @@ protected void validate() {
 			if (getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.JWKS_URI) != null) {
 				validateURL(getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.JWKS_URI), "jwksUri must be a valid URL");
 			}
+			if (getClaims().get(OAuth2AuthorizationServerMetadataClaimNames.USER_INFO_ENDPOINT) != null) {
```
sjohnr, line 356: Related to supporting only `oidc-configuration`, please move this to `OidcProviderConfiguration`. Also, I think we should add a test to ensure this is covered.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/AbstractOAuth2AuthorizationServerMetadata.java
```diff
@@ -138,6 +138,16 @@ public B jwkSetUrl(String jwkSetUrl) {
 			return claim(OAuth2AuthorizationServerMetadataClaimNames.JWKS_URI, jwkSetUrl);
 		}
 
+		/**
+		 * Use this {@code userinfo_endpoint} in the resulting {@link AbstractOAuth2AuthorizationServerMetadata}, OPTIONAL.
+		 *
+		 * @param userInfoEndpoint the {@code URL} of the OAuth 2.0 UserInfo Endpoint
+		 * @return the {@link AbstractBuilder} for further configuration
+		 */
+		public B userInfoEndpoint(String userInfoEndpoint) {
```
sjohnr, line 147: Related to supporting only `oidc-configuration`, please move this to `OidcProviderConfiguration` and update the javadoc accordingly.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/core/OAuth2AuthorizationServerMetadataClaimNames.java
```diff
@@ -51,6 +51,11 @@
 	 */
 	String JWKS_URI = "jwks_uri";
 
+	/**
+	 * {@code userinfo_endpoint} - the {@code URL} of the OAuth 2.0 UserInfo Endpoint
+	 */
+	String USER_INFO_ENDPOINT = "userinfo_endpoint";
```
sjohnr, line 57: Related to supporting only `oidc-configuration`, please move this to `OidcProviderMetadataClaimNames`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcProviderConfigurationEndpointFilter.java
```diff
@@ -83,6 +83,7 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
 				.tokenEndpointAuthenticationMethod(ClientAuthenticationMethod.CLIENT_SECRET_BASIC.getValue())
 				.tokenEndpointAuthenticationMethod(ClientAuthenticationMethod.CLIENT_SECRET_POST.getValue())
 				.jwkSetUrl(asUrl(this.providerSettings.getIssuer(), this.providerSettings.getJwkSetEndpoint()))
+				.userInfoEndpoint(asUrl(this.providerSettings.getIssuer(), this.providerSettings.getOidcUserInfoEndpoint()))
```
sjohnr, line 86: I think we should add to an existing test to ensure this is covered.

sjohnr (changes requested):
Hi @enlightenedmouse, thanks for the PR! Sorry for the long delay (over the holidays), but I wanted to pick this thread up again and help move it forward. I have provided some feedback for your PR below. Would you be able to take a look?

sjohnr:
Just a friendly reminder also that it is now 2022 ❗😱, so you will want to think about updating changed files with 2022 in the copyright headers. 😄

sjohnr:
@enlightenedmouse have you had a chance to review the above comments?

sjohnr:
@enlightenedmouse I went ahead and merged this into `main` as 4081d460a26701ad1828542f688ec922f110416a and added a polish commit 5412f10ff80c9e1f4169d9989e7dfbfea448dff7 for the feedback.

Issue #495, opened 2021-11-16
Registration access token cannot be de-serialized when calling Client Configuration Endpoint


ovidiupopa91:
Because the initial registration access token saves the authorized scope in a `SingletonSet` when calling the Client Configuration Endpoint the access token cannot be deserialized because there's no Mixin registered. 

A new mixin and custom deserializer should be registered in `OAuth2AuthorizationServerJackson2Module`

jgrandja:
@ovidiupopa91 Instead of 

> A new mixin and custom deserializer

Changing `Collections.singleton()` to new `HashSet()` would be the easiest fix.

ovidiupopa91:
@jgrandja that's true. This is how it was initially, but in your polish commit you've changed it and I think it makes sense to have an immutable `Set` here. I can change it to `Collections.unmodifiableSet`

PR #497, opened 2021-11-16
Fix registration access token cannot be deserialized

ovidiupopa91:
Closes gh-495

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
@@ -21,6 +21,7 @@ import java.time.Instant;
 import java.util.Base64;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.UUID;
@@ -218,7 +219,9 @@ public final class OidcClientRegistrationAuthenticationProvider implements Authe
 	private OAuth2Authorization registerAccessToken(RegisteredClient registeredClient) {
 		JoseHeader headers = JwtUtils.headers().build();
 
-		Set<String> authorizedScopes = Collections.singleton(DEFAULT_CLIENT_CONFIGURATION_AUTHORIZED_SCOPE);
+		Set<String> authorizedScopes = new HashSet<>();
+		authorizedScopes.add(DEFAULT_CLIENT_CONFIGURATION_AUTHORIZED_SCOPE);
+		authorizedScopes = Collections.unmodifiableSet(authorizedScopes);
 
 		JwtClaimsSet claims = JwtUtils.accessTokenClaims(
 				registeredClient, this.providerSettings.getIssuer(), registeredClient.getClientId(), authorizedScopes)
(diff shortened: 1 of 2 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
```diff
@@ -222,7 +223,8 @@ private OidcClientRegistrationAuthenticationToken registerClient(OidcClientRegis
 	private OAuth2Authorization registerAccessToken(RegisteredClient registeredClient, String issuer) {
 		JoseHeader headers = JwtUtils.headers().build();
 
-		Set<String> authorizedScopes = Collections.singleton(DEFAULT_CLIENT_CONFIGURATION_AUTHORIZED_SCOPE);
+		Set<String> authorizedScopes = Collections.unmodifiableSet(
```
jgrandja, line 226: This could be changed to:

```java
		Set<String> authorizedScopes = Collections.unmodifiableSet(
				Collections.singleton(DEFAULT_CLIENT_CONFIGURATION_AUTHORIZED_SCOPE));
```

to reduce the extra `HashSet` creation but it may be confusing passing an immutable set to `Collections.unmodifiableSet()`. I think it would less confusing with:

```java
Set<String> authorizedScopes = new HashSet<>();
authorizedScopes.add(DEFAULT_CLIENT_CONFIGURATION_AUTHORIZED_SCOPE);
authorizedScopes = Collections.unmodifiableSet(authorizedScopes2);
```

What do you think?

Can you also please add a test. Thanks.

ovidiupopa91: Sure @jgrandja , I'll do the change. 
I can change `authenticateWhenClientRegistrationRequestAndValidAccessTokenThenReturnClientRegistration` test to check if the authorized scope is unmodifiable. 

There is already an assertion checking that only one scope is available. 
```
assertThat(authorizationResult.getAccessToken().getToken().getScopes())
				.containsExactly("client.read");
```

I will assert that UnsupportedOperationException is thrown when a new element is added to the Set (it can be changed to `isUnmodifiable` from assertj when the version will be upgraded to `3.21.0`)

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProviderTests.java
```diff
@@ -375,8 +375,9 @@ public void authenticateWhenClientRegistrationRequestAndValidAccessTokenThenRetu
 
 		// assert "registration" access token, which should be used for subsequent calls to client configuration endpoint
 		OAuth2Authorization authorizationResult = authorizationCaptor.getAllValues().get(0);
-		assertThat(authorizationResult.getAccessToken().getToken().getScopes())
-				.containsExactly("client.read");
+		Set<String> authorizedScopes = authorizationResult.getAccessToken().getToken().getScopes();
```
jgrandja, line 378: Let's revert this change. Instead, configure the integration tests in `OidcClientRegistrationTests` to use `JdbcOAuth2AuthorizationService` instead of `InMemoryOAuth2AuthorizationService` and it will reproduce the error.

jgrandja: @ovidiupopa91 Do you have time to apply this change by EOD tomorrow? I'd like to get this into this Wed's release. If you don't have time, I can apply it. Please let me know.

ovidiupopa91: hi @jgrandja. I had some problems with my laptop. It's working now. I will apply the change right now

jgrandja:
Thanks for the updates @ovidiupopa91 ! This is now in main.

PR #507, opened 2021-11-23
Fix OAuth2ClientAuthenticationToken deserialization

### LINKED ISSUE #482: OAuth2ClientAuthenticationToken should not be persisted across requests
**Describe the bug**
Sometimes the server returns a 500 error with the message:

> java.lang.IllegalArgumentException: The class with org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientAuthenticationToken and name of org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientAuthenticationToken is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See spring-projects/spring-security#4370 for details

**To Reproduce**

I encountered the bug while testing a resource server in conjunction with a new Spring Authorization Server's authorization code flow. The API tool I use automatically does the OAuth2 authentication, using the same session cookie from last time when it gets a new token, and that then stops working due to the 500 error. This is a contrived example, but it does trigger the bug reliably:

1) Modify the sample authorization server, adding `.redirectUri("https://oidcdebugger.com/debug")` to the `RegisteredClient` and then run the server.

2) Open a browser, go to https://oidcdebugger.com/debug, open the network tab of the browser inspector.

3) Fill the form in, log-in, and get an authorization code.

4) Look in the browser inspector to get the JSESSIONID cookie from `/oauth2/authorize`.

4) I'm not sure how quickly the next steps need to be done, but now POST directly to the `/oauth2/token` endpoint. Run curl or your preferred API client:

> export AUTHCODE="YOUR CODE HERE" && curl -X "POST" "http://127.0.0.1:9000/oauth2/token" \\
     -H 'Cookie: JSESSIONID=COOKIE FROM BROWSER' \\
     -H 'Content-Type: application/x-www-form-urlencoded; charset=utf-8' \\
     -u 'messaging-client:secret' \\
     --data-urlencode "grant_type=authorization_code" \\
     --data-urlencode "code=$AUTHCODE" \\
     --data-urlencode "redirect_uri=https://oidcdebugger.com/debug"

5) A token should be correctly returned.

6) Click "Start Over" in the browser, which will hopefully cause the browser to send the same JSESSIONID with the new call to `/oauth2/authorize`, and get the new authorization code. Try the POST again with the new code and the same JSESSIONID. It should return a 500 error, and in the logs will be the exception.

**Expected behavior**

I expected to get a new token, and if not, an error message rather than an exception.

edwardzjl:
Fixes #482

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/jackson2/AuthorizationGrantTypeMixin.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/jackson2/AuthorizationGrantTypeMixin.java
@@ -1,0 +1,30 @@
+package org.springframework.security.oauth2.server.authorization.jackson2;
+
+import com.fasterxml.jackson.annotation.JsonAutoDetect;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
+
+/**
+ * Jackson Mixin class helps in serialize/deserialize
+ * {@link org.springframework.security.oauth2.core.AuthorizationGrantType}.
+ *
+ * @author Junlin Zhou
+ */
+@JsonTypeInfo(use = JsonTypeInfo.Id.CLASS)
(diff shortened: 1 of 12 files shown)


jgrandja:
@edwardzjl Closing this in favour of d0e1107. Please see [comment](https://github.com/spring-projects/spring-authorization-server/issues/482#issuecomment-987999937).

PR #504, opened 2021-11-19
Bump up Gradle enterprise plugin to 3.7.2

jprinet:

PR #547, opened 2021-12-20
Update gradle enterprise to 3.8 to address CVE-2021-45105.

theArtechnology:
https://security.gradle.com/advisory/2021-11

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -6,7 +6,7 @@ pluginManagement {
 }
 
 plugins {
-	id "com.gradle.enterprise" version "3.7.2"
+	id "com.gradle.enterprise" version "3.8"
 	id "io.spring.ge.conventions" version "0.0.7"
 }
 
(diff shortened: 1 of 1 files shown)


sjohnr:
Thanks @theArtechnology! Looks like there's also a [3.8 version available](https://plugins.gradle.org/plugin/com.gradle.enterprise/3.8) as of Monday (Dec 20, 2021). Unfortunately, I'm not familiar enough with Gradle Enterprise to be confident to merge this yet, so I think we'll wait until after the holidays anyway. Let me know if you feel that's ill-advised.

sjohnr:
Note that this seems to be a duplicate of #504 as well.

theArtechnology:
> 

Thank you for reviewing my Pull Request, it is a recommendation from Gradle themselves: https://security.gradle.com/advisory/2021-11

I can make the change so it reflects 3.8 as you mentioned.

sjohnr:
Thanks! This is now in `main` as 1370f7e51aee7fa6a187776f4186cbcbd1a004ef.

PR #554, opened 2021-12-24
Add docs outline with Antora skeleton

sjohnr:


docs/modules/ROOT/pages/core-components.adoc
```diff
@@ -0,0 +1,28 @@
+= Core Components
+
+This page is under construction.
+
+[[registered-client-repository]]
+== `RegisteredClientRepository` / `RegisteredClient`
+
+This section is under construction.
```
jgrandja, line 8: @Buzzardo Is this the appropriate wording to use for TODO sections?

docs/modules/guides/nav.adoc
```diff
@@ -0,0 +1,2 @@
+.xref:index.adoc[Guides]
```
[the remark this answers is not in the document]
Buzzardo: I would add "How-to", for clarity's sake.

docs/modules/ROOT/pages/core-components.adoc
Buzzardo: The heading is a problem. With only the two object names, there is no indication of what you are going to say about them. Should it be "Using `RegisteredClientRepository` and `RegisteredClient`" or "Choosing between `RegisteredClientRepository` and `RegisteredClient`"?

Note that, in either case, I replaced the slash with "and". We see a slash used that way all the time in informal documents (email and so on), but using a slash in place of a conjunction is bad practice in formal documents (as all documentation is).

The "under construction" wording is fine. Would it be better to not show pages that have no content, though?

sjohnr: @Buzzardo, regarding showing no-content pages, I'm hoping to begin linking between sections of the guide based on your recommendation to link early, link often :wink: (even before the sections are finished). While some things will change, I'm hoping we can create some semblance of structure prior to writing the content for each section.

sjohnr: Regarding the headings, I agree and think we can meet to discuss good headings for these sections in the coming weeks. In the meantime, I'm going to go ahead and merge what I've got here to just match the initial draft outline, and we'll address this iteratively.

jgrandja:
Merged via 438dd3e

PR #580, opened 2022-01-17
Update OAuth2EndpointUtils.java

sheiy:
make public for implement custom AuthenticationConverter


jgrandja:
@sheiy `OAuth2EndpointUtils` is meant for internal use only. There are no plans to expose it.

PR #581, opened 2022-01-17
Throw invalid_grant when invalid token request with PKCE

Kehrlann:
- When providing an invalid `authorization_code` in a token request with a PKCE code_verifier, throw `invalid_grant`, as per RFC 6749. It is consistent with behavior in `OAuth2AuthorizationCodeAuthenticationProvider#authenticate`
- When providing an invalid `code_verifier`, throw `invalid_grant`, as per RFC7636.

There is an additional corner-case in `OAuth2ClientAuthenticationProvider#authenticatePkceIfAvailable`, when there is no `code_challenge` in the authorization request and the client requires PKCE.
- In the "normal" case, that should not be possible because the OAuth2AuthorizationCodeRequestAuthenticationProvider#authenticateAuthorizationRequest`will throw an `invalid_request`.
- There might be an edge case where the client configuration changes between the time an authorization code is issued and the time it is exchanged for a token.

I am unsure what we should throw in this case, I put `invalid_grant` for consistency, feel free to change it.


jgrandja:
Thanks for this fix @Kehrlann ! This is now in main.

PR #582, opened 2022-01-17
Bump up Gradle plugin dependencies

jprinet:


jgrandja:
@jprinet The `io.spring.ge.conventions:0.0.9` (latest version) builds against `com.gradle.enterprise:3.8` (see plugin [properties](https://github.com/spring-io/gradle-enterprise-conventions/blob/v0.0.9/gradle.properties#L3).

I'd like to keep the dependency in sync so we'll remain on `3.8`.

Issue #558, opened 2021-12-28
Support for NoSql datastores (Redis)


mikesaurus:
**Expected Behavior**
Some infrastructures do not include SQL-based data stores and instead make extensive use of NoSql data stores such as Redis, making the introduction of a SQL data store to support OAuth capabilities incredibly burdensome.  spring-authorization-server data objects that are persisted as part of the process flow (such as OAuth2Authorization) should be easily (de)serializable using something like a JSON mapper in order to allow for either Spring provided/packaged or app developer created Redis implementations of OAuth2AuthorizationConsentService and OAuth2AuthorizationService.

**Current Behavior**
Serializing/deserializing spring-auth-server data objects to/from JSON requires custom data mapping that can create a maintenance nightmare in terms spring-auth-server compatibility (especially in the early/pre-v1.x stages of the spring-auth-server project).

**Context**
We currently use Redis as the backing data store for our customized spring-security-oauth2-based authorization server.  We would like to upgrade to a spring-authorization-server-based authorization server without requiring a SQL database and all of the maintenance and operational overhead associated with it, since all of our current infrastructure and operations are geared toward Redis. We can try to work around the (de)serialization issues with custom data mapping, but custom data mapping tightly couples our code to the underlying data object implementations and concerns us in terms of maintenance/compatibility with spring-authorization-server upgrades.

mikesaurus:
One option would be to include a JsonMapper<T> in spring-authorization-server that encapsulates the OAuth2 Object<->JSON mapping so that the actual data objects don't have to change, but anyone wishing to implement a document-based persistence strategy can easily transform the objects to/from JSON without requiring the internal data mapping to be implemented outside of spring-auth-server.

bjornharvold:
Agree.

I started creating our own MongoOAuth2AuthorizationService and MongoOAuth2AuthorizationConsentService. Serializing / Deserializing to MongoDb is a nightmare. I tried to look at the JDBC version but didn't provide much guidance in terms of an easy way to port.

Attaching what I've done so far. The MongoRegisteredClientRepository seems OK but currently chose to use the in-memory implementations of OAuth2AuthorizationService and OAuth2AuthorizationConsentService.

If someone can provide some guidance, I'd be more than happy to contribute the MongoDb version of these interfaces.

[Archive.zip](https://github.com/spring-projects/spring-authorization-server/files/7791379/Archive.zip)
.

Skippr-admin:
[...]

PR #588, opened 2022-01-18
Add mappings to/from intermediate objects for NoSQL support

sjohnr:
This is a draft solution for supporting NoSQL databases. It uses intermediary objects (currently called "Resources", but open to better names) to simplify a transformation to/from JSON. The intermediary objects support full serialization to JSON with the use of the `OAuth2AuthorizationServerJackson2Module`. Supports the following:

* Handles iterating/mapping supported token types, removing the need for applications to do this
* Converts enumerated types to/from strings with support for "custom" types
* Converts authorities to/from strings
* Supports custom settings values

Issue gh-558


HKMV:
Hello, there is no normal merging

sjohnr:
Closing for now. May re-open when we revisit this.

Issue #611, opened 2022-01-27
Surprising "error description" when denying approval


bnutzer:
**Describe the bug^Wissue**
When POSTing to the authorize "endpoint" with an empty scope list (i.e., denying approval), it responds with a redirection to the redirect uri, setting the "error" parameter to "access_denied" (as expected). However, the "error_description" is set to "OAuth 2.0 Parameter: client_id", which does not really reflect the issue and is not helpful while analyzing the any failures.

**Expected behavior**
A better error_description would be preferable. The (old) spring-security-oauth2 server replied with an error_description "User denied access".

**Sample**
Any of the sample projects shows this behavior.

Code is in OAuth2AuthorizationCodeRequestAuthenticationProvider.java

Verified with spring-authorization-server 0.2.2

Thanks :)

jgrandja:
@bnutzer I agree a better error description would be something like "User denied access", however, the expectation is that the client application would consume the "error" parameter and display its own custom error message.

Alternatively, the returned OAuth 2.0 Error can also be overridden via a custom `AuthenticationFailureHandler` using this custom configuration:

```java
OAuth2AuthorizationServerConfigurer<HttpSecurity> authorizationServerConfigurer =
		new OAuth2AuthorizationServerConfigurer<>();
authorizationServerConfigurer
		.authorizationEndpoint(authorizationEndpoint ->
				authorizationEndpoint.errorResponseHandler(...)
```


Also notice that **ALL** OAuth 2.0 error descriptions generated by `OAuth2AuthorizationCodeRequestAuthenticationProvider` default to "OAuth 2.0 Parameter: [parameterName]", not just the "access_denied" error flow.

FYI, it's best practice NOT to reveal too much information related to the error as this can provide "hints" to the caller, that may be malicious, and hence allow them to navigate the next call.

I'm going to close this as works-as-designed.

PR #612, opened 2022-01-31
changes error message when revoking permission to "User Denied Access"

discodamone:
I saw the issue at https://github.com/spring-projects/spring-authorization-server/issues/611 and believe that these changes resolve it. I understand if you do not decide to merge (especially if this interferes with something else); my motivation was to complete a university course requirement.


jgrandja:
Thanks for the PR @discodamone, however, we won't be merging as explained in this [comment](https://github.com/spring-projects/spring-authorization-server/issues/611#issuecomment-1033018245).

PR #626, opened 2022-02-15
Core 2422

bob-walters:
DONT MERGE please.  This is just here to make browsing the changes easier.

Most of the changes are NOT to the main spring-authorization-service, and even those superficial changes (e.g. private to public, and/or one extra method added, etc.) can probably be avoided with a little deeper analysis of the way that the spring-security library DSL works in carrying out configuration.   Some hints to that might be found in https://github.com/spring-projects/spring-authorization-server/compare/main...sjohnr:identity-federation which also contains an alternative approach


sjohnr:
@bob-walters, thanks for your interest in the project!

Am I correct in assuming that you are using the GitHub UI to view a set of changes you're working through for your own project? If so, you can actually create a PR from your fork's feature branch to your fork's `main` branch [here](https://github.com/accesso/spring-authorization-server/compare/main...CORE-2422?expand=1). I'm going to close this PR as it doesn't look like it's intended for the main project. Let me know if I'm missing anything.

Issue #414, opened 2021-08-20
Seperate JWT Token generation


GrmpfNarf:
**Expected Behavior**
It would be nice when the JWT Token generation from ``OAuth2ClientCredentialsAuthenticationProvider`` and ``OAuth2AuthorizationCodeAuthenticationProvider`` is moved to a sperate class so I can reuse it e.g. for an custom AuthenticationProvider
**Current Behavior**
Currently the JWT Token generation for the Access Token is inside the ``authenticate`` method of the both classes. To use it in a custom Authenticationprovider I have to copy the logic and everything around e.g. the ``JwtUtil`` and the resolve of the JwtEncoder from ``OAuth2ConfigurerUtils.getJwtEncoder(builder)``

**Context**
I'm adding a custom Grant Type by adding a converter to the tokenEndpoint which provides an custom AutenticationToken which will be resolved by the custom AuthenticationProvider.

jgrandja:
Thanks for the report @GrmpfNarf.

Hiding the `Jwt` token generation in the `AuthenticationProvider`'s was intentional. However, the plan is to introduce an interface, e.g. `OAuth2TokenIssuer`, which can support any type of token. We'll soon be introducing opaque token support which will trigger some re-factoring.

Rest assured, the `Jwt` token generation will be extracted into a re-useable component, we're just not there yet.

I'll assign this to `0.2.1` but it may get pushed to a later release.

denis111:
The same happens to me, I had to copy-paste `JwtUtil` besides it seems that `JoseHeader` needs to be public because `JwtUtil` returns `JoseHeader.Builder` if you call `JwtUtils.headers()` that is needed to call then `JwtEncodingContext.with(headersBuilder, claimsBuilder)`.  But I see that conflicting class `JoseHeader` in Spring Security 5.6 is not public! So to make any custom Provider we `JoseHeader` to be public or maybe the future refactoring will not require it?

erised:
Hi @jgrandja.
May I take this up if it is not already assigned to anyone ?

jgrandja:
@erised Let me get back to you next week. I'm playing catch up this week :)

jgrandja:
@erised gh-493 and gh-571 are available. Do any of these interest you?

erised:
Hi @jgrandja. Sure. I can work on [gh-493](https://github.com/spring-projects/spring-authorization-server/issues/493).

jgrandja:
[...]

Issue #500, opened 2021-11-17
Support opaque access tokens


jgrandja:
The currently supported format for access tokens is JWT. We should also support opaque access tokens.

Related gh-414

dfcoffin:
This continues to be pushed to the next milestone release.  Any idea what it will take to finally support opaque access tokens?  I have a UT at Dallas capstone project starting today that requires an authorization server that supports opaque access tokens.

jgrandja:
Apologies @dfcoffin but there is a lot of work still to do and we're doing our best to prioritize features.

This one is top priority for `0.2.3` so I don't see it getting pushed back further.

dfcoffin:
Thanks, @jgrandja.

jgrandja:
@dfcoffin I just merged opaque access token support! Thanks for your patience.

As you know, we're in the process of writing the reference documentation so there is no documentation on usage just yet.

Take a look at the test [OAuth2TokenIntrospectionTests.requestWhenObtainReferenceAccessTokenAndIntrospectThenActive()](https://github.com/spring-projects/spring-authorization-server/blob/32414451f5163235739c045a3ce361bcc2eec599/oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2TokenIntrospectionTests.java#L253), which demonstrates how to configure a `RegisteredClient` for opaque access tokens and performs obtaining the token and then introspecting the token.

FYI, the `AuthorizationServerConfiguration` in this test completely disables JWT support. If you also want to enable JWT support along with opaque tokens, then simply register a `JWKSource<SecurityContext>` `@Bean` or a `JwtEncoder` `@Bean`. Alternatively, you can register a `OAuth2TokenGenerator` `@Bean`, with whatever capabilities you require. Take a look at the default in [OAuth2ConfigurerUtils.getTokenGenerator()](https://github.com/spring-projects/spring-authorization-server/blob/32414451f5163235739c045a3ce361bcc2eec599/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2ConfigurerUtils.java#L93).

Please give this a try and would love to hear your feedback on usage. We want to ensure the API is easy to use so feedback on this is important.

Let me know if you have any questions.
Thanks.

dfcoffin:
[...]

PR #628, opened 2022-02-16
Introduce OAuth2TokenGenerator

jgrandja:
This PR introduces `OAuth2TokenGenerator`, enabling an implementation to generate an `OAuth2Token`.

The initial implementation is `JwtGenerator`, which consolidates the `Jwt` encoding logic that existed in the various `AuthenticationProvider`'s and `JwtUtils`.

The `OAuth2TokenGenerator` will enable support for Opaque Access Tokens gh-500, which is the next feature up.

Furthermore, the `OAuth2TokenGenerator` is flexible enough to allow for any type of custom format for access tokens and/or refresh tokens.

Closes gh-414


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationProvider.java
```diff
@@ -66,45 +69,64 @@
...
 public final class OAuth2RefreshTokenAuthenticationProvider implements AuthenticationProvider {
+	private static final String ERROR_URI = "https://datatracker.ietf.org/doc/html/rfc6749#section-5.2";
 	private static final OAuth2TokenType ID_TOKEN_TOKEN_TYPE = new OAuth2TokenType(OidcParameterNames.ID_TOKEN);
 	private static final StringKeyGenerator DEFAULT_REFRESH_TOKEN_GENERATOR =
 			new Base64StringKeyGenerator(Base64.getUrlEncoder().withoutPadding(), 96);
 	private final OAuth2AuthorizationService authorizationService;
-	private final JwtEncoder jwtEncoder;
-	private OAuth2TokenCustomizer<JwtEncodingContext> jwtCustomizer = (context) -> {};
+	private final OAuth2TokenGenerator<? extends OAuth2Token> tokenGenerator;
 	private Supplier<String> refreshTokenGenerator = DEFAULT_REFRESH_TOKEN_GENERATOR::generateKey;
```
sjohnr, line 83: If we eventually have refresh tokens produced by `OAuth2TokenGenerator` as well, would the plan be to deprecate the `refreshTokenGenerator` field and use a single generator for all tokens? In this case, would the default implementation be a delegation pattern that encapsulates a `JwtGenerator` and `OpaqueTokenGenerator`, delegated to based on token type? Or will there be a second `OAuth2TokenGenerator` just for refresh tokens?

I'm just trying to visualize how we will have a combination of `Jwt` and non-`Jwt` generators in one class.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2TokenGenerator.java
```diff
@@ -0,0 +1,44 @@
...
+ * using the attributes contained in the {@link OAuth2TokenContext}.
+ *
+ * @author Joe Grandja
+ * @since 0.2.3
+ * @see OAuth2Token
+ * @see OAuth2TokenContext
+ * @param <T> the type of the OAuth 2.0 Token
+ */
+@FunctionalInterface
+public interface OAuth2TokenGenerator<T extends OAuth2Token> {
```
sjohnr, line 32: In the case of `JwtGenerator`, the generic type `T` is a `Jwt`. When there's an `OpaqueTokenGenerator` class, what will the generic type be? Is it just `OAuth2AccessToken`?

sjohnr (commented):
This looks pretty great so far! I have a few questions below for discussion, that could change something or other (but hopefully won't change much if anything).

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationProvider.java
jgrandja: > would the plan be to deprecate the `refreshTokenGenerator` field and use a single generator for all tokens

Yes. We would leverage a delegation-based pattern where `DelegatingOAuth2TokenGenerator` would be composed of `JwtGenerator` and `OAuth2OpaqueTokenGenerator`. The `OAuth2OpaqueTokenGenerator` would contain pre-processing logic as such:

```java
public class OAuth2OpaqueTokenGenerator implements OAuth2TokenGenerator<OAuth2OpaqueToken> {

	@Override
	public OAuth2OpaqueToken generate(OAuth2TokenContext context) {
		if (OAuth2TokenType.ACCESS_TOKEN.equals(context.getTokenType()) &&
				context.getRegisteredClient().getTokenSettings().getAccessTokenType().isOpaque()) {

			// Generate opaque access token

		}

		if (OAuth2TokenType.REFRESH_TOKEN.equals(context.getTokenType())) {

			// Generate opaque refresh token

		}
```

We will need to introduce a setting in `TokenSettings` that allows us to determine which token type format to generate for an access token (jwt or opaque) and we'll default to opaque for refresh tokens.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2TokenGenerator.java
jgrandja: I think it will be something like:

```java
public class OAuth2OpaqueToken extends AbstractOAuth2Token {
	private final Map<String, Object> claims;

	public OAuth2OpaqueToken(String tokenValue, Instant issuedAt, Instant expiresAt, @Nullable Map<String, Object> claims) {
		super(tokenValue, issuedAt, expiresAt);
		this.claims = Collections.unmodifiableMap(claims != null ? claims : Collections.emptyMap());
	}

}
```

This would be used for opaque access token (with `claims`) and potentially used for refresh tokens (with or without `claims`).

sjohnr (approved):

oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
```diff
@@ -146,6 +148,19 @@
...
 	}
 
+	/**
+	 * Sets the token generator.
+	 *
+	 * @param tokenGenerator the token generator
+	 * @return the {@link OAuth2AuthorizationServerConfigurer} for further configuration
+	 * @since 0.2.3
+	 */
+	public OAuth2AuthorizationServerConfigurer<B> tokenGenerator(OAuth2TokenGenerator<? extends OAuth2Token> tokenGenerator) {
```
sjohnr, line 158: 👍

Issue #493, opened 2021-11-14
Allow Token Introspection to be customized


zf1976:
- How can I return the information of the custom token？

> The OAuth2TokenIntrospectionAuthenticationProvider authentication provider cannot fully return token information, for example, I added custom authorities information. This is important to me because the resource server needs it to extract the corresponding permissions for the token

- code
```
private static OAuth2TokenIntrospection withActiveTokenClaims(
			OAuth2Authorization.Token<AbstractOAuth2Token> authorizedToken, RegisteredClient authorizedClient) {

		OAuth2TokenIntrospection.Builder tokenClaims = OAuth2TokenIntrospection.builder(true)
				.clientId(authorizedClient.getClientId());

		// TODO Set "username"

		AbstractOAuth2Token token = authorizedToken.getToken();
		if (token.getIssuedAt() != null) {
			tokenClaims.issuedAt(token.getIssuedAt());
		}
		if (token.getExpiresAt() != null) {
			tokenClaims.expiresAt(token.getExpiresAt());
		}

		if (OAuth2AccessToken.class.isAssignableFrom(token.getClass())) {
			OAuth2AccessToken accessToken = (OAuth2AccessToken) token;
			tokenClaims.scopes(scopes -> scopes.addAll(accessToken.getScopes()));
			tokenClaims.tokenType(accessToken.getTokenType().getValue());

			Map<String, Object> claims = authorizedToken.getClaims();
			if (!CollectionUtils.isEmpty(claims)) {
				// Assuming JWT as it's the only (currently) supported access token format
				JwtClaimAccessor jwtClaims = () -> claims;

				Instant notBefore = jwtClaims.getNotBefore();
				if (notBefore != null) {
					tokenClaims.notBefore(notBefore);
				}
				tokenClaims.subject(jwtClaims.getSubject());
				List<String> audience = jwtClaims.getAudience();
				if (!CollectionUtils.isEmpty(audience)) {
					tokenClaims.audiences(audiences -> audiences.addAll(audience));
				}
				tokenClaims.issuer(jwtClaims.getIssuer().toExternalForm());
				String jti = jwtClaims.getId();
				if (StringUtils.hasText(jti)) {
					tokenClaims.id(jti);
				}
			}
		}

		return tokenClaims.build();
	}
```

- return
```
{
    "active": true,
    "client_id": "system",
    "iat": 1636896937,
    "exp": 1636900537,
    "scope": "all",
    "token_type": "Bearer",
[...]

Issue #476, opened 2021-11-02
Allow Token Revocation to be customized


arfatbk:
**Expected Behavior**
Successful refresh Token Revocation should provide a token blocklist implementation for all related access tokens. Or Raise an Event, which then can be consumed to implement a custom token blocklist.

**Current Behavior**
After Refresh Token revocation, the access token remains valid and can be used by OAuth2 Resource servers.

**Context**
If we go with the 'Event' approach, after successful revocation of a refresh token, developers can implement blocklist as in memory, JDBC, Communicate over MQ.

I am using the OAuth Resource server that validates access token using the OAuth server's JWKs keys. After Refresh Token is revoked, There seems no way to communicate this to the OAuth Resource server. AFAIK @jgrandja please correct me or point me in the right direction. Thanks

jgrandja:
Thanks for getting in touch @arfatbk.

Spring Authorization Server does revoke the associated access token during a refresh token revocation (see [test](https://github.com/spring-projects/spring-authorization-server/blob/7680505eedc2116d012f0f3598361a84e2a31c9d/oauth2-authorization-server/src/test/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2TokenRevocationTests.java#L144)), but the JWT access token can still be used at the Resource Server, if the Resource Server verifies the token locally instead of calling the Introspection Endpoint.

This is an important use case to solve and as far as I am aware, there are no RFC standards that solve this problem. The Authorization Server and Resource Server(s) would need to communicate with each other so that the Resource Server's are up-to-date with all revoked tokens so it can implement the blocklist.

Instead of implementing the blocklist on the Resource Server, you could call the Introspection Endpoint with the JWT to determine if it's active or not. However, this would obviously downgrade performance with that extra hop.

The goal of this framework is to implement to spec and provide customization hooks for applications to customize the standard behaviour however they choose. Given that this would be custom behaviour, we would only provide customization hooks.

@arfatbk Would you be able to do some R&D to see what other providers have implemented to solve this problem?

arfatbk:
Thanks for your response. I will look for references.

arfatbk:
I am not able to find exact implementation by other providers. Closest I could manage is to route `revocation` request though Spring Cloud Gateway and on successful revocation, propagate the event to all resource servers.

Resource servers may store this in memory blacklist or Centralized cache.


[...]

PR #630, opened 2022-02-19
Allow Token Introspection to be customized

gourav:
Closes gh-493

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/config/annotation/web/configurers/oauth2/server/authorization/OAuth2AuthorizationServerConfigurer.java
@@ -51,7 +51,6 @@ import org.springframework.security.oauth2.server.authorization.web.OAuth2Author
 import org.springframework.security.oauth2.server.authorization.web.OAuth2TokenIntrospectionEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.OAuth2TokenRevocationEndpointFilter;
 import org.springframework.security.oauth2.server.authorization.web.ProviderContextFilter;
-import org.springframework.security.web.access.intercept.FilterSecurityInterceptor;
 import org.springframework.security.web.authentication.HttpStatusEntryPoint;
 import org.springframework.security.web.authentication.preauth.AbstractPreAuthenticatedProcessingFilter;
 import org.springframework.security.web.context.HttpRequestResponseHolder;
@@ -71,6 +70,7 @@ import org.springframework.util.Assert;
  * @author Daniel Garnier-Moiroux
  * @author Gerardo Roza
  * @author Ovidiu Popa
+ * @author Gaurav Tiwari
  * @since 0.0.1
  * @see AbstractHttpConfigurer
  * @see OAuth2ClientAuthenticationConfigurer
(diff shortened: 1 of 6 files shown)


gourav:
Hi @jgrandja 
It is ready for your review.

This is my first time contributing. Please let me know the changes required. I will do them right away.

jgrandja:
Apologies @gourav, as I have led you down the wrong path with my [comment](https://github.com/spring-projects/spring-authorization-server/issues/493#issuecomment-1036633369):

> We need the capability to customize the default claims initialized by `withActiveTokenClaims()`.
>
> Take a look at `OidcUserInfoAuthenticationProvider.setUserInfoMapper()` for inspiration, as the implementation might be similar. For example, we could create a new class `OAuth2TokenIntrospectionAuthenticationContext` that provides a getter for accessing `OAuth2TokenIntrospection.Builder`, allowing to customize the claims before `build()` is called.

There are 2 changes that are required.

1) The `OAuth2TokenIntrospectionAuthenticationProvider.withActiveTokenClaims()` initializes the standard claims as defined in the spec, however, it does not include any other claims that may be available in `authorizedToken.getClaims()`. For example, if an application configured an `OAuth2TokenCustomizer`, which added custom claims to the `Jwt` claim set then this would not be included in `OAuth2TokenIntrospection`. So we need to update this method to include ALL claims available `authorizedToken.getClaims()`, e.g. `OAuth2TokenIntrospection.withClaims(authorizedToken.getClaims())`

2) We need to add the customization hooks, which will be very similar to the hooks that were recently added for Token Revocation in gh-476. Take a look at the commits a846e936e9a5b72567058b5b96f69aa383bc2a4a and 4d5b2881164e3948524b402cbb8b2966a06afa95 and apply the similar changes.

gourav:
Hi @jgrandja. I am pretty sure that I did not understand your comment properly. So, no apologies required. :)
Let me get update it as per the suggestion.

gourav:
Hi @jgrandja 
What would these hooks be configuring ? 
As per your suggestion, I can simply add `OAuth2TokenIntrospection.withClaims(authorizedToken.getClaims())` to add all claims available in `authorizedToken.getClaims()`. 

What would the hook be customising as I believe all configuration is already present in the context ?
Or is it for customising `Function<OAuth2TokenIntrospectionAuthenticationContext, OAuth2TokenIntrospection> activeClaimsMapper` that will be introduced by current state of this pull request ?

jgrandja:
@gourav 

> Or is it for customising `Function<OAuth2TokenIntrospectionAuthenticationContext, OAuth2TokenIntrospection> activeClaimsMapper` that will be introduced by current state of this pull request

Please ignore my comment regarding `activeClaimsMapper` as this is not needed.

> What would these hooks be configuring ?

There are use cases where an application needs to customize the request and/or response for a specific protocol flow. For example, an application may want to add a custom parameter in the access token response. How would they do that? They would supply a custom `OAuth2TokenEndpointFilter.setAuthenticationSuccessHandler()` to customize the response using `OAuth2AccessTokenResponseHttpMessageConverter`.

The same applies for the token request. For example, a custom parameter may be needed and therefore the application can supply `OAuth2TokenEndpointFilter.setAuthenticationConverter()` to read the custom parameter so it's carried through to the `AuthenticationProvider`.

We need the same customization capabilities for token introspection. For example, an application may not want to return all the claims in `OAuth2TokenIntrospection.withClaims(authorizedToken.getClaims())` and possibly filter out one or more claims via `OAuth2TokenIntrospectionEndpointFilter.setAuthenticationSuccessHandler()` using `OAuth2TokenIntrospectionHttpMessageConverter` to customize the response.

Please review the code for `OAuth2TokenEndpointConfigurer` and `OAuth2TokenRevocationEndpointConfigurer` and see how all the components work together and ultimately allow for complete customization of request and response.

jgrandja:
Apologies for the delay @gourav. I've been quite busy with some priority tasks for the `0.2.3` release.

I came back to this yesterday and there were quite a few changes still needed before we can merge.
I'd like to include this in the `0.2.3` release so I went ahead and applied a polish commit.

Please review all the changes and let me know if you have any questions. Thanks.

gourav:
Changes seem good to me. :)
Please let me know if there is any other issue I can take on.

Issue #616, opened 2022-02-03
Ability to modify OIDC provider configuration


tcchambers-els:
**Expected Behavior**
When enabling OIDC it would be good to be able to customise `/.well-known/openid-configuration` information.  In my case I would like to modify the `scopes_supported` claim

**Current Behavior**
I can see how I can create an instance of `OidcProviderConfiguration` but I can't see how I can modify the one used here https://github.com/spring-projects/spring-authorization-server/blob/82e4f3a345101b454bc19d18fa0c127c6274439e/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcProviderConfigurationEndpointFilter.java#L79 in the filter without completely creating a new one (unless I've missed something)

**Context**
We would like the ability to modify some of the optional OIDC provider information to accurately reflect the actual provider.

sjohnr:
@tcchambers-els currently I believe you are correct, you would need to implement your own filter to replace this one. It looks like `scopes_supported` is a `RECOMMENDED` metadata value for this endpoint, so it could be a candidate for addition at some point.

Since the server supports multiple clients, how would you envision filling in this field? Would it just be hard-coded to `openid` and possibly other standard scope values? Or would it need to be a setting in `ProviderSettings` for your use case?

tcchambers-els:
Hi @sjohnr, I think either of those solutions would work - as you say `openid` needs to be mandatory but we would also need to add other standard OIDC scopes like `profile` plus custom scopes.  So adding the config to `ProviderSettings` would satisfy our use case, with either `openid` hardcoded as it is or an exception thrown if it's missing.
Thanks for your time in taking a look at this.

jgrandja:
@tcchambers-els @sjohnr I'm not sure adding the config to `ProviderSettings` is going to give us the flexibility for all potential use cases.

I think what we'll need to do is expose a setter for `OidcProviderConfigurationEndpointFilter.providerConfigurationHttpMessageConverter`, which allows a custom `Converter` to be set via `OidcProviderConfigurationHttpMessageConverter.setProviderConfigurationParametersConverter()`.

ovidiupopa91:
hi @jgrandja. I can start working on this issue.

its-felix:
Maybe it's already a bit too late poping up here, but I'd like to be able to completely disable the OIDC Part of the authorization server and only use oauth2.

Should I open another issue for this request?

PR #635, opened 2022-02-22
Ability to modify OIDC provider configuration

ovidiupopa91:
Closes gh-616

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcProviderConfigurationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcProviderConfigurationEndpointFilter.java
@@ -17,6 +17,7 @@ package org.springframework.security.oauth2.server.authorization.oidc.web;
 
 import java.io.IOException;
 import java.util.List;
+import java.util.Map;
 import java.util.function.Consumer;
 
 import javax.servlet.FilterChain;
@@ -24,6 +25,7 @@ import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.MediaType;
 import org.springframework.http.server.ServletServerHttpResponse;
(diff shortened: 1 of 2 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcProviderConfigurationEndpointFilter.java
```diff
@@ -103,6 +104,20 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
+	/**
+	 * Sets the {@link HttpMessageConverter} used for converting the {@link OidcProviderConfiguration}
+	 * from and to HTTP requests and responses
+	 *
+	 * @param providerConfigurationHttpMessageConverter the {@link HttpMessageConverter} used for converting a
+	 * representation of the OpenID Provider Configuration from and to HTTP requests and responses
+	 *
+	 * @since 0.2.3
+	 */
+	public void setProviderConfigurationHttpMessageConverter(HttpMessageConverter<OidcProviderConfiguration> providerConfigurationHttpMessageConverter) {
```
jgrandja, line 116: The thing about exposing `HttpMessageConverter<OidcProviderConfiguration>` is that it handles both read and write, but this `Filter` will only ever use the write. I wonder if we should just expose `Converter<OidcProviderConfiguration, Map<String, Object>>`, which ultimately hooks into `OidcProviderConfigurationHttpMessageConverter.setProviderConfigurationParametersConverter()`

ovidiupopa91: hi @jgrandja. I was thinking about this as well when I wrote the JavaDoc for the setter :-) 

I think it's a good idea to only expose the setter for the converter as the read method will never be used.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcProviderConfigurationEndpointFilter.java
```diff
@@ -103,6 +105,21 @@ protected void doFilterInternal(HttpServletRequest request, HttpServletResponse
...
+	/**
+	 * Sets the {@link Converter} used for converting the {@link OidcProviderConfiguration} to a
+	 * {@code Map} representation of the OpenID Provider Configuration.
+	 *
+	 * @param providerConfigurationParametersConverter the {@link Converter} used for converting to a
+	 * {@code Map} representation of the OpenID Provider Configuration
+	 *
+	 * @since 0.2.3
+	 */
+	public void setProviderConfigurationParametersConverter(
```
jgrandja, line 117: Let's rename to `setProviderConfigurationResponseConverter()`

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcProviderConfigurationEndpointFilterTests.java
```diff
@@ -71,6 +84,53 @@ public void doFilterWhenNotConfigurationRequestThenNotProcessed() throws Excepti
 
 		verify(filterChain).doFilter(any(HttpServletRequest.class), any(HttpServletResponse.class));
 	}
+	@Test
+	public void providerConfigurationParametersConverterWhenCustomThenAbleToOverride() throws Exception{
```
jgrandja, line 88: Please rename method to be consistent with other methods...should start with `doFilter...`

jgrandja (changes requested):
Thanks for the updates @ovidiupopa91. Please see review comments.

As well, we should add a customization hook in `OidcConfigurer`, for example `providerConfigurationResponseConverter()`

jgrandja:
Hi @ovidiupopa91. I'm just circling back to this PR now. Are you still available to work on this? No worries if you don't have time. Please let me know either way. Thanks.

jgrandja:
Hi @ovidiupopa91. I'm assuming you've been busy so I went ahead and merged this enhancement via 0994a1e1e14263d106373e3397f8c5c44f9a7708.

Issue #538, opened 2021-12-17
How-to: Authenticate using social login


jgrandja:
Publish a guide on How-to: Authenticate using social login, e.g. Google

Related gh-499, gh-106

shanzhaozhen:
This method can only be used for login, but cannot be used to register new users or bind old users

Issue #499, opened 2021-11-17
Add reference documentation


jgrandja:
This issue will serve to provide a content outline for the reference documentation.

Our goal is to avoid the "fluff" and only write content that will be valuable to our users and community. The only way to achieve this is too directly involve our users in driving out the content they want to see.

**We'll be prioritizing the "How-to" Guides based on upvotes.** Please log an issue for a new How-to Guide if it's not already listed.

We also encourage users to log an issue recommending new content that is not already listed in the proposed [content outline](https://github.com/spring-projects/spring-authorization-server/issues/499#issuecomment-989022953).

NotFound403:
Architecture Pic

mabujaber:
- Best practice section for web, mobile, server, and ..etc

sjohnr:
Since I missed linking the commit, see 525eca63d247478b0f4bbc5aed2d998eef70512f (initial draft of How-to: Implement the core services with JPA)

NotFound403:
Relation between Spring Security and Spring Authorization Server

mabujaber:
there is alot to do in the documentation. I started working on an authorization server integrated with spring security authentication and found a lot of stuff need to be clarified such as
1- phone verification page after the user login 
2- how to specify multiple login pages based on the client_id 
3- how to make multi-tenancy and this mostly will be done through an extra table linked with clients and users. 

some idea to help people start their project, I'll contribute my findings for any of above points

colin-riddell:
Where will the documentation end up being hosted? Is there any way to review work-in-progress docs and help out with contribution to that?

sjohnr:
Eventually, it will be hosted on docs.spring.io, but it might be a bit before we automate pushing documentation artifacts out there.

If you'd like to review progress, check out the [`docs`](https://github.com/spring-projects/spring-authorization-server/tree/main/docs) directory. You can run `npm install` and then `antora local-antora-playbook.yml` to build the site, then launch `docs/build/site/index.html` in your browser.

Not sure we're quite ready for lots of documentation contributions to begin flowing in yet, as we're still sorting out how it's going to look and nothing has settled yet. The most useful contributions at this point will be:

* Suggestions for content and guides, and up-votes to help us prioritize the ones we work on
* Feedback on any drafted content so we know whether we're on the right track (for example, "This was helpful..." or "The guide seems to be missing this piece of information..." etc.)
* Quick/easy updates such as typo fixes, etc.

[...]

Issue #106, opened 2020-09-01
Add a login with Google Authorization Server Sample


asaikali:
The current sample authorization server uses spring security form login. A sample showing how to have the authorization server delegate to google or another social login provider is quite useful. For my use case I to be able to support from based login and social OIDC logins. Implementing this sample can help ensure that the configuration APIs are flexible enough to allow multiple authentication providers.

erikmartino:
There should probably be two different where the user can choose between them, for example Google and Github. If you host a SAAS solution, it is quite useful to delegate authentication to each customer organisations own OIDC provider.

ojhughes:
I was thinking about this, should the authorization server ever delegate authorization? I understand the authorization server  would delegate the storage of users / credentials to LDAP or a DB but I thought that the process of authorizing a client should be performed by the authorization server. The existing Spring resource server already works with the various social logins.

jgrandja:
@ojhughes 

> should the authorization server ever delegate authorization

There are a number of OAuth/OIDC providers that deliver this type of feature via Social login authorization delegation. This is typically delivered in a product, so it's not clear if we would deliver such support within the framework.

Either way, it will be a good exercise to deliver a sample that demonstrates this authorization delegation pattern.

pangyiwei:
I had tried something similar to this out. What I did was rather straightforward:

1. Add spring-boot-starter-oauth2-client to the authorization project dependencies
2. Configure the default security filter chain with the oauth2Login().
3. Update `application.yml ` with the oauth2 client registration config (With your Google OAuth2 Credentials)

Would this be right for the sample? Would be glad to submit a PR for this if necessary

jgrandja:
@pangyiwei The sample you created does not integrate Authorization Server. It simply uses client features only. We would need the authorization related data to be stored via `OAuth2AuthorizationService`. The sample you have stores the data via `OAuth2AuthorizedClientService`.

pangyiwei:
@jgrandja Ahh I see. I had the impression that this could be done similarly to how the sample was for the form login where the resource owner will have to be authenticated (by Google) first before the authorization is saved by `OAuth2AuthorizationService`. Noted that this was the flow done in `processAuthorizationRequest()` in the `OAuth2AuthorizationEndpointFilter` class. Hence, I thought that the sample would mainly involve configuration of Google as an OAuth2 client (for the authorization server).

Thank you for the clarification.

natami:
Any progress/timeline on this issue?

[...]

PR #641, opened 2022-02-28
Federated Identity sample

sjohnr:
Issue gh-538
Issue gh-499
Issue gh-106

### DIFF
--- a/samples/federated-identity-authorizationserver/src/main/java/sample/security/FederatedIdentityConfigurer.java
+++ b/samples/federated-identity-authorizationserver/src/main/java/sample/security/FederatedIdentityConfigurer.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright 2020-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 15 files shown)


samples/federated-identity-authorizationserver/src/main/java/sample/security/FederatedIdentityConfigurer.java
```diff
@@ -0,0 +1,141 @@
...
+					);
+				}
+			});
+	}
+	// @formatter:on
+
+	@Override
+	public void setBuilder(HttpSecurity http) {
+		// Set the default the OAuth2TokenCustomizer so the OAuth2AuthorizationServerConfigurer can use it
+		http.setSharedObject(OAuth2TokenCustomizer.class, idTokenCustomizer());
```
jgrandja, line 132: With the recent changes of opaque access token support, `OAuth2TokenCustomizer` is no longer a shared object and instead is discovered as an optional `@Bean`. Register `FederatedIdentityIdTokenCustomizer` as a `@Bean` and should be good to go.

jgrandja (changes requested):
Thanks for the updates @sjohnr. Please see review comments.

jgrandja:
Thanks @sjohnr ! This is now in main via 3fe6f866598068326f2aadc25779a7402a8b046e

PR #646, opened 2022-03-14
Use port from configuration in the sample

akuhtz:
This allows to change the port in the application.properties and use the sample server.


jgrandja:
Thanks for the PR @akuhtz.

Allowing for this dynamic configuration via property would not work with the proposed change since `messages-client` and `messages-resource` sample are both configured with `issuer-uri: http://localhost:9000`.

Also, there are a few other hardcoded references to the `9000` port that would require updates as well, e.g. README, `custom-consent-authorizationserver` and `federated-identity-authorizationserver`.

The samples are meant to be as simple as possible with the minimum amount of code required to get up and running. Users can take the samples as a starting point and modify as necessary.

We're going to keep the default port as `9000` for all samples in order to keep things simple.

PR #693, opened 2022-04-20
jdk version is incorrect in Prerequisities

appchemist:
when `JAVA_HOME` environment variable points to `jdk1.8.0` folder & ./gradlew build, I met the below error.
```
> Task :buildSrc:compileJava FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':buildSrc:compileJava'.
> invalid source release: 11

* Try:
> Run with --stacktrace option to get the stack trace.
> Run with --info or --debug option to get more log output.
> Run with --scan to get full insights.

* Get more help at https://help.gradle.org

BUILD FAILED in 1s

```

After `JAVA_HOME` environment variable points to `jdk11` folder, build success

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -52,7 +52,7 @@ In the instructions below, https://vimeo.com/34436402[`./gradlew`] is invoked fr
 a cross-platform, self-contained bootstrap mechanism for the build.
 
 === Prerequisites
-https://help.github.com/set-up-git-redirect[Git] and the https://www.oracle.com/technetwork/java/javase/downloads[JDK8 build].
+https://help.github.com/set-up-git-redirect[Git] and the https://www.oracle.com/technetwork/java/javase/downloads[JDK11 build].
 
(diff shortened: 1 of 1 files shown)


appchemist:
sourceCompatibility is also `JavaVersion.VERSION_11` in buildSrc/build.gradle

sjohnr:
Hi @appchemist, thanks for reaching out! As part of upgrading to Gradle 7, we are taking the time to make the project consistent with Spring Security, which currently requires Java 11 for the `buildSrc` folder that powers the Gradle build. So while the project itself still only requires JDK 8, moving forward I think we will require JDK 11 to build the project. Can you please try upgrading to JDK 11 and see if the problem is resolved?

sjohnr:
Oops, @appchemist, I didn't realize this was a PR and thought it was a support issue. Looks like you already knew what I said, so you can disregard above.

PR #710, opened 2022-04-29
Add token revocation endpoint to OpenID Provider Configuration endpoint 

### LINKED ISSUE #687: OpenID Provider Configuration endpoint should include the revocation token endpoint
**Expected Behavior**
The OpenID Connect Discovery endpoint should provide the revocation token endpoint.
i.e. adding this:
"revocation_endpoint": "http://{host_issuer}:{port_issuer}/oauth2/revoke"

**Current Behavior**
The current response for /.well-known/openid-configuration does not contain the endpoint revocation.

**Context**
Some angular lib such as angular-auth-oidc-client expect the revocation endpoint to be in the response when requesting the OpenID Connect Discovery endpoint.

appchemist:
fix #687

Revocation endpoint filter is already implemented
So only add revocation token endpoint at OpenId Provider Configuration endpoint


jgrandja:
Thanks for the PR @appchemist ! This is now in main.

PR #738, opened 2022-05-12
Make the default scope empty for client_credentials grant

yonyes:
If the Request doesn't specify scopes, return the token without them.

Before: If the scope parameter is empty or missing, the token is generated with all possible scopes.

After: The received token will be without any token if there are no scopes in the token request.

Issue gh-780


jgrandja:
Thanks for the PR @yonyes. This is a breaking change so we'll apply it in the `0.4.0` release. I'll revisit this PR when we start working on `0.4.0`.

9527286768 (approved):

lucwillems:
i have commented on GH-780 about this. for me this behaviour is required. so could we make this configurable on client level so we can determine the behaviour where needed. 

something like settings.client.enable-default-scoping=true|false

jgrandja:
@yonyes Thanks again for the PR. This is now merged. FYI, I added a minor polish commit.

Issue #567, opened 2022-01-05
Access token with custom claims having value Long is throwing error: The class with java.lang.Long and name of java.lang.Long is not in the allowlist.


Ravis22:
**Describe the bug**
Custom access token with claim having value as Long is not getting revoked as during deserialization of access token object mapper throws error.

**To Reproduce**
User this access token to verify the deserialisation is failing
{"@class":"java.util.Collections$UnmodifiableMap","metadata.token.claims":{"@class":"java.util.Collections$UnmodifiableMap","sub":"535","user-authorities":["java.util.ArrayList",["ROLE_AUTHENTICATED"]],"iss":["java.net.URL","http://xxxxxxxl"],**"userId":["java.lang.Long",535]**,"aud":["java.util.Collections$SingletonList",["messaging-client"]],"nbf":["java.time.Instant",1641370496.488606000],"permissions":["java.util.HashSet",[]],"scope":["java.util.Collections$UnmodifiableSet",["message.read"]],"name":"abcd","exp":["java.time.Instant",1641374096.488606000],"iat":["java.time.Instant",1641370496.488606000],"email":"abcd@gmail.com"},"metadata.token.invalidated":false}

**Expected behavior**
Access token should be properly deserialized and access token should be revoked.

**Exception:**

java.lang.IllegalArgumentException: The class with java.lang.Long and name of java.lang.Long is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details
	at org.springframework.security.oauth2.server.authorization.JdbcOAuth2AuthorizationService$OAuth2AuthorizationRowMapper.parseMap(JdbcOAuth2AuthorizationService.java:441)
	at org.springframework.security.oauth2.server.authorization.JdbcOAuth2AuthorizationService$OAuth2AuthorizationRowMapper.mapRow(JdbcOAuth2AuthorizationService.java:370)
	at org.springframework.security.oauth2.server.authorization.JdbcOAuth2AuthorizationService$OAuth2AuthorizationRowMapper.mapRow(JdbcOAuth2AuthorizationService.java:308)
	at org.springframework.jdbc.core.RowMapperResultSetExtractor.extractData(RowMapperResultSetExtractor.java:94)
[...]

Issue #397, opened 2021-08-16
Getting error java.lang.IllegalArgumentException For Authorization Code flow with JDBCUserDetailService, JDBC Registered Client and JDBCAutjorization Service: The class with pk.training.basit.jpa.entity.UserPrincipal and name of pk.training.basit.jpa.entity.UserPrincipal is not in the allowlist.


Basit-Mahmood:
**Describe the bug**
With JDBCUserDetailService, JdbcRegisteredClientRepository with AuthorizationGrantType.AUTHORIZATION_CODE and JdbcOAuth2AuthorizationConsentService. While fetching the token will result in following error

`java.lang.IllegalArgumentException: The class with pk.training.basit.jpa.entity.UserPrincipal and name of pk.training.basit.jpa.entity.UserPrincipal is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details

	at org.springframework.security.oauth2.server.authorization.JdbcOAuth2AuthorizationService$OAuth2AuthorizationRowMapper.parseMap(JdbcOAuth2AuthorizationService.java:441) ~[spring-security-oauth2-authorization-server-0.1.2.jar:0.1.2]
	at org.springframework.security.oauth2.server.authorization.JdbcOAuth2AuthorizationService$OAuth2AuthorizationRowMapper.mapRow(JdbcOAuth2AuthorizationService.java:337) ~[spring-security-oauth2-authorization-server-0.1.2.jar:0.1.2]
	at org.springframework.security.oauth2.server.authorization.JdbcOAuth2AuthorizationService$OAuth2AuthorizationRowMapper.mapRow(JdbcOAuth2AuthorizationService.java:308) ~[spring-security-oauth2-authorization-server-0.1.2.jar:0.1.2]
	at org.springframework.jdbc.core.RowMapperResultSetExtractor.extractData(RowMapperResultSetExtractor.java:94) ~[spring-jdbc-5.3.9.jar:5.3.9]
	at org.springframework.jdbc.core.RowMapperResultSetExtractor.extractData(RowMapperResultSetExtractor.java:61) ~[spring-jdbc-5.3.9.jar:5.3.9]
[...]

PR #743, opened 2022-05-18
Fix deserialize incompatibility with Okta ver claim

kinsersh:
Okta includes a "ver" claim in the ID tokens it issues that gets serialized to json, as follows:
{
  "@class": "java.util.Collections$UnmodifiableMap",
  "ver": [
    "java.lang.Long",
    1
  ]
}

This claim is added by Okta automatically (see https://developer.okta.com/docs/reference/api/oidc/#id-token).

Spring Authorization Server encounters an error when attempting to deserialize this claim from Okta, as follows:
The class with java.lang.Long and name of java.lang.Long is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details
java.lang.IllegalArgumentException: The class with java.lang.Long and name of java.lang.Long is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details
	at org.springframework.security.jackson2.SecurityJackson2Modules$AllowlistTypeIdResolver.typeFromId(SecurityJackson2Modules.java:253)

The issue at https://github.com/spring-projects/spring-authorization-server/issues/567 has a similar symptom, though the situation being addressed in this pull request is different in that it would be encountered by any project wanting to do OpenID Connect with Okta. I thus recommend a general fix rather than a project-specific fix as recommended at https://github.com/spring-projects/spring-authorization-server/issues/397.

This commit defines a mixin that doesn't alter the json formatting but does instruct SecurityJackson2Modules.AllowlistTypeIdResolver to permit the deseralization of that Okta claim.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/jackson2/LongMixin.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/jackson2/LongMixin.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2020-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.jackson2;
+
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * This mixin class is used to serialize/deserialize {@link Long} for solving a compatibility issue with the Okta "ver"
+ * claim.
+ *
+ * @author Stephen Kinser
+ * @since 0.2.4
+ */
+abstract class LongMixin {
+	@SuppressWarnings("unused")
+	@JsonProperty("long")
+	Long value;
+}
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/jackson2/OAuth2AuthorizationServerJackson2Module.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/jackson2/OAuth2AuthorizationServerJackson2Module.java
@@ -82,6 +82,7 @@ public class OAuth2AuthorizationServerJackson2Module extends SimpleModule {
 		context.setMixInAnnotations(SignatureAlgorithm.class, JwsAlgorithmMixin.class);
 		context.setMixInAnnotations(MacAlgorithm.class, JwsAlgorithmMixin.class);
 		context.setMixInAnnotations(OAuth2TokenFormat.class, OAuth2TokenFormatMixin.class);
+		context.setMixInAnnotations(Long.class, LongMixin.class);
 	}
 
 }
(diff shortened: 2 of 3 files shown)


sjohnr:
Hi @kinsersh, welcome to the project!

It sounds like you're using Okta as your identity provider. Is this in the context of federated identity (similar to the [federated identity sample](https://github.com/spring-projects/spring-authorization-server/tree/main/samples/federated-identity-authorizationserver))?

Is there a reason you can't add this mixin yourself as in, for example, [this comment](https://github.com/spring-projects/spring-authorization-server/issues/397#issuecomment-900148920)?

kinsersh:
> Hi @kinsersh, welcome to the project!
> 
> It sounds like you're using Okta as your identity provider. Is this in the context of federated identity (similar to the [federated identity sample](https://github.com/spring-projects/spring-authorization-server/tree/main/samples/federated-identity-authorizationserver))?
> 
> Is there a reason you can't add this mixin yourself as in, for example, [this comment](https://github.com/spring-projects/spring-authorization-server/issues/397#issuecomment-900148920)?

Yes, I'm consuming Okta and others. I have added that mixin per that comment to work around this issue - it works well. I get it that Spring Authorization Server should stay free of project-specifics

Here are my reasons for submitting this pull request:
1. Reduce friction for others that also want to integrate Okta with Spring Authorization Server, given how widespread Okta is used
2. Avoid customizing the internals of Spring Authorization Server to simplify maintenance and improve future version compatibility. While customizing the Object Mapper isn't difficult, one does have to go through a few layers to get to it. If this pull request is included in the next version of Spring Authorization Server, I can remove my work around.

sjohnr:
Thanks for the feedback and your response @kinsersh! I see your points and while, as you mention, we want to stay free of embedding any customization needed for a specific use case in the core (for the time being), there are two sides to every coin such as increased friction and developer experience in this case.

We could open an issue to outline some general improvements to the `OAuth2AuthorizationServerJackson2Module`, which could include generic support for primitives/boxed types like `long`/`Long`. I will have to take this back and discuss it first, but it doesn't seem like we'd merge this as-is since it's too narrow a fix. I imagine we'd either fix it all, or not at all.

> If this pull request is included in the next version of Spring Authorization Server, I can remove my work around.

It's a small detail, but in this case I'm not sure it's a workaround so much as a customization. So while we want to make customization possible, there are lots of examples where it will be required to customize, as that's the framework's entire purpose to some degree. Just wanted to clarify that point in case it's relevant later.

For now, I'll leave this open as a placeholder for that more general issue I want to open, once we've discussed it a bit.

kinsersh:
> Thanks for the feedback and your response @kinsersh! I see your points and while, as you mention, we want to stay free of embedding any customization needed for a specific use case in the core (for the time being), there are two sides to every coin such as increased friction and developer experience in this case.
> 
> We could open an issue to outline some general improvements to the `OAuth2AuthorizationServerJackson2Module`, which could include generic support for primitives/boxed types like `long`/`Long`. I will have to take this back and discuss it first, but it doesn't seem like we'd merge this as-is since it's too narrow a fix. I imagine we'd either fix it all, or not at all.
> 
> > If this pull request is included in the next version of Spring Authorization Server, I can remove my work around.
> 
> It's a small detail, but in this case I'm not sure it's a workaround so much as a customization. So while we want to make customization possible, there are lots of examples where it will be required to customize, as that's the framework's entire purpose to some degree. Just wanted to clarify that point in case it's relevant later.
> 
> For now, I'll leave this open as a placeholder for that more general issue I want to open, once we've discussed it a bit.

Sounds good. I'm all for a general fix. What I have here is a narrow fix - solves a compatibility issue with Okta right now, though what's the likelihood that Okta or another provider releases a change to their ID token format that uses a different primitive class for serializing/deserializing.

jgrandja:
@kinsersh Thanks for all the details. As discussed, there are customization points that allow an application to configure serialization/deserialization for JSON types. Adding `Long` doesn't make sense for the framework since it is not used directly by the framework and would only be used for Okta specific flows.

Furthermore, adding this narrow change will open us to adding other narrow changes requested by users (in the future), and ultimately forcing us to maintain more code.

I see this as a minor customization on your end and I'm assuming there are other customizations required to federate Spring Authorization Server with Okta, so they all fall into that same group of customizations required to integrate with Okta.

Based on the explanation above, we cannot accept this change as it's not required for the framework specific code.

PR #768, opened 2022-06-08
Add authenticationDetailsSource to AuthorizationEndpointFilter

Kehrlann:
We do audit logging on calls made against our auth server. When we get un-authenticated requests to the `/oauth2/authorize/` endpoint, specifically invalid requests, we want to track where the request came from.

Therefore, we want the web auth details set on the anonymous authentication that `OAuth2AuthorizationCodeRequestAuthenticationConverter` creates.

This is a straightforward stab at it. 

However there might be more considerations at play here. Maybe the `OAuth2AuthorizationEndpointFilter` be after the `AnonymousAuthenticationFilter` if present? Maybe there should be an extension point for creating the anonymous authentication?

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
@@ -28,6 +28,7 @@ import javax.servlet.http.HttpServletResponse;
 import org.springframework.http.HttpMethod;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.MediaType;
+import org.springframework.security.authentication.AuthenticationDetailsSource;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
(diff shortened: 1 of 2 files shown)


jgrandja:
@Kehrlann I think we should apply the same change to `OAuth2AuthorizationEndpointFilter` as was applied to `OAuth2TokenEndpointFilter` in these commits https://github.com/spring-projects/spring-authorization-server/commit/b62b161b95a603a0bc5baed7a9ac21691b119192 https://github.com/spring-projects/spring-authorization-server/commit/c9954af084693876712e8c3735e01c1cff74851a

Kehrlann:
@jgrandja Updated according to your comment. It slipped my mind that we did this in the token endpoint.

I'm still curious though - shouldn't the `OAuth2AuthorizationEndpointFilter` come after the AnoynmousAuthenticationFilter in any case, so that (in most cases) we don't need to populate the SecurityContext in `OAuth2AuthorizationCodeRequestAuthenticationConverter`?

jgrandja:
Thanks for the updates @Kehrlann ! This is now in main.

> Shouldn't the `OAuth2AuthorizationEndpointFilter` come after the `AnoynmousAuthenticationFilter`

Take a look a [Section 4.1.1](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-05#section-4.1.1).  Authorization Request:

> The authorization server validates the request to ensure that all
   required parameters are present and valid.
   ...
   If the request is valid, the authorization server authenticates the
   resource owner and obtains an authorization decision (by asking the
   resource owner or by establishing approval via other means).

The authorization request must be validated first. If all parameters are valid then authentication occurs (if not already authenticated). This is why it's before `AbstractPreAuthenticatedProcessingFilter` to ensure authentication happens after request validation. Hope this makes sense?

Issue #783, opened 2022-06-15
allow refresh_token types thru org.springframework.security.oauth2.server.authorization.token.JwtGenerator


andre-fxcm:
I'd like to do something like the following to customize refresh_tokens as JWT and leverage JwtGenerator to do it
```java
@Slf4j
@RequiredArgsConstructor
public class RefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
    private final JwtGenerator jwtGenerator;

    @Override
    public OAuth2RefreshToken generate(OAuth2TokenContext context) {
        Jwt encode = jwtGenerator.generate(context);
        return new OAuth2RefreshToken(encode.getTokenValue(), encode.getIssuedAt());
    }
}
```

But JwtGenerator doesnt allow "refresh_token" types thru its generate method. Any chance we can allow this possibility?
https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtGenerator.java#L74-L82

andre-fxcm:
possible solution
```java
@Slf4j
@RequiredArgsConstructor
public class RefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
    private final JwtGenerator jwtGenerator;

    @Override
    public OAuth2RefreshToken generate(OAuth2TokenContext context) {
        Jwt encode = jwtGenerator.buildJwt(context);
        return new OAuth2RefreshToken(encode.getTokenValue(), encode.getIssuedAt());
    }
}
```
https://github.com/spring-projects/spring-authorization-server/pull/784

jgrandja:
@andre-fxcm Refresh tokens are typically opaque. Why do you need refresh tokens in Jwt format?

FYI, you can provide your own custom `OAuth2TokenGenerator<OAuth2RefreshToken>` that uses `JwtEncoder` instead of `JwtGenerator`. See [reference documentation](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/core-model-components.html#oauth2-token-generator) on how to configure.

andre-fxcm:
@jgrandja Hi, thanks for the feedback. 
I have use case as follows for jwt on the refresh_token:

When a user authenticates successfully, an operation happens on my side and that unique generated value is persisted as a claim into the refresh token.

Every time a new access token is created from that refresh token, that claim should statelessly transfer to the access token

[...]

PR #784, opened 2022-06-15
make building of JWT independent of Token Generation logic

andre-fxcm:


jgrandja:
@andre-fxcm There is no need to expose `buildJwt(OAuth2TokenContext context)` as there is an alternative implementation strategy as mentioned in this [comment](https://github.com/spring-projects/spring-authorization-server/issues/783#issuecomment-1156173336). Let's discuss there first before we decide a PR is needed.

Issue #642, opened 2022-03-04
Allow configuration for authorization code time-to-live


guo-jun-airwallex:
**Expected Behavior**
Should allow different clients to customize different authorization code time-to-live

**Current Behavior**
The authorization code time-to-live is always 5 minutes.

**Context**
```
private OAuth2AuthorizationCode generateAuthorizationCode() {
		Instant issuedAt = Instant.now();
		Instant expiresAt = issuedAt.plus(5, ChronoUnit.MINUTES);		// TODO Allow configuration for authorization code time-to-live
		return new OAuth2AuthorizationCode(this.authorizationCodeGenerator.get(), issuedAt, expiresAt);
	}
```
The TODO in `generateAuthorizationCode` is already existed in two minor version.
Can I know the ETA of this feature.
I hope there is a `getAuthorizationCodeTimeToLive` method in `TokenSettings` class to customize authorization code time-to-live.

arunsah:
Hi Guo,
Not sure, but seem like you might be looking for this:

```java
// client is simple DTO fetched from DB

RegisteredClient.Builder registeredClientBuilder = RegisteredClient.withId(client.getId())
        .clientId(client.getClientName()).clientName(client.getClientName()).clientSecret(client.getClientSecret());
// some other settings

var tokenSettingsBuilder = TokenSettings.builder();

// access token time to live settings
tokenSettingsBuilder.accessTokenTimeToLive(Duration.ofSeconds(client.getAccessTokenTimeToLiveInSeconds()));
tokenSettingsBuilder.refreshTokenTimeToLive(Duration.ofSeconds(client.getRefreshTokenTimeToLiveInSeconds()));
tokenSettingsBuilder.reuseRefreshTokens(client.isReuseRefreshTokens());

registeredClientBuilder.tokenSettings(tokenSettingsBuilder.build());

RegisteredClient registeredClient =  registeredClientBuilder.build();

```

Thanks,
Arun.

jgrandja:
Thanks @guo-jun-airwallex. We'll look at adding this soon.

jgrandja:
The issue is yours @gourav. Please review the [Getting Started](https://github.com/spring-projects/spring-authorization-server#getting-started) in case you haven't already. Thanks.

gourav:
Hi @jgrandja 
Unfortunately, I am not able to contribute due to health issues I am facing currently. 

I apologise for the delay and inconvenience. 

[...]

PR #786, opened 2022-06-19
Add the time-to-live config for an authorization code at TokenSettings

appchemist:
Add the time-to-live configuration for authorization code at TokenSettings in the same level as other tokens(access token, refresh token)
issue #642


jgrandja:
Thanks for the PR @appchemist ! This is now in main.

appchemist:
Thanks for polishing code! @jgrandja 
I will pay attention more.

PR #787, opened 2022-06-20
Use configured ID Token signature algorithm

721806280:


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtGenerator.java
```diff
@@ -125,7 +125,7 @@ public Jwt generate(OAuth2TokenContext context) {
 		}
 		// @formatter:on
 
-		JwsHeader.Builder headersBuilder = JwsHeader.with(SignatureAlgorithm.RS256);
+		JwsHeader.Builder headersBuilder = JwsHeader.with(registeredClient.getTokenSettings().getIdTokenSignatureAlgorithm());
```
jgrandja, line 128: `registeredClient.getTokenSettings().getIdTokenSignatureAlgorithm()` should only be used when minting an ID Token. The existing default should be preserved for access tokens.

jgrandja (changes requested):
Thanks for the PR @721806280. Please see review comment.

jgrandja (changes requested):
Thanks for the updates @721806280. Please see review comments.

Also, please rebase off of `0.4.x` and squash the commits on next push. Thanks.

jgrandja (changes requested):
Thanks for the updates @721806280. Please see the one review comment.

In addition to the one requested change, please add a test in `JwtGeneratorTests` to ensure the defaults are set as expected. Also, please rebase this PR off of `0.4.x` branch and squash the commits.
Thanks.

jgrandja:
Thanks for the PR @721806280. This is now merged.

FYI, I added a polish commit with some minor updates.

PR #788, opened 2022-06-20
Update Gradle Enterprise plugin

jprinet:


jgrandja:
Thanks for the update @jprinet. This is now in main.

PR #793, opened 2022-06-27
Add examples that support WeChat Applets

xuxiaowei-com-cn:
Custom authorization method grant_type is wechat_applet


jgrandja:
Thanks for the PR @xuxiaowei-com-cn.

There is quite a bit of custom code in this sample that we would not want to maintain in the main repository.

As well, I do recall working on a couple of GitHub issues in the Spring Security `oauth2-client` module related to WeChat integration and discovered at that time that WeChat does not fully conform to the OAuth2 specifications. One example is that it uses `appid` instead of `clientId` in the auth flows. This is another reason why we do not want to maintain a sample application that deviates from standard OAuth2 spec.

I would suggest that you setup a branch on your fork with this sample. Then you can point users to your branch (from this PR) so they can see how the integration works.

xuxiaowei-com-cn:
支付宝小程序、支付宝开放平台、钉钉dingtalk、飞书平台网页应用、码云gitee、github、gitlab、QQ开放平台、QQ小程序平台、微信小程序、微信公众平台、微信开放平台、企业微信平台、微博weibo平台等等参见：
https://github.com/xuxiaowei-com-cn/spring-boot-starter

PR #802, opened 2022-07-13
Update custom consent page sample

wingzero0:
wrap custom consent page action link with thymeleaf syntax, because server.servlet.context-path might not be root;


jgrandja:
Thanks for the PR @wingzero0 ! This is now in main.

PR #807, opened 2022-07-21
Spring cloud gateway example

salutonmondo:
add a simple example of integrating spring authorization server with spring cloud gateway.


sjohnr:
@salutonmondo thanks for your interest in the project!

Just an FYI that we already have a similar [sample branch](https://github.com/sjohnr/spring-authorization-server/tree/bff-demo/samples/backend-for-frontend-client) that we've been using that goes along with a semi-recent [webinar](https://tanzu.vmware.com/content/webinars/mar-10-getting-started-with-spring-authorization-server). We may eventually decide to merge the backend-for-frontend-client sample into this repository but don't have a plan to do so any time soon.

You'll want to take a look at that sample and compare/contrast what you've done here, though it's quite nice and I appreciate the effort. However, do make sure to reach out through an issue in the future before doing any work on submitting a PR as we don't want anyone wasting or duplicating effort. With that in mind, I'm going to go ahead and close this PR for now, but feel free to let me know if you have any other thoughts.

khauser:
I think the main difference here is that the UI of an application moves behind a gateway which is caring about user authentication together with the autorization server. In his PR the UI is represented by the resources server in having a secured and an unsecured public endpoint.
The BFF (thanks for the nice and comprehensive webinar!) approach doesn't cover this way of using a gateway in my opinion and I would love to have that sample available. Okay at least I can copy it from the PR for now.

sjohnr:
@khauser thanks for sharing your thoughts on this. I agree that there are many different ways of architecting these types of applications and they are all slightly different. I imagine it would be quite difficult to have a distinct sample for all of the different deployment models and architectures, so we would tend to want to maintain only the primary ones that demonstrate the largest number of key concepts. Most of the time it's only small variations from there to get the end result you're after. Having said that, we'll definitely continue to listen, and if many are bumping up against similar challenges with integrating we can always add an official sample after some discussion.

uniquejava:
@sjohnr Hello, the webinar is quite inspiring and the `Back component pattern` is what we applied in our current project. There is one problem which lead us to think about using spring cloud gateway. 

In the webinar demo, looks like for each and every API from resource server, we have to plugin/proxy it somewhere in the BFF controller. 

Currently we have just 4 micro services(resource servers) with about 80 APIs.  Do we have to duplicate these 80 APIs in BFF controllers.  That's a lot of work, isn't it?  And quickly the spring cloud gateway comes to my mind. We'd need it to do some simple route for us and so we don't have to duplicate these 80 APIs in BFF controllers,  yet we'd like the spring cloud gateway do the heavy authz work for us as an oauth2 client .

So we really appreciate someone could do one demo for this kind of scenario.   🥲 

I searched stack overflow and find similar questions like the following one but no one answer it. 

[How to use spring-cloud-gateway for authentication and authorization?](https://stackoverflow.com/questions/72495825/how-to-use-spring-cloud-gateway-for-authentication-and-authorization)

sjohnr:
Hi @uniquejava. In the webinar, we didn't have time to demonstrate using the `TokenRelay` filter in Spring Cloud Gateway as one way of adding routing to the gateway that can successfully make calls to resource servers that sit behind the gateway. However, the commits of the [sample branch](https://github.com/sjohnr/spring-authorization-server/tree/bff-demo/samples/backend-for-frontend-client) demonstrate this.

It's worth noting that the existing `TokenRelay` filter has a limitation in that it can only use the token available for the current authenticated user that comes from `.oauth2Login()`. An enhancement could be added to overcome this limitation, but hasn't been added yet.

Does that answer your question?

uniquejava:
@sjohnr Wow,  you are that "BFF(Best Friend Forever)" teacher in that webinar. Thank you for your reply, I did a lot of homework using the keyword you provided and also referred to the sample branch. I was able to successfully set up the example using Spring Authorization Server, Gateway and Vue3(vite)

Thanks a lot.

I cleared the following obstacles along the way.
1. WebClient API: the so called reactive RestTemplate which I just realized it's being deprecated.
2. I thought UserDetailService is a must, then I found we can totally do authz without it, via the`TokenConverter > AuthenticationProvider` routine mentioned in the webinar.
3. I thought Token is just some bearer token, the all kinds of XxxxToken in spring security confused me for long time, I am now not confused about it any more 🥇 
4. The fancy ExchangeFilterFunction around WebClient, they are existing for handling acquire/check/renew token, etc, so we don't care about it.
5. I didn't even notice there are both `http.oauth2Login()` and `http.oauth2Client`, and for now I know we should use oauth2Login which will do the magic behind the scene, we don't even need to consider storing token in the frontend, the cookie based solution is just fine.

Still a lot to conquer but a least I know where to go. 🥂

PR #876, opened 2022-08-26
fix: Restructure URL encoding parameters to ensure proper encoding 

### LINKED ISSUE #875: State parameter does not handle plus sign properly
**Describe the bug**
The state parameter applies incorrect URL encoding when a plus sign is present in the state query parameter.

Probably related to spring-projects/spring-framework/issues/21577

**To Reproduce**
Initiate a standard OAuth2 authorization code flow with a state query parameter containing a plus sign. When the redirect happens with the authorization code, the state query parameter sent will not match the state query parameter that was provided on the initial authorize call. Per specification, the state value in the response must be the 'exact value received from the client' 
**Expected behavior**
Per the OAuth2 specification, the state value in the response must be the 'exact value received from the client' 
**Sample**
Attempt to initiate an OAuth2 authorization code flow with state like below (examples provided with URL parameter encoded and not encoded

With state component encoded:
```
http://localhost:8080/oauth2/authorize?audience=foo&scope=openid&response_type=code&client_id=tkk78aDthMrMkpJEphYOd66i5GMkWxkv&redirect_uri=https://jwt.io/&state=awrD0fCnEcTUPFgmyy2SU89HZNcnAJ60ZW6l39YI0KyVjmIZ%2B004pwm9j55li7BoydXYysH4enZMF21QFuVjThcHTL67sAj6ZPvmpQM6gao9wZQfRjXIyBYsBZ3N8H4R%2BwCKSR1ltgUenX9eWDL%2FKA%3D%3D

Resulting redirect:
https://jwt.io/?code=oKLM7rW-I1dnF4t92sqUcGubxNHgJ36OBjknkUyl5YogkMwEaIMDfuAi8MkBI-A3s3W4Gx4FvsrjJWLN7_5HIUMYVF7VN9t55A48nynqs3MMxXDJNZJV7YlLf7Zrn55K&state=awrD0fCnEcTUPFgmyy2SU89HZNcnAJ60ZW6l39YI0KyVjmIZ+004pwm9j55li7BoydXYysH4enZMF21QFuVjThcHTL67sAj6ZPvmpQM6gao9wZQfRjXIyBYsBZ3N8H4R+wCKSR1ltgUenX9eWDL/KA%3D%3D

```

Without state component encoded:
```
http://localhost:8080/oauth2/authorize?audience=foo&scope=openid&response_type=code&client_id=tkk78aDthMrMkpJEphYOd66i5GMkWxkv&state=awrD0fCnEcTUPFgmyy2SU89HZNcnAJ60ZW6l39YI0KyVjmIZ+004pwm9j55li7BoydXYysH4enZMF21QFuVjThcHTL67sAj6ZPvmpQM6gao9wZQfRjXIyBYsBZ3N8H4R+wCKSR1ltgUenX9eWDL/KA==&redirect_uri=https://jwt.io/

Resulting redirect:
https://jwt.io?code=izOXIFsvOEnA38FdxZLf66qx1G5EFFVENGwhbihuyRoq989kdfcQMk_bMWrypLyLJNenFGkGHTWTaC04NAwST3M2Lt04nSk6YAoRV-B341TVRVDbNR17I3NFXco5jY8L&state=awrD0fCnEcTUPFgmyy2SU89HZNcnAJ60ZW6l39YI0KyVjmIZ%20004pwm9j55li7BoydXYysH4enZMF21QFuVjThcHTL67sAj6ZPvmpQM6gao9wZQfRjXIyBYsBZ3N8H4R%20wCKSR1ltgUenX9eWDL/KA%3D%3D

```

backjo:
Fixes #875 

Restructure how URL encoding is done to ensure that plus signs in query parameters for the authorization code flow are encoded properly. This PR also updates the test value for state from "state" to a url string that requires encoding to be done, so that unit tests will catch encoding breakages in the future. 

Signed-off-by: Jonah Back <jonah@jonahback.com>

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
@@ -239,11 +239,13 @@ public final class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilte
 		String state = authorizationCodeRequestAuthenticationResult.getState();
 
 		if (hasConsentUri()) {
-			String redirectUri = UriComponentsBuilder.fromUriString(resolveConsentUri(request))
+			UriComponentsBuilder uriBuilder = UriComponentsBuilder.fromUriString(resolveConsentUri(request))
 					.queryParam(OAuth2ParameterNames.SCOPE, String.join(" ", requestedScopes))
 					.queryParam(OAuth2ParameterNames.CLIENT_ID, clientId)
-					.queryParam(OAuth2ParameterNames.STATE, state)
-					.toUriString();
+					.queryParam(OAuth2ParameterNames.STATE, "{state}");
+			HashMap<String, String> queryParams = new HashMap<>();
+			queryParams.put(OAuth2ParameterNames.STATE, state);
+			String redirectUri = uriBuilder.build(queryParams).toString();
 			this.redirectStrategy.sendRedirect(request, response, redirectUri);
 		} else {
 			DefaultConsentPage.displayConsent(request, response, clientId, principal, requestedScopes, authorizedScopes, state);
@@ -310,9 +312,12 @@ public final class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilte
 			uriBuilder.queryParam(OAuth2ParameterNames.ERROR_URI, error.getUri());
 		}
 		if (StringUtils.hasText(authorizationCodeRequestAuthentication.getState())) {
-			uriBuilder.queryParam(OAuth2ParameterNames.STATE, authorizationCodeRequestAuthentication.getState());
+			uriBuilder.queryParam(OAuth2ParameterNames.STATE, "{state}");
 		}
-		this.redirectStrategy.sendRedirect(request, response, uriBuilder.toUriString());
+		HashMap<String, String> queryParams = new HashMap<>();
+		queryParams.put(OAuth2ParameterNames.STATE, authorizationCodeRequestAuthentication.getState());
+		String redirectUri = uriBuilder.build(queryParams).toString();
+		this.redirectStrategy.sendRedirect(request, response, redirectUri);
 	}
 
 	/**
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
@@ -17,6 +17,7 @@ package org.springframework.security.oauth2.server.authorization.web;
 
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.Set;
 
@@ -273,11 +274,15 @@ public final class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilte
 				(OAuth2AuthorizationCodeRequestAuthenticationToken) authentication;
 		UriComponentsBuilder uriBuilder = UriComponentsBuilder
 				.fromUriString(authorizationCodeRequestAuthentication.getRedirectUri())
-				.queryParam(OAuth2ParameterNames.CODE, authorizationCodeRequestAuthentication.getAuthorizationCode().getTokenValue());
+				.queryParam(OAuth2ParameterNames.CODE, "{code}");
 		if (StringUtils.hasText(authorizationCodeRequestAuthentication.getState())) {
-			uriBuilder.queryParam(OAuth2ParameterNames.STATE, authorizationCodeRequestAuthentication.getState());
+			uriBuilder.queryParam(OAuth2ParameterNames.STATE, "{state}");
 		}
-		this.redirectStrategy.sendRedirect(request, response, uriBuilder.toUriString());
+		HashMap<String, String> queryParams = new HashMap<>();
+		queryParams.put(OAuth2ParameterNames.CODE, authorizationCodeRequestAuthentication.getAuthorizationCode().getTokenValue());
+		queryParams.put(OAuth2ParameterNames.STATE, authorizationCodeRequestAuthentication.getState());
+		String redirectUri = uriBuilder.build(queryParams).toString();
+		this.redirectStrategy.sendRedirect(request, response, redirectUri);
 	}
 
 	private void sendErrorResponse(HttpServletRequest request, HttpServletResponse response,
--- a/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/TestOAuth2Authorizations.java
+++ b/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/TestOAuth2Authorizations.java
@@ -39,44 +39,52 @@ import org.springframework.util.CollectionUtils;
 public class TestOAuth2Authorizations {
 
 	public static OAuth2Authorization.Builder authorization() {
-		return authorization(TestRegisteredClients.registeredClient().build());
+		return authorization(TestRegisteredClients.registeredClient().build(), "state");
+	}
+
+	public static OAuth2Authorization.Builder authorization(RegisteredClient registeredClient, String state) {
+		return authorization(registeredClient, Collections.emptyMap(), state);
 	}
 
 	public static OAuth2Authorization.Builder authorization(RegisteredClient registeredClient) {
-		return authorization(registeredClient, Collections.emptyMap());
+		return authorization(registeredClient, Collections.emptyMap(), "state");
 	}
 
 	public static OAuth2Authorization.Builder authorization(RegisteredClient registeredClient,
-			Map<String, Object> authorizationRequestAdditionalParameters) {
+			Map<String, Object> authorizationRequestAdditionalParameters, String state) {
 		OAuth2AuthorizationCode authorizationCode = new OAuth2AuthorizationCode(
 				"code", Instant.now(), Instant.now().plusSeconds(120));
 		OAuth2AccessToken accessToken = new OAuth2AccessToken(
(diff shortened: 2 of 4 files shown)


jgrandja:
Thanks for the PR @backjo. Can you please fix the failing tests.

Also, please change the `state` parameter in `OAuth2AuthorizationCodeGrantTests` to use a value that requires encoding.

backjo:
@jgrandja updated as per request.

jgrandja:
@backjo It appears there is no issue with the `state` parameter decoding.

The `+` character is reserved (see [RFC 3986 Section-2.2](https://www.rfc-editor.org/rfc/rfc3986#section-2.2)) and therefore **must be percent-encoded on the client** before redirecting to the authorization server.

For example, if the `state` parameter value is `state+12+34` then the percent-encoded value is `state%2B12%2B34`. **The percent-encoded value must be used in the authorization request URI**. The use case you provided is not percent-encoding the `state` parameter which triggers the error, however, the use case is not valid since percent-encoding was not applied to the reserved character.

Makes sense?

jgrandja:
@backjo

> It may be that the tests are a bit confusing here because it appears that the non-encoded state is being passed in as a parameter - this is due to `MockHttpServletRequestBuilder` performing encoding internally.

There are a couple of things happening here.

The first thing is if you want the parameters to be encoded then you need to use [MockHttpServletRequestBuilder.queryParams()](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/main/java/org/springframework/test/web/servlet/request/MockHttpServletRequestBuilder.java#L411) instead of `MockHttpServletRequestBuilder.params()`.

The second thing is if you use reserved characters in one of those query parameter values then you need to percent-encode it before passing it in as a query parameter. **NOTE:** `MockHttpServletRequestBuilder.queryParams()` uses [UriComponents.encode()](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/util/UriComponents.java#L136) internally to perform the encoding, however, as stated in the javadoc, **it does not perform encoding for reserved characters**:

> ... this method only replaces non-ASCII and illegal (within a given URI component type) characters, **but not characters with reserved meaning.**

Furthermore, I tested this out in the [default messages sample](https://github.com/spring-projects/spring-authorization-server/tree/main/samples) and the current implementation works as expected.

If you want to test it out yourself, updated [SecurityConfig](https://github.com/spring-projects/spring-authorization-server/blob/main/samples/messages-client/src/main/java/sample/config/SecurityConfig.java#L31) with the following:

```java
@EnableWebSecurity
public class SecurityConfig {
	@Autowired
	private ClientRegistrationRepository clientRegistrationRepository;

	@Bean
	WebSecurityCustomizer webSecurityCustomizer() {
		return (web) -> web.ignoring().antMatchers("/webjars/**");
	}

	// @formatter:off
	@Bean
	SecurityFilterChain securityFilterChain(HttpSecurity http) throws Exception {
		http
			.authorizeRequests(authorizeRequests ->
				authorizeRequests.anyRequest().authenticated()
			)
			.oauth2Login(oauth2Login ->
				oauth2Login
					.authorizationEndpoint(authorizationEndpoint ->
						authorizationEndpoint
							.authorizationRequestResolver(authorizationRequestResolver(this.clientRegistrationRepository)))
					.loginPage("/oauth2/authorization/messaging-client-oidc"))
			.oauth2Client(withDefaults());
		return http.build();
	}
	// @formatter:on

	private OAuth2AuthorizationRequestResolver authorizationRequestResolver(
			ClientRegistrationRepository clientRegistrationRepository) {

		DefaultOAuth2AuthorizationRequestResolver authorizationRequestResolver =
				new DefaultOAuth2AuthorizationRequestResolver(
						clientRegistrationRepository, "/oauth2/authorization");
		authorizationRequestResolver.setAuthorizationRequestCustomizer(
				authorizationRequestCustomizer());

		return authorizationRequestResolver;
	}

	private Consumer<OAuth2AuthorizationRequest.Builder> authorizationRequestCustomizer() {
		return customizer -> customizer.state("state%2B12%2B34");	// Set percent-encoded value of "state+12+34"
	}

}
```

The sample code demonstrates when the `messages-client` sends a percent-encoded value for the `state` parameter, the OpenID Connect Flow succeeds and therefore the `state` is matched on the callback.

I'm going to close this PR and associated issue as the current implementation works as expected.

backjo:
hi @jgrandja - I really appreciate you diving into this. 

Following your example with the messaging client, the following authorization URL was generated:
```text
http://localhost:9000/oauth2/authorize?response_type=code&client_id=messaging-client&scope=openid%20profile&state=state%252B12%252B34&redirect_uri=http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc&nonce=t5Zo5BgrBL3dYKk86S7a1-p3WaOxsvF6hH6n-z1N05s
```

This results in the following authorization code response
```text
http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc?code=HwnK80PpJWqHd_UueP3a2cvwzpUclIyb6Ura0sBhVSuOhzY5ZbN_bIJG2FK2Z1bVt6PPjNx5NjSZLmrKN0ccThBxylHDtVyhYYv65tRHwHR4JZdCer31leXNAh0XhH_g&state=state%252B12%252B34
```

This is correct behavior - the state in both is `state%252B12%252B34` 

When I submit the following authorization request 
```text
http://localhost:9000/oauth2/authorize?response_type=code&client_id=messaging-client&scope=openid%20profile&state=state%2B12%2B34&redirect_uri=http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc&nonce=t5Zo5BgrBL3dYKk86S7a1-p3WaOxsvF6hH6n-z1N0
```

The following authorization response is generated
```text
http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc?code=dVSyke5KdtMutdM2FPAUriErpM1Tnf1ifDBq4Uf-aOL0q0YSNgJ4ymEC-GkbgJ2osgHNhjuE32MxUsKqTKVOHqFkGSVONgjfJEUcsHvKQiTTa9J9WcOCOxWrMwWELPRL&state=state%2012%2034
```

The state in both is not the same - in the authorization request, it is `state%2B12%2B34` (`state+12+34` encoded, per specification 'If data for a URI component would conflict with a reserved
   character's purpose as a delimiter, then the conflicting data must be
   percent-encoded before the URI is formed.' ). 

In the authorization code redirect, it is `state%2012%2034` (`state 12 34` encoded). 

Is this correct behavior? I would expect that the state parameter in the authorization code response would be `state%2B12%2B34`

jgrandja:
@backjo 

> When I submit the following authorization request 
```text
http://localhost:9000/oauth2/authorize?response_type=code&client_id=messaging-client&scope=openid%20profile&state=state%2B12%2B34&redirect_uri=http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc&nonce=t5Zo5BgrBL3dYKk86S7a1-p3WaOxsvF6hH6n-z1N0
```

Again, this is not a valid authorization request. The `state` parameter value is **not URL encoded** - it is just percent encoded.

For example:

state+12+34 --> original value
state%2B12%2B34 --> percent encoded of reserved characters
state%252B12%252B34 --> URL encoded (double-encoded)

backjo:
@jgrandja - apologies if it feels like I'm beating a dead horse here, but I don't think this interpretation is correct. 

This issue (and the proposed PR fix) lines up with exactly what is discussed in https://github.com/spring-projects/spring-framework/issues/14464 and https://github.com/spring-projects/spring-framework/issues/21399. When sending the request to Spring Authorization Server as "state%2B12%2B34" in the request parameter, it is interpreted correctly as "state+12+34" when it's being handled - it's only when it's regenerating the redirect uri using UriComponentsBuilder, which as described in the issues referenced above, will not be encoded because "+" is an allowed character. In this case, we do want to encode the "+", because it's from a value that has already been decoded.

In the UriComponentsBuilder documentation, this behavior is explicitly called out (documentation written by @rstoyanchev - maybe he could also chime in here given his work on URL encoding in Spring Framework?) :
```
For most cases, this method is more likely to give the expected result because in treats URI variables as opaque data to be fully encoded, while [UriComponents.encode()](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponents.html#encode--) is useful when intentionally expanding URI variables that contain reserved characters.

For example ';' is legal in a path but has reserved meaning. This method replaces ";" with "%3B" in URI variables but not in the URI template. By contrast, [UriComponents.encode()](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/web/util/UriComponents.html#encode--) never replaces ";" since it is a legal character in a path.
```

In the existing implementation as it exists today (which invokes https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/web/util/UriComponentsBuilder.java#L544) , the URIComponents encode method is the one being used - which will never replace a legal character.  Since state is opaque data (per specification), we really do need to it encode it as it suggests we do in the UriComponentsBuilder documentation - which is what this PR is doing. 


For what it's worth, we went and tested multiple different OAuth2 authorization providers (both OSS and commercial), and they all act consistently here.

KeyCloak:
```text
http://localhost:8080/realms/test/protocol/openid-connect/auth?client_id=account&response_type=code&redirect_uri=https://jwt.io/&state=state%2B12%2B34
```
```text
https://jwt.io/?state=state%2B12%2B34&session_state=0c31923f-1366-497a-a597-cbee678d3168&code=1a76590d-8294-49d1-be3f-674e17a5113b.0c31923f-1366-497a-a597-cbee678d3168.3090c447-4f49-453c-ad1d-0796517e705b
```

Auth0
```text
https://<test-tenant>.auth0.com/authorize?&client_id=tkk78aDthMrMkpJEphYOd66i5GMkWxkv&response_type=code&redirect_uri=https://jwt.io&scope=openid+profile&state=state%2B12%2B34
```

```text
https://jwt.io/?code=Kvq7wcVCsZgzw6YOy5y6BOal1KitQoIN0b6VV4QJByhom&state=state%2B12%2B34
```

Even google handles this the same way on search queries:

This searches for state+12+34
```text
https://www.google.com/search?q=state%2B12%2B34
```

This searches for state%2B12%2B34
```text
https://www.google.com/search?q=state%252B12%252B34
```

PR #915, opened 2022-09-27
Modify the access modifier of OAuth2EndpointUtils to public

Hccake:
`OAuth2EndpointUtils` is widely used in `AuthenticationConverter`, but now the default access modifier is used and it cannot be imported in user custom `AuthenticationConverter`.

In version 0.4.2-M2 use rcan adding an AuthenticationProvider and AuthenticationConverter, modify the access modifier to public makes it easier to reuse code.


jgrandja:
@Hccake `OAuth2EndpointUtils` is meant for internal use only.

FYI, there are a number of classes that are package-private and are only meant to be used internally by the framework. Even if we were to expose one of these utility classes, we would need to re-evaluate the contract for each of the methods to ensure the design is flexible. Simply changing the modifier to `public` is not sufficient.

I'm going to close this as `OAuth2EndpointUtils` is not designed to be publicly exposed.

PR #940, opened 2022-10-25
Update OAuth 2.1 spec link in README.adoc

octopusthu:


jgrandja:
Thanks for the update @octopusthu. This is now merged.

Issue #159, opened 2020-11-16
Add logging


jgrandja:
We need to add logging to allow for more efficient troubleshooting during error conditions.

rlewczuk:
Not sure when we'll reach MVP, lack of logs hampers authorization server in real world workloads. The same can be said about audit events, albeit to lesser extent and I'm not sure if it belongs to this ticket or new one should be created.

I can try implementing at least some if it's not too early and if (as an outsider) I'm proper person for such task (this is comprehensive, I'll probably need some initial guidance that long time spring contributors already know).

sjohnr:
@rlewczuk,

Regarding audit events, see [this comment](https://github.com/spring-projects-experimental/spring-authorization-server/pull/217#issuecomment-796893680). The decision was made not to pursue that at this time as there is already a hook to hang auditing on in Spring Security. If however you are referring to something more specific, please open an issue with the details and we can discuss that.

I think we'll need to do some design work before anything can be put in code, so probably best not to start anything yet.

ikomissarov:
What about the original request to add some logging? You don't need to do a design work for that, do you?
I was really surprised not to see a single line in logs produced by the library. Not even when I switched to TRACE level. Not even in case of some error scenario, all exceptions are just swallowed without logging.

steinwelberg:
Unfortunately the auditing events of Spring Security are not fine-grained enough for our use case. We'd like to log details about the successful or failed Authorization requests for example. Currently, in order to achieve this, we need to override the`AuthenticationSuccessHandler` and `AuthenticationSuccessHandler` used in all the filters. This is quite ugly and error prone as we'd have to manually check if the implementation in the Spring Authorization server has been changed when we are upgrading the Spring authorization server version. 

I might have an idea, how to be able to achieve this. What about giving the possibility to provide a `FunctionalInterface` that gets executed. If no interface is provided, nothing happens. We'd need to have one for both the success and failure handlers, in order to distinguish the different scenarios.

An example `FunctionalInterface` for an Authorization Error.
```
@FunctionalInterface
public interface AuthorizationErrorFunction {

  void onError(OAuth2Error error);
}
```

An example Authentication ErrorHandler (many things are removed to keep the code focused on the goal)
```
public class AuthorizationErrorResponseHandler implements AuthenticationFailureHandler {
  private final RedirectStrategy redirectStrategy = new DefaultRedirectStrategy();
[...]

PR #956, opened 2022-11-02
Add logging

sjohnr:
For consistency with Spring Security (best effort), the logs introduced in this PR use the following keywords:
* "Retrieved"
* "Validated"
* "Generated"
* "Saved"
* "Removed"
* "Revoked"
* "Authenticated"
* "Set SecurityContextHolder"
* "Did not ... since ..."

**Note:** We also use a console color highlighting plugin (GrepConsole) in IntelliJ that looks for keywords in Spring Security logs which the above could be added to. Currently we look for:
* "Invoking"
* "Authenticating"
* "Authenticated"
* "Authorizing"
* "Authorized"
* "Securing"
* "Secured"
* "Set SecurityContextHolder"

Here's an example log file with Spring Security trace enabled:

[application.log](https://github.com/spring-projects/spring-authorization-server/files/9914966/application.log)

Examples:

Screen Shot 2022-11-01 at 7 45 15 PM

Screen Shot 2022-11-01 at 7 48 54 PM

Closes gh-159

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java
@@ -17,6 +17,9 @@ package org.springframework.security.oauth2.server.authorization.authentication;
 
 import java.time.Instant;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
@@ -47,6 +50,7 @@ import org.springframework.util.Assert;
  */
 public final class ClientSecretAuthenticationProvider implements AuthenticationProvider {
 	private static final String ERROR_URI = "https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1";
+	private final Log logger = LogFactory.getLog(getClass());
 	private final RegisteredClientRepository registeredClientRepository;
 	private final CodeVerifierAuthenticator codeVerifierAuthenticator;
 	private PasswordEncoder passwordEncoder;
@@ -95,6 +99,10 @@ public final class ClientSecretAuthenticationProvider implements AuthenticationP
 			throwInvalidClient(OAuth2ParameterNames.CLIENT_ID);
 		}
 
+		if (this.logger.isTraceEnabled()) {
+			this.logger.trace("Retrieved registered client");
+		}
+
 		if (!registeredClient.getClientAuthenticationMethods().contains(
 				clientAuthentication.getClientAuthenticationMethod())) {
 			throwInvalidClient("authentication_method");
@@ -114,8 +122,16 @@ public final class ClientSecretAuthenticationProvider implements AuthenticationP
 			throwInvalidClient("client_secret_expires_at");
 		}
 
+		if (this.logger.isTraceEnabled()) {
+			this.logger.trace("Validated client authentication parameters");
+		}
+
 		// Validate the "code_verifier" parameter for the confidential client, if available
 		this.codeVerifierAuthenticator.authenticateIfAvailable(clientAuthentication, registeredClient);
+
+		if (this.logger.isTraceEnabled()) {
+			this.logger.trace("Authenticated client secret");
+		}
 
 		return new OAuth2ClientAuthenticationToken(registeredClient,
 				clientAuthentication.getClientAuthenticationMethod(), clientAuthentication.getCredentials());
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
@@ -20,6 +20,10 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.core.log.LogMessage;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
@@ -68,6 +72,7 @@ import static org.springframework.security.oauth2.server.authorization.authentic
  */
 public final class OAuth2AuthorizationCodeAuthenticationProvider implements AuthenticationProvider {
 	private static final String ERROR_URI = "https://datatracker.ietf.org/doc/html/rfc6749#section-5.2";
+	private final Log logger = LogFactory.getLog(getClass());
 	private static final OAuth2TokenType AUTHORIZATION_CODE_TOKEN_TYPE =
 			new OAuth2TokenType(OAuth2ParameterNames.CODE);
 	private static final OAuth2TokenType ID_TOKEN_TOKEN_TYPE =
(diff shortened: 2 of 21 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -166,6 +171,7 @@ private void onAuthenticationSuccess(HttpServletRequest request, HttpServletResp
 		SecurityContext securityContext = SecurityContextHolder.createEmptyContext();
 		securityContext.setAuthentication(authentication);
 		SecurityContextHolder.setContext(securityContext);
+		logger.debug(LogMessage.format("Set SecurityContextHolder to %s", securityContext));
```
jgrandja, line 174: Let's avoid passing objects that will result in `toString()` since this may log sensitive information. For example, a custom `SecurityContext` may return sensitive information in it's `toString()`. Instead, let's log the `Authentication.getClass().getSimpleName()`

jgrandja, line 174: Check `if (logger.isDebugEnabled())`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java
```diff
@@ -47,6 +50,7 @@
  */
 public final class ClientSecretAuthenticationProvider implements AuthenticationProvider {
 	private static final String ERROR_URI = "https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1";
+	private static final Log logger = LogFactory.getLog(ClientSecretAuthenticationProvider.class);
```
jgrandja, line 53: I wonder if we should change this to `private final Log logger = LogFactory.getLog(getClass())` for all `AuthenticationProvider`'s?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java
```diff
@@ -86,6 +90,7 @@ public Authentication authenticate(Authentication authentication) throws Authent
 
 		if (!ClientAuthenticationMethod.CLIENT_SECRET_BASIC.equals(clientAuthentication.getClientAuthenticationMethod()) &&
 				!ClientAuthenticationMethod.CLIENT_SECRET_POST.equals(clientAuthentication.getClientAuthenticationMethod())) {
+			logger.trace("Did not match client authentication method");
```
jgrandja, line 93: Let's remove this - the log might get noisy. Apply the same to the other providers.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
```diff
@@ -115,6 +125,7 @@ public Authentication authenticate(Authentication authentication) throws Authent
 				// Invalidate the authorization code given that a different client is attempting to use it
 				authorization = OAuth2AuthenticationProviderUtils.invalidate(authorization, authorizationCode.getToken());
 				this.authorizationService.save(authorization);
+				logger.warn(LogMessage.format("Invalidated authorization code used by %s", registeredClient.getClientId()));
```
jgrandja, line 128: Instead of logging `getClientId()` use `getId()`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java
```diff
@@ -161,6 +170,8 @@ public Authentication authenticate(Authentication authentication) throws Authent
 			Set<String> currentAuthorizedScopes = currentAuthorizationConsent != null ?
 					currentAuthorizationConsent.getScopes() : null;
 
+			logger.debug("Generated state and saved authorization");
```
jgrandja, line 173: Let's enhance this message to indicate the `state` is for authorization consent flow

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationConsentAuthenticationProvider.java
```diff
@@ -97,6 +101,8 @@ public Authentication authenticate(Authentication authentication) throws Authent
 					authorizationConsentAuthentication, null, null);
 		}
 
+		logger.trace("Retrieved authorization with state");
```
jgrandja, line 104: Let's enhance this message to indicate this is for consent flow

jgrandja (changes requested):
Thanks for the PR @sjohnr.

Please add logs to `OidcClientRegistrationAuthenticationProvider` and `OidcClientConfigurationAuthenticationProvider`.

Also, for all `Filter`'s before calling `AuthenticationManager.authenticate()`, let's add the log:

```
if (logger.isTraceEnabled()) {
	logger.trace(LogMessage.format("Authenticating: %s", authenticationRequest.getClass().getSimpleName()));
}
```

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java
sjohnr: Sounds good. I couldn't find consensus in Spring Security on it, and different projects tend to prefer different ways (`LOGGER` vs `logger`, static vs instance, etc.)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/web/OidcClientRegistrationEndpointFilter.java
```diff
@@ -202,6 +207,7 @@ private void sendClientRegistrationResponse(HttpServletRequest request, HttpServ
 	private void sendErrorResponse(HttpServletRequest request, HttpServletResponse response,
 			AuthenticationException authenticationException) throws IOException {
 		OAuth2Error error = ((OAuth2AuthenticationException) authenticationException).getError();
+		logger.trace(LogMessage.format("Client registration request failed: %s", error), authenticationException);
```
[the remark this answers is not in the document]
sjohnr: > We should also check `if (logger.isTraceEnabled())` before we `logger.trace()`.

Sounds good. I'm assuming you mean in cases where we `LogMessage.format(...)` to skip object creation when trace is not enabled, correct?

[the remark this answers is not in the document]
jgrandja: For all cases. It's common pattern to check if it's enabled before calling the log statement. Even in simple `String` messages, it's still allocating a `String` object.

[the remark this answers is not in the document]
sjohnr: Right. Anywhere we'd be computing a String, I'm using `LogMessage.format(...)` as Spring Security does, and I've wrapped those in `isXyzEnabled()` now. The only log statements that aren't wrapped are static strings, as there should not be any overhead in these cases. Shall I still add it to all other locations?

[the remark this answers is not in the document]
jgrandja: Even with static strings, the implementation behind the logger (e.g. `logger.trace()`) may have some overhead depending how its implemented in the logger. Either way, I think it's best practice to check for `isXyzEnabled()`

Issue #925, opened 2022-10-11
Simplify customizing the access token response


Hccake:
In [OAuth2.1 draft](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-06#section-3.2.3)，Token Response can be added with additional parameters.

It can currently be solved by replacing the default `AuthenticationSuccessHandler`.

Hope there is a component that can be easily implemented, like [`TokenEnhancer`](https://github.com/spring-attic/spring-security-oauth/blob/main/spring-security-oauth2/src/main/java/org/springframework/security/oauth2/provider/token/TokenEnhancer.java) in `spring-security-oauth2`.

javamachr:
There is org.springframework.security.oauth2.server.authorization.token.OAuth2TokenCustomizer example [here](https://github.com/spring-projects/spring-authorization-server/blob/51d00742f3939dbcd3b59f7c16aa9f64f1957130/docs/src/docs/asciidoc/examples/src/main/java/sample/userinfo/idtoken/IdTokenCustomizerConfig.java)

Hccake:
> There is org.springframework.security.oauth2.server.authorization.token.OAuth2TokenCustomizer example [here](https://github.com/spring-projects/spring-authorization-server/blob/51d00742f3939dbcd3b59f7c16aa9f64f1957130/docs/src/docs/asciidoc/examples/src/main/java/sample/userinfo/idtoken/IdTokenCustomizerConfig.java)

I think token endpoint response additional parameters and token claims are different things, and by default token claims are not in the response.

For example, the idToken attribute of OIDC is officially stored in additionalParameters: 
see `OAuth2AuthorizationCodeAuthenticationProvider` line 179 - 213
```java
		// ----- ID token -----
		OidcIdToken idToken;
		if (authorizationRequest.getScopes().contains(OidcScopes.OPENID)) {
			// @formatter:off
			tokenContext = tokenContextBuilder
					.tokenType(ID_TOKEN_TOKEN_TYPE)
					.authorization(authorizationBuilder.build())	// ID token customizer may need access to the access token and/or refresh token
					.build();
			// @formatter:on
			OAuth2Token generatedIdToken = this.tokenGenerator.generate(tokenContext);
			if (!(generatedIdToken instanceof Jwt)) {
				OAuth2Error error = new OAuth2Error(OAuth2ErrorCodes.SERVER_ERROR,
						"The token generator failed to generate the ID token.", ERROR_URI);
[...]

PR #961, opened 2022-11-04
Add OAuth2TokenResponseEnhancer to enhance the access token response

Hccake:
Closes gh-925


jgrandja:
@Hccake As mentioned in this [comment](https://github.com/spring-projects/spring-authorization-server/issues/925#issuecomment-1299189443):

> I feel the duplication of initializing the defaults for `OAuth2AccessTokenResponse.Builder` is quite minimal and don't think there is much value for reuse at this point.

The proposed `OAuth2TokenResponseEnhancer` is a fine-grained interface and I'm not sure how much use it will receive at this point. Although `AuthenticationSuccessHandler` is more coarse-grained, it provides greater flexibility in customizing the access token response. I still feel the code duplication for initializing the `OAuth2AccessTokenResponse.Builder` in the consuming application is acceptable. Please keep in mind that this project is a framework and code customization is expected in the consuming application. We do our best to allow for code reuse but it's also expected that the consuming application will customize on their end. If the code duplication is quite an effort then we will look at improving. But in this case, the code duplication is quite minimal and adding a new interface does not make sense at this point.

Either way, I will leave gh-925 open and will monitor the comments to see if further improvements are needed.

PR #1011, opened 2022-12-16
URL-encoded parameters in redirect URI are encoded twice

andreasf:
This fixes an issue with redirect URIs that already have URL-encoded parameters baked in.

I recently built an OIDC provider with Authorization Server for a client and noticed that some of their redirect URIs would cause problems.

PHP's Yii framework creates redirect URIs in the format `https://example.com/index.php?r=site%2Fauth&authclient=foo`. The authorization endpoint would reencode the `%2F` in the parameter to `%252F` and redirect users to a non-existing route.

This PR fixes the behavior while making sure that the encoding of the state parameter is unchanged.

Related to #875

### DIFF
--- a/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/TestRegisteredClients.java
+++ b/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/TestRegisteredClients.java
@@ -39,6 +39,18 @@ public class TestRegisteredClients {
 				.scope("scope1");
 	}
 
+	public static RegisteredClient.Builder registeredClientEncodedUri() {
+		return RegisteredClient.withId("registration-1")
+				.clientId("client-1")
+				.clientIdIssuedAt(Instant.now().truncatedTo(ChronoUnit.SECONDS))
+				.clientSecret("secret-1")
+				.authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)
+				.authorizationGrantType(AuthorizationGrantType.REFRESH_TOKEN)
+				.clientAuthenticationMethod(ClientAuthenticationMethod.CLIENT_SECRET_BASIC)
+				.redirectUri("https://example.com?param=is%2Fencoded")
+				.scope("scope1");
+	}
+
 	public static RegisteredClient.Builder registeredClient2() {
 		return RegisteredClient.withId("registration-2")
 				.clientId("client-2")
--- a/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilterTests.java
+++ b/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilterTests.java
@@ -83,7 +83,7 @@ import static org.mockito.Mockito.when;
 public class OAuth2AuthorizationEndpointFilterTests {
 	private static final String DEFAULT_AUTHORIZATION_ENDPOINT_URI = "/oauth2/authorize";
 	private static final String AUTHORIZATION_URI = "https://provider.com/oauth2/authorize";
-	private static final String STATE = "state";
+	private static final String STATE = "previously encoded/state";
 	private static final String REMOTE_ADDRESS = "remote-address";
 	private AuthenticationManager authenticationManager;
 	private OAuth2AuthorizationEndpointFilter filter;
@@ -299,7 +299,7 @@ public class OAuth2AuthorizationEndpointFilterTests {
 		verifyNoInteractions(filterChain);
 
 		assertThat(response.getStatus()).isEqualTo(HttpStatus.FOUND.value());
-		assertThat(response.getRedirectedUrl()).isEqualTo("https://example.com?error=errorCode&error_description=errorDescription&error_uri=errorUri&state=state");
+		assertThat(response.getRedirectedUrl()).isEqualTo("https://example.com?error=errorCode&error_description=errorDescription&error_uri=errorUri&state=previously%20encoded%2Fstate");
 		assertThat(SecurityContextHolder.getContext().getAuthentication()).isSameAs(this.principal);
 	}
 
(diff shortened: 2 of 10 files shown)


oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/client/TestRegisteredClients.java
```diff
@@ -39,6 +39,18 @@ public static RegisteredClient.Builder registeredClient() {
 				.scope("scope1");
 	}
 
+	public static RegisteredClient.Builder registeredClientEncodedUri() {
```
jgrandja, line 42: Remove this and leverage the existing `TestRegisteredClients.registeredClient()` and add the url-encoded redirect-uri

samples/custom-consent-authorizationserver/samples-custom-consent-authorizationserver.gradle
```diff
@@ -8,6 +8,10 @@ group = project.rootProject.group
 version = project.rootProject.version
 sourceCompatibility = "17"
 
+test {
```
jgrandja, line 11: Please revert all changes in `custom-consent-authorizationserver` since we have test coverage in `default-authorizationserver` and there is no need to duplicate

samples/default-authorizationserver/samples-default-authorizationserver.gradle
```diff
@@ -8,6 +8,10 @@ group = project.rootProject.group
 version = project.rootProject.version
 sourceCompatibility = "17"
 
+test {
+	useJUnitPlatform()
```
jgrandja, line 12: This can be removed since it's automatically applied by the `SpringJavaPlugin`

samples/default-authorizationserver/src/main/java/sample/config/AuthorizationServerConfig.java
```diff
@@ -87,6 +87,7 @@ public RegisteredClientRepository registeredClientRepository(JdbcTemplate jdbcTe
 				.authorizationGrantType(AuthorizationGrantType.REFRESH_TOKEN)
 				.authorizationGrantType(AuthorizationGrantType.CLIENT_CREDENTIALS)
 				.redirectUri("http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc")
+				.redirectUri("http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc?param=is%2Fencoded")
```
jgrandja, line 90: Let's change `is%2Fencoded` to `encoded%20parameter%20value`

samples/default-authorizationserver/src/test/java/sample/DefaultAuthorizationServerApplicationTests.java
```diff
@@ -46,15 +50,18 @@
...
+	public static final String REENCODED_STATE = "some%20%20encoded%2Fstate";
 
 	private static final String AUTHORIZATION_REQUEST = UriComponentsBuilder
 			.fromPath("/oauth2/authorize")
 			.queryParam("response_type", "code")
 			.queryParam("client_id", "messaging-client")
 			.queryParam("scope", "openid")
-			.queryParam("state", "some-state")
-			.queryParam("redirect_uri", REDIRECT_URI)
+			.queryParam("state", STATE)
```
jgrandja, line 62: Please remove all changes related to the `state` parameter since this was fixed in gh-875 and test coverage is included in the associated commit

samples/default-authorizationserver/src/test/java/sample/DefaultAuthorizationServerApplicationTests.java
```diff
@@ -67,6 +74,16 @@ public void setUp() {
 		this.webClient.getCookieManager().clearCookies();	// log out
 	}
 
+	@Test
+	void authorizationRequestProperlyEncoded() {
```
jgrandja, line 78: Please remove this test since it directly tests `UriComponentsBuilder`, which is not owned by this project.

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilterTests.java
```diff
@@ -83,7 +83,7 @@
 public class OAuth2AuthorizationEndpointFilterTests {
 	private static final String DEFAULT_AUTHORIZATION_ENDPOINT_URI = "/oauth2/authorize";
 	private static final String AUTHORIZATION_URI = "https://provider.com/oauth2/authorize";
-	private static final String STATE = "state";
+	private static final String STATE = "previously encoded/state";
```
jgrandja, line 86: Let's change "previously encoded/state" to "encoded state value"

samples/default-authorizationserver/src/test/java/sample/DefaultAuthorizationServerConsentTests.java
```diff
@@ -58,15 +62,18 @@ public class DefaultAuthorizationServerConsentTests {
 	@MockBean
 	private OAuth2AuthorizationConsentService authorizationConsentService;
 
-	private final String redirectUri = "http://127.0.0.1/login/oauth2/code/messaging-client-oidc";
+	private final String redirectUri = "http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc?param=is%2Fencoded";
```
jgrandja, line 65: Please revert all changes in `DefaultAuthorizationServerConsentTests` since we have same test coverage in `DefaultAuthorizationServerApplicationTests`

jgrandja (changes requested):
Thanks for the PR @andreasf ! Please see review comments.

samples/default-authorizationserver/samples-default-authorizationserver.gradle
andreasf: I spent quite a while figuring out why the tests don't run locally and in CI, in my opinion this config should have been added in the "Upgrade to JUnit 5" commit.

Did you try it out and are tests running (and failing the build) on your machine without explicitly enabling JUnit 5?

jgrandja: I just realized that the samples do not use the `io.spring.convention.spring-module` plugin, so `SpringJavaPlugin` is not applied as I previously mentioned. Sorry for the confusion there.

However, I'm wondering why the tests are failing on your end? The tests in `default-authorizationserver` sample are passing locally and in CI for quite some time now.

jgrandja:
@andreasf Just checking back to see if you're able to apply the requested changes?

samples/default-authorizationserver/samples-default-authorizationserver.gradle
andreasf: Sorry, I didn't make myself fully clear. The tests are not failing in their current state.

What I meant was: if Gradle says the build was successful, that does not necessarily mean that it ran tests. To check if it does, I like to make a test fail, e.g. by throwing a RuntimeException. The failing test should also lead to a failing build.

For CI, I checked the logs of several commits for the test output – I think there's a DB-related message. This message did not appear after the upgrade to JUnit 5.

jgrandja:
@andreasf We have a release next week and I wanted to get this fix in so I went ahead and added a polish commit and merged.

Issue #1044, opened 2023-01-16
How-to: Customize client metadata during dynamic client registration


jgrandja:
We should provide a guide that demonstrates how to customize the client metadata in `OidcClientRegistration` when it's provided to the [OpenID Connect 1.0 Dynamic Client Registration Endpoint](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/protocol-endpoints.html#oidc-client-registration-endpoint) `OidcClientRegistrationAuthenticationProvider`.

Related gh-647

jgrandja:
The extension point for customizing client metadata before it's saved to `RegisteredClientRepository` is `OidcClientRegistrationAuthenticationProvider.setRegisteredClientConverter()`.

Here is a sample configuration:

```java
@Bean
public SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http) throws Exception {
	OAuth2AuthorizationServerConfigurer authorizationServerConfigurer =
			new OAuth2AuthorizationServerConfigurer();
	authorizationServerConfigurer
		.oidc(oidc ->
			oidc
				.clientRegistrationEndpoint(clientRegistration ->
					clientRegistration
						.authenticationProviders(configureRegisteredClientConverter())
				)
			);

	...

}

private Consumer<List<AuthenticationProvider>> configureRegisteredClientConverter() {
	return (authenticationProviders) ->
		authenticationProviders.forEach(authenticationProvider -> {
			if (authenticationProvider instanceof OidcClientRegistrationAuthenticationProvider) {
				OidcClientRegistrationAuthenticationProvider clientRegistrationAuthenticationProvider =
						(OidcClientRegistrationAuthenticationProvider) authenticationProvider;
				clientRegistrationAuthenticationProvider.setRegisteredClientConverter(
						new CustomRegisteredClientConverter());
			}
		});
}

private static final class CustomRegisteredClientConverter implements Converter<OidcClientRegistration, RegisteredClient> {

	@Override
	public RegisteredClient convert(OidcClientRegistration clientRegistration) {
		// TODO Implement conversion from OidcClientRegistration to RegisteredClient
	}

}
```

ddubson:
@jgrandja can I be assigned this issue?

Should the contents of this how-to be appended to the existing "Register a client dynamically" guide? or should it be a follow-up standalone how-to guide?

PR #1056, opened 2023-01-31
Fix to save after encoding the secret when registering the client

uc4w6c:
In the Client Registration Endpoint the authentication method is client_secret_basic or client_secret_post or client_secret_jwt When , it should encode the secret and save it.
Since it is currently stored as plain text, the client cannot issue tokens.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OidcClientRegistrationEndpointConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OidcClientRegistrationEndpointConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 the original author or authors.
+ * Copyright 2020-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.config.annotation.ObjectPostProcessor;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
+import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.server.authorization.oidc.OidcClientRegistration;
@@ -221,6 +222,10 @@ public final class OidcClientRegistrationEndpointConfigurer extends AbstractOAut
 						OAuth2ConfigurerUtils.getRegisteredClientRepository(httpSecurity),
 						OAuth2ConfigurerUtils.getAuthorizationService(httpSecurity),
 						OAuth2ConfigurerUtils.getTokenGenerator(httpSecurity));
+		PasswordEncoder passwordEncoder = OAuth2ConfigurerUtils.getOptionalBean(httpSecurity, PasswordEncoder.class);
+		if (passwordEncoder != null) {
+			oidcClientRegistrationAuthenticationProvider.setPasswordEncoder(passwordEncoder);
+		}
 		authenticationProviders.add(oidcClientRegistrationAuthenticationProvider);
 
 		OidcClientConfigurationAuthenticationProvider oidcClientConfigurationAuthenticationProvider =
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 the original author or authors.
+ * Copyright 2020-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -34,8 +34,10 @@ import org.springframework.core.convert.converter.Converter;
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
+import org.springframework.security.crypto.factory.PasswordEncoderFactories;
 import org.springframework.security.crypto.keygen.Base64StringKeyGenerator;
 import org.springframework.security.crypto.keygen.StringKeyGenerator;
+import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.ClaimAccessor;
 import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
@@ -79,6 +81,7 @@ import org.springframework.util.StringUtils;
  * @see OAuth2TokenGenerator
  * @see OidcClientRegistrationAuthenticationToken
  * @see OidcClientConfigurationAuthenticationProvider
+ * @see PasswordEncoder
  * @see <a href="https://openid.net/specs/openid-connect-registration-1_0.html#ClientRegistration">3. Client Registration Endpoint</a>
  */
 public final class OidcClientRegistrationAuthenticationProvider implements AuthenticationProvider {
@@ -90,6 +93,8 @@ public final class OidcClientRegistrationAuthenticationProvider implements Authe
 	private final OAuth2TokenGenerator<? extends OAuth2Token> tokenGenerator;
 	private final Converter<RegisteredClient, OidcClientRegistration> clientRegistrationConverter;
 	private Converter<OidcClientRegistration, RegisteredClient> registeredClientConverter;
+
+	private PasswordEncoder passwordEncoder;
 
 	/**
 	 * Constructs an {@code OidcClientRegistrationAuthenticationProvider} using the provided parameters.
@@ -109,6 +114,21 @@ public final class OidcClientRegistrationAuthenticationProvider implements Authe
 		this.tokenGenerator = tokenGenerator;
 		this.clientRegistrationConverter = new RegisteredClientOidcClientRegistrationConverter();
 		this.registeredClientConverter = new OidcClientRegistrationRegisteredClientConverter();
+		this.passwordEncoder = PasswordEncoderFactories.createDelegatingPasswordEncoder();
+	}
+
+	/**
+	 * Sets the {@link PasswordEncoder} used to encode the clientSecret
+	 * the {@link RegisteredClient#getClientSecret() client secret}.
+	 * If not set, the client secret will be encoded using
+	 * {@link PasswordEncoderFactories#createDelegatingPasswordEncoder()}.
+	 *
+	 * @param passwordEncoder the {@link PasswordEncoder} used to encode the clientSecret
+	 * @since 1.1.0
+	 */
+	public void setPasswordEncoder(PasswordEncoder passwordEncoder) {
+		Assert.notNull(passwordEncoder, "passwordEncoder cannot be null");
+		this.passwordEncoder = passwordEncoder;
 	}
 
 	@Override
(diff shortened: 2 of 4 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
```diff
@@ -183,7 +195,26 @@ private OidcClientRegistrationAuthenticationToken registerClient(OidcClientRegis
 		}
 
 		RegisteredClient registeredClient = this.registeredClientConverter.convert(clientRegistrationAuthentication.getClientRegistration());
-		this.registeredClientRepository.save(registeredClient);
+
+		// Copy RegisteredClient and encode only secret
+		String rawClientSecret = registeredClient.getClientSecret();
```
jgrandja, line 200: The extension point for customizing client metadata before it's saved to `RegisteredClientRepository` is `OidcClientRegistrationAuthenticationProvider.setRegisteredClientConverter()`. It's not currently documented in the reference manual but there is an open issue gh-1044 that provides sample code on how to configure.

jgrandja:
@uc4w6c Not all `RegisteredClient`'s will be assigned a client secret during dynamic registration. The extension point `OidcClientRegistrationAuthenticationProvider.setRegisteredClientConverter()` gives flexibility for consuming applications to assign a client secret (if necessary) and encode using the strategy of their choice. But it's not limited to just the client secret since any client metadata can be customized at that point. Adding `setPasswordEncoder()` would be redundant since this can be customized using `setRegisteredClientConverter()`.

I'm going to close this PR as the preference is to leverage the extension point `setRegisteredClientConverter()` for any client metadata customization.

uc4w6c:
I don't want to customize the client metadata before saving it to the RegisteredClientRepository.

I would like to solve the bug that the newly created client cannot be authenticated because the secret is not encoded at the time of client registration, but the secret is decoded at the time of client authentication.

I have created a simple example below.
https://github.com/uc4w6c/spring-authorization-server-pr-1056

`OidcClientRegistrationAuthenticationProvider.java` does not set the secret encoded with PasswordEncoder.
However, `ClientSecretAuthenticationProvider.java` decodes the secret with PasswordEncoder after getting the clientSecret from registeredClientRepository and compares it with the secret sent by the user.
https://github.com/spring-projects/spring-authorization-server/blob/67305abc54c2e7095b9176cb57840749a7de96e5/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java#L116

Also, if I were to encode the secret with the RegisteredClientConverter, the encoded secret would be returned as json in the response, which is not a desirable behavior.
So I think should encode the secret only in the argument to RegisteredClientRepository.

jgrandja:
@uc4w6c 

> Also, if I were to encode the secret with the RegisteredClientConverter, the encoded secret would be returned as json in the response, which is not a desirable behavior.

You could customize the response by setting a custom `OidcClientRegistrationEndpointFilter.setAuthenticationSuccessHandler()` that decodes the encoded clientSecret. I realize this might not be ideal but I'm not keen on exposing `OidcClientRegistrationAuthenticationProvider. setPasswordEncoder()`.

Exposing `OidcClientRegistrationAuthenticationProvider.setClientRegistrationConverter()` might be the option to consider.

I'll re-open this to look into this further.

uc4w6c:
@jgrandja 
Thanks for reopening.
Sure.I would be happy to have `OidcClientRegistrationAuthenticationProvider.setClientRegistrationConverter()`.
But I think the problem this time is a bug in the default configuration of Spring Authorization Server.

My explanation so far may have been difficult to understand, so I will describe the phenomenon again.

---

1. Below is the OIDC configuration.
Default settings except add a clientRegistrationEndpoint.

```
    OAuth2AuthorizationServerConfiguration.applyDefaultSecurity(http);
      .oidc(oidc -> oidc.clientRegistrationEndpoint(Customizer.withDefaults()));
```

2. Register a new client with the `/connect/register` endpoint.
The parameters look like this.

```
'{ "redirect_uris": "http://127.0.0.1/callback", "grant_types": ["authorization_code", "client_credentials"] }' 
```

```
{
	"client_id": "UuUnmeUMcXzQ7BMYKyIYBsbvuFaP4rbHUYUVSudClqk",
	"client_secret": "o-TpPEQ4UWOLyAcnWqX0vec2S7_A2FjJpL3fVLBAAQJP3WRVAlIB53UmpGD9ydyH",
        // Omit other items
}
```

3. Issue a token with the newly created client with the `/oauth2/token` endpoint.

```
curl -X POST "http://localhost:8080/oauth2/token" -H 'Content-Type: application/x-www-form-urlencoded' -d "grant_type=client_credentials"  -u 'UuUnmeUMcXzQ7BMYKyIYBsbvuFaP4rbHUYUVSudClqk:o-TpPEQ4UWOLyAcnWqX0vec2S7_A2FjJpL3fVLBAAQJP3WRVAlIB53UmpGD9ydyH'
```

An error occurs.
```
java.lang.IllegalArgumentException: There is no PasswordEncoder mapped for the id "null"
	at org.springframework.security.crypto.password.DelegatingPasswordEncoder$UnmappedIdPasswordEncoder.matches(DelegatingPasswordEncoder.java:289) ~[spring-security-crypto-6.0.1.jar:6.0.1]
	at org.springframework.security.crypto.password.DelegatingPasswordEncoder.matches(DelegatingPasswordEncoder.java:237) ~[spring-security-crypto-6.0.1.jar:6.0.1]
	at org.springframework.security.oauth2.server.authorization.authentication.ClientSecretAuthenticationProvider.authenticate(ClientSecretAuthenticationProvider.java:116) ~[spring-security-oauth2-authorization-server-1.0.0.jar:1.0.0]
	at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:182) ~[spring-security-core-6.0.1.jar:6.0.1]
	at org.springframework.security.oauth2.server.authorization.web.OAuth2ClientAuthenticationFilter.doFilterInternal(OAuth2ClientAuthenticationFilter.java:122) ~[spring-security-oauth2-authorization-server-1.0.0.jar:1.0.0]
  // Omit
```

---

The reason for the error is that the `/connect/register` endpoint does not encode the secret, but the `/oauth2/token` endpoint decodes. with default settings.
Why you don't want to add `OidcClientRegistrationAuthenticationProvider.setPasswordEncoder()` when We already have `ClientSecretAuthenticationProvider.setPasswordEncoder()`?

Pardon my broken English.

uc4w6c:
@jgrandja 
I really want to solve this problem.
Is there anything I can do?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OidcClientRegistrationEndpointConfigurer.java
```diff
@@ -26,6 +26,7 @@
 import org.springframework.security.authentication.AuthenticationProvider;
 import org.springframework.security.config.annotation.ObjectPostProcessor;
```
jgrandja, line 27: Please update copyright year to 2023

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
```diff
@@ -18,14 +18,7 @@
...
 import java.time.Instant;
-import java.util.Base64;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
jgrandja, line 21: Please revert as wildcard imports are not allowed.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
```diff
@@ -109,6 +107,20 @@ public OidcClientRegistrationAuthenticationProvider(RegisteredClientRepository r
 		this.tokenGenerator = tokenGenerator;
 		this.clientRegistrationConverter = new RegisteredClientOidcClientRegistrationConverter();
 		this.registeredClientConverter = new OidcClientRegistrationRegisteredClientConverter();
+		this.passwordEncoder = PasswordEncoderFactories.createDelegatingPasswordEncoder();
+	}
+
+	/**
+	 * Sets the {@link PasswordEncoder} used to validate
+	 * the {@link RegisteredClient#getClientSecret() client secret}.
```
jgrandja, line 115: Please update the javadoc as the passwordencoder is not used to validate the clientSecret but instead used to encode the clientSecret.
Also, please add `@since 1.1.0`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
```diff
@@ -183,7 +195,26 @@ private OidcClientRegistrationAuthenticationToken registerClient(OidcClientRegis
...
 		RegisteredClient registeredClient = this.registeredClientConverter.convert(clientRegistrationAuthentication.getClientRegistration());
-		this.registeredClientRepository.save(registeredClient);
+
+		// Copy RegisteredClient and encode only secret
+		String rawClientSecret = registeredClient.getClientSecret();
+		String clientSecret = null;
+		if (rawClientSecret != null) {
+			clientSecret = passwordEncoder.encode(rawClientSecret);
+		}
+		RegisteredClient saveRegisteredClient = RegisteredClient.withId(registeredClient.getId())
```
jgrandja, line 205: Use `RegisteredClient.from(registeredClient)` to make a copy and then update `clientSecret`. Also, the `RegisteredClient` should only be copied/updated if the `clientSecret` is set, otherwise it should use the original returned from the converter.

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProviderTests.java
```diff
@@ -67,8 +69,7 @@
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatIllegalArgumentException;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
-import static org.mockito.ArgumentMatchers.any;
-import static org.mockito.ArgumentMatchers.eq;
+import static org.mockito.ArgumentMatchers.*;
```
jgrandja, line 72: Revert as wildcard imports are not allowed.

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProviderTests.java
```diff
@@ -627,6 +650,60 @@ public void authenticateWhenValidAccessTokenThenReturnClientRegistration() {
 		assertThat(clientRegistrationResult.getRegistrationAccessToken()).isEqualTo(jwt.getTokenValue());
 	}
 
+	@Test
+	public void authenticateWhenSettingPasswordEncoder() {
```
jgrandja, line 654: Please remove this test and add an integration test to `OidcClientRegistrationTests`.

Here's a template of the test:

```java
// gh-1056
@Test
public void requestWhenClientRegistersWithSecretThenClientAuthenticationSuccess() throws Exception {
	// TODO Update the PasswordEncoder @Bean to PasswordEncoderFactories.createDelegatingPasswordEncoder()
	this.spring.register(AuthorizationServerConfiguration.class).autowire();

	// TODO
	// Create new client with client_credentials grant
	OidcClientRegistration clientRegistration = OidcClientRegistration.builder() ...

	// Register client
	OidcClientRegistration clientRegistrationResponse = registerClient(clientRegistration);

	// TODO
	// Use the newly registered client to obtain an access token using the client_credentials grant
	// if the access token is obtained then client authentication succeeded

}

```

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OidcClientRegistrationTests.java
```diff
@@ -290,7 +291,6 @@ public void requestWhenClientConfigurationRequestAuthorizedThenClientRegistratio
 
 		assertThat(clientConfigurationResponse.getClientId()).isEqualTo(clientRegistrationResponse.getClientId());
 		assertThat(clientConfigurationResponse.getClientIdIssuedAt()).isEqualTo(clientRegistrationResponse.getClientIdIssuedAt());
-		assertThat(clientConfigurationResponse.getClientSecret()).isEqualTo(clientRegistrationResponse.getClientSecret());
```
uc4w6c, line 293: `ClientConfigurationResponse.getClientSecret()` returns plain Secret, but `ClientRegitionrationRespone. getClientSecret()` returns the encoded secret, so do not assert.

uc4w6c:
@jgrandja 
Thank you.
Since I have corrected it, please review it again.

jgrandja:
Thanks for the updates @uc4w6c. FYI, I added a polish commit with some minor updates.

Also, the [setPasswordEncoder()](https://github.com/spring-projects/spring-authorization-server/commit/addd6e13d57eb1567f951c9ac1102998faac7b90#diff-505cdbfa98f8ccdec7bbc429848b30d663e3e62e163034f4c38660e24e46ccd3R178) method could not be exposed in the patch releases `0.4.2` and `1.0.2`, however, it is available in [main](https://github.com/spring-projects/spring-authorization-server/commit/d197c188ba75c71aab546316ac9e1ec37f4b6fbc#diff-505cdbfa98f8ccdec7bbc429848b30d663e3e62e163034f4c38660e24e46ccd3R182) targeted for 1.1.0 release.

uc4w6c:
@jgrandja 
Thanks for the merge and update.
Looking forward to the release.

PR #1078, opened 2023-02-19
fix: error description not url-encoded

Edsuns:
`error.getDescription()` is not encoded here: 

https://github.com/spring-projects/spring-authorization-server/blob/98e3fe807a56da3236bc92b4c07f61e61a80efbc/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java#L353

That will cause `IllegalArgumentException` below: 

```
java.lang.IllegalArgumentException: Invalid character ' ' for QUERY_PARAM in "OAuth 2.0 Parameter: client_id"
	at org.springframework.web.util.HierarchicalUriComponents.verifyUriComponent(HierarchicalUriComponents.java:421) ~[spring-web-6.0.2.jar:6.0.2]
	at org.springframework.web.util.HierarchicalUriComponents.lambda$verify$4(HierarchicalUriComponents.java:389) ~[spring-web-6.0.2.jar:6.0.2]
	at org.springframework.util.UnmodifiableMultiValueMap.lambda$forEach$0(UnmodifiableMultiValueMap.java:114) ~[spring-core-6.0.2.jar:6.0.2]
	at java.base/java.util.Map.forEach(Map.java:713) ~[na:na]
```

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
@@ -345,7 +345,8 @@ public final class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilte
 					UriUtils.encode(error.getDescription(), StandardCharsets.UTF_8));
 		}
 		if (StringUtils.hasText(error.getUri())) {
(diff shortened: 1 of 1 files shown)


jgrandja:
Thank you for reporting this @Edsuns. I completely missed the encoding of the `OAuth2Error` while applying polish to gh-1011. Luckily, I was able to apply the updates via 0cd594082e0beb5346eb0e2c94e8e82c71c11f1e just before the releases today. Thanks again for the quick feedback! I will close this as a duplicate.

PR #1105, opened 2023-02-27
Upgrade client secret when appropriately available

### LINKED ISSUE #1099: Enable `upgradeEncoding` for OAuth2 client secrets
Presently, there isn't a built in mechanism to upgrade the encoding on an OAuth2 Client outside of simply outright changing the credential. I think it would be a tremendous addition to allow the `PasswordEncoder.upgradleEncoding(String)` method to be called upon successfully authenticating the OAuth2 client. This would then enable usage of a `DelegatingPasswordEncoder` to be able to transition the encoding of secrets from one encoder to another.

Initially looking, this seems like it would happen here:
https://github.com/spring-projects/spring-authorization-server/blob/eae6630599c139ddab3a1e9060a124fd47224059/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java#L116

Maybe something like:
```
if (!this.passwordEncoder.matches(clientSecret, registeredClient.getClientSecret())) {
	throwInvalidClient(OAuth2ParameterNames.CLIENT_SECRET);
} else {
	RegisteredClient updated = RegisteredClient.from(registeredClient)
			.secret(this.passwordEncoder.upgradeEncoding(clientSecret))
			.build();
	this.registeredClientRepository.save(updated);
}
```

shanman190:
Closes #1099.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProvider.java
@@ -122,6 +122,13 @@ public final class ClientSecretAuthenticationProvider implements AuthenticationP
 			throwInvalidClient("client_secret_expires_at");
 		}
 
+		if (this.passwordEncoder.upgradeEncoding(registeredClient.getClientSecret())) {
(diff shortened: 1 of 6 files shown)


jgrandja:
Thanks for the PR @shanman190 ! This is now merged.

FYI, I added a polish commit with some very minor updates.

PR #1134, opened 2023-03-23
Add SecurityContextRepository to OAuth2ClientAuthenticationFilter

viablecell:
Add SecurityContextHolderStrategy to OAuth2ClientAuthenticationFilter
Add SecurityContextRepository to OAuth2ClientAuthenticationFilter

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
@@ -34,6 +34,7 @@ import org.springframework.security.core.Authentication;
 import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.context.SecurityContext;
 import org.springframework.security.core.context.SecurityContextHolder;
+import org.springframework.security.core.context.SecurityContextHolderStrategy;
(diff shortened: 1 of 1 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.java
```diff
@@ -82,6 +85,9 @@ public final class OAuth2ClientAuthenticationFilter extends OncePerRequestFilter
 	private AuthenticationSuccessHandler authenticationSuccessHandler = this::onAuthenticationSuccess;
 	private AuthenticationFailureHandler authenticationFailureHandler = this::onAuthenticationFailure;
 
+	private SecurityContextHolderStrategy securityContextHolderStrategy = SecurityContextHolder
+			.getContextHolderStrategy();
+	private SecurityContextRepository securityContextRepository = new RequestAttributeSecurityContextRepository();
```
jgrandja, line 90: `RequestAttributeSecurityContextRepository` does not have to be explicitly uses since `OAuth2ClientAuthenticationToken` is marked with `@Transient` and therefore will never be persisted regardless of the backing `SecurityContextRepository`

jgrandja:
@viablecell Thanks for the PR. The current implementation works as expected so no changes are necessary at this point.

PR #1139, opened 2023-03-30
Fix refresh token error code INVALID_CLIENT to INVALID_GRANT

HuiYeong:
Hello.

We are trying to authenticate OIDC Basic OP Profile using your Spring-Authorization-Server library.

We encountered an issue where if another client requests token reissue.
It should return an 'invalid_grant' error, but instead, it returns an 'invalid_client' error.
(test fail log : https://www.certification.openid.net/log-detail.html?log=CrPyNh6RPaLJd05&public=true)

We looked through the official documentation and found that the 'invalid_grant' error should be returned when another client requests token reissue. Please refer to this link. -> https://www.rfc-editor.org/rfc/rfc6749#section-5.2

I would appreciate it if you could review it quickly and include it in the next version. Thank you!

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2RefreshTokenAuthenticationProvider.java
@@ -111,7 +111,7 @@ public final class OAuth2RefreshTokenAuthenticationProvider implements Authentic
 		}
 
 		if (!registeredClient.getId().equals(authorization.getRegisteredClientId())) {
-			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_CLIENT);
+			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
 		}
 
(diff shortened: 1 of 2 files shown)


neochae1:
I think this pull request is a good step towards conforming to the specifications.

Spec
invalid_grant
    The provided authorization grant (e.g., authorization
    code, resource owner credentials) or refresh token is
    invalid, expired, revoked, does not match the redirection
    URI used in the authorization request, or was issued to
    another client

When exchanging tokens with the OAuth2AuthorizationCodeAuthenticationProvider, 
the returned error for this scenario is 'INVALID_GRANT'. However, when using the OAuth2RefreshTokenAuthenticationProvider, the returned error is 'INVALID_CLIENT

OAuth2AuthorizationCodeAuthenticationProvider
    if (!registeredClient.getClientId().equals(authorizationRequest.getClientId())) {
        if (!authorizationCode.isInvalidated()) {
            // Invalidate the authorization code given that a different client is attempting to use it
            authorization = OAuth2AuthenticationProviderUtils.invalidate(authorization, authorizationCode.getToken());
            this.authorizationService.save(authorization);
            if (this.logger.isWarnEnabled()) {
                this.logger.warn(LogMessage.format("Invalidated authorization code used by registered client '%s'", registeredClient.getId()));
            }
        }
        throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
    }

jgrandja:
Thanks for the fix @HuiYeong. I will merge this shortly.

PR #1141, opened 2023-03-31
fix: limit PKCE code verifier to authorization code grant_type

backjo:
**Context**
Currently, the CodeVerifierAuthenticator checks the grant_type in the authenticate method, and returns false if the grant_type is not authorization_code. While this is fine for the authenticateIfAvailable method, this causes what I believe is incorrect behavior in authenticateRequired. My understanding is that the following is the desired the behavior of `authenticateRequired`:
```
If the grant_type is authorization_code, PKCE must be used and valid. If not, throw an exception. 
``` 

Currently, the behavior is as follows:
```
PKCE must be used and valid. If not, throw an exception. 
```

Since PKCE is only supported in the authorization_code flow, I believe the behavior defined above as desired behavior is what we want to happen. 

**Use case driving this change**
While it is not supported today by spring-authorization-server, we are starting to explore implementing functionality around issuing refresh tokens to a public client (a native iOS app, in our case). As we attempted to implement custom functionality to PoC supporting this, we realized that this current implementation is attempting to enforce PKCE for public clients on the refresh_token grant type - throwing an exception because authenticate returns false because the grant_type is not auth code. This PR is not intended to get down the rabbit hole of full implementation for public clients of access to refresh tokens. It's simply intended to restrict the PKCE-required validation for public clients to the grant_type it is intended for - authorization_code.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
@@ -60,6 +60,9 @@ final class CodeVerifierAuthenticator {
 
 	void authenticateRequired(OAuth2ClientAuthenticationToken clientAuthentication,
 			RegisteredClient registeredClient) {
+		if (!authorizationCodeGrant(clientAuthentication.getAdditionalParameters())) {
+			return;
+		}
 		if (!authenticate(clientAuthentication, registeredClient)) {
 			throwInvalidGrant(PkceParameterNames.CODE_VERIFIER);
 		}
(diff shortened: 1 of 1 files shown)


jgrandja:
@backjo 

> this causes what I believe is incorrect behavior in `authenticateRequired()`

`PublicClientAuthenticationProvider` calls `CodeVerifierAuthenticator.authenticateRequired()` to enforce the requirement of a valid `code_verifier`. This is the intended behaviour.

`PublicClientAuthenticationProvider` sole intent is to authenticate public clients using the `authorization_code` grant type with PKCE. **It is not designed or intended to authenticate public clients using other grant types**, e.g. `refresh_token`. Furthermore, the change proposed in this PR "opens things up" for public clients and other grant types, which is insecure by default.

> we are starting to explore implementing functionality around issuing refresh tokens to a public client (a native iOS app, in our case)

`authorizationServerConfigurer.clientAuthentication()` provides the ability to customize client authentication, so you can add your own implementation for authenticating public clients for `refresh_token` or any `grant_type`. As a starting point, take a look at `OAuth2ClientAuthenticationConfigurer`, which configures `OAuth2ClientAuthenticationFilter` with `PublicClientAuthenticationConverter` and adds `PublicClientAuthenticationProvider` to authenticate the `OAuth2ClientAuthenticationToken` returned by `PublicClientAuthenticationConverter`.

See the [reference documentation](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/configuration-model.html#configuring-client-authentication) for all available configuration options for `authorizationServerConfigurer.clientAuthentication()`.

Issue #1111, opened 2023-03-03
Production Logging


dciarniello:
Production level logging is required to be able to easily troubleshoot production issues.

Logging was added in gh-159 but at the TRACE level.  TRACE is fine for development but not for production.

Please add logging more suitable for troubleshooting production issue.

sjohnr:
@dciarniello, thanks for reaching out!

> Logging was added in https://github.com/spring-projects/spring-authorization-server/issues/159 but at the TRACE level. TRACE is fine for development but not for production.

The goal of gh-159 was to add enough logging to allow efficient troubleshooting during the development cycle without adding too much detail. Adding more details and more logging requires specific use cases and will not likely be addressed as a general request.

> Please add logging more suitable for troubleshooting production issue.

Please provide some additional information about your request, and be specific if possible. Depending on the number of use cases you can provide, we could potentially open an issue for each use case or an issue for one or more themes (groups of related use cases) that emerge from the request.

sjohnr:
Thanks @dciarniello, that's a pretty good list to start with.

One thing we're thinking about right now (that is a bit of a detour from logging) is customization around the processing flow itself. I feel like if we add the right types of hooks into the processing flow of e.g. the `AuthenticationProvider`s or `Filter`s (such as in gh-1075 and a few other suggestions), there's potential to solve two problems at once. Meaning, even if the core framework didn't have direct support for every single logging case, it would ideally be trivial to add your own logging in a specific place without much work. This would help unblock some of those needs from being bound to a specific future release, if you need it sooner. So I see two things here:

1. Add hooks into the processing flow, specifically for failure cases.
2. Add additional logging in the framework itself.

Depending on where either of these fits in the priority list, one could be addressed earlier than the other. I also think that the second request (this issue) could be addressed as contributions from the community since we have some logging in place already to show the general (simple) pattern. So if you're interested in sending any PRs, let me know!

For now, I'll do some checking, and see about adding an issue for each of the logging use cases you point out above so they are small bite-sized issues, then I'll close this issue as a duplicate. Sound good? Let me know if you have any other thoughts.

sjohnr:
Closing as a duplicate of gh-1245, gh-1246, gh-1247, and gh-1248.

dciarniello:
Hi, @sjohnr.

Thanks for looking into this and creating the individual tickets.  Much appreciated.

PR #1212, opened 2023-05-15
Dedicated authentication-handler #WiP

dlehammer:
Hi spring-authorization-server gurus,

I'm probably holding it wrong, but here goes :)

My use-case is pretty straight-forward; add ["security logging"](https://owasp.org/Top10/A09_2021-Security_Logging_and_Monitoring_Failures/) as recommended by OWASP.

I've looked into the authorization-server, but didn't find logging-support that matches the recommendation - my apologies if I've missed the support.

So I've created this non-exhaustive draft to demonstrate one possible approach supporting my use-case.

**Expected Behavior**

Augmenting framework error-/success-handling is a 1st class citizen.

Ex. based on the drafted approach I would be able to provide my own authentication-handler ala.

```
...

class MyOAuth2TokenEndpointFailureHandler implements AuthenticationFailureHandler {

    OAuth2TokenEndpointFailureHandler delegate = new OAuth2TokenEndpointFailureHandler()

    @Override
    public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response,
        AuthenticationException exception) throws IOException, ServletException {

        MySecurityLogger.log(request, response, exception);

        delegate.onAuthenticationFailure(request, response, exception);
    }
}
```

```
OAuth2TokenEndpointFilter#setAuthenticationFailureHandler(new MyOAuth2TokenEndpointFailureHandler())
```

As demonstrated above, this approach would support augmenting without duplicating code from the framework.

I suggest this change, from private method to dedicated class, is applied to all filter authentication-handlers - as that would support any append/augment use-case I can think of 🤓

If this approach is acceptable I would be happy to take a stab at refactoring.

**Current Behavior**

Currently each authentication-handler implemented by the framework is encapsulated in the class in a private method unfit for reuse.

**Context**

 > How has this issue affected you?

In-order to augment the endpoint filters, I've accumulated an undesired responsibility for maintaining a mirror of the default authentication-handlers implementation.

> What are you trying to accomplish?

Augment framework behavior, optimally by appending a bit of custom configuration/code - without changing the core-behavior.

> What other alternatives have you considered? 

Undoing the private encapsulation at runtime via reflection, but rejected the approach as it's brittle & time-consuming.

> Are you aware of any workarounds?

Unfortunately no :/

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
@@ -27,14 +27,12 @@ import jakarta.servlet.http.HttpServletResponse;
 
 import org.springframework.core.log.LogMessage;
 import org.springframework.http.HttpMethod;
-import org.springframework.http.HttpStatus;
 import org.springframework.http.converter.HttpMessageConverter;
 import org.springframework.http.server.ServletServerHttpResponse;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationDetailsSource;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.core.Authentication;
-import org.springframework.security.core.AuthenticationException;
 import org.springframework.security.core.context.SecurityContextHolder;
 import org.springframework.security.oauth2.core.OAuth2AccessToken;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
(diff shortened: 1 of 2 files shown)


dlehammer:
This approach might also be useful for #1111 & #925 :)

jgrandja:
Thanks for the PR @dlehammer.

I'm not sure about exposing `OAuth2TokenEndpointFailureHandler` but I agree we should somehow allow for reuse of the default `AuthenticationSuccessHandler` and `AuthenticationFailureHandler`.

We're currently planning for the `1.2` release so please give me a couple of weeks and I'll get back to you on this and we'll figure out a design approach that can't be applied across the `Filter`'s.

jgrandja:
@dlehammer Instead of exposing `OAuth2TokenEndpointFailureHandler`, we're considering adding to `OAuth2TokenEndpointFilter`:

`public static final AuthenticationFailureHandler DEFAULT_AUTHENTICATION_FAILURE_HANDLER = OAuth2TokenEndpointFilter::sendErrorResponse;`

dlehammer:
Hi @jgrandja,

Sounds like an acceptable compromise, allowing for reuse when augmenting default behaviour 👍 

Please consider utilizing the approach consistently for all default-handlers, enabling support for yet unknown use-cases 🙏

jgrandja:
@dlehammer We just merged gh-1384 so I'll close this PR in favour of the new `OAuth2ErrorAuthenticationFailureHandler`.

We will continue to work on exposing additional `AuthenticationFailureHandler` and `AuthenticationSuccessHandler` in gh-1369.

PR #1230, opened 2023-05-24
fix: same clientId but different ID values, saving triggers duplicate client error.

haoxianrui:
In the sample configuration, when saving with the same clientId and randomly generated different ID, the error message "Registered client must be unique. Found duplicate client identifier: messaging-client" is displayed.


jgrandja:
@haoxianrui 

> when saving with the same clientId and randomly generated different ID, the error message "Registered client must be unique. Found duplicate client identifier: messaging-client" is displayed.

The `demo-authorizationserver` sample saves `messaging-client` and `device-messaging-client` at startup which works correctly since the clientId do not match. If they did match then a duplicate error will be thrown, which is the expected behaviour.

I'm closing this since the sample works as expected.

PR #1231, opened 2023-05-24
Fix :spring-authorization-server-docs:asciidoctor cacheability

jprinet:
## Issues
- The task is relying on [inputs with absolute paths](https://github.com/spring-projects/spring-authorization-server/blob/878a75d6ed4a2f848316245d91fc55c774070d81/docs/spring-authorization-server-docs.gradle#L12) and therefore not getting cache hits when the cache is relocated.

- The whole `docs/src/docs/asciidoc` folder is [silently](https://github.com/asciidoctor/asciidoctor-gradle-plugin/blob/a70bd00b654976ceba099708f0b216063222ba77/base/src/main/groovy/org/asciidoctor/gradle/base/AbstractAsciidoctorBaseTask.groovy#L632) passed as an input to the task and this modifies the inputs depending on `test` task being fetched from build cache or not:

Screenshot 2023-05-23 at 4 38 38 PM

## Fix
This PR:
- converts entries from the `attributes` Map with absolute paths to relative paths (`docs-java` requires an absolute path and has to be moved into an asciidoc internal file `attributes.adoc`)
- add an exclusion on `AsciiDocTask` resources


sjohnr:
Thanks @jprinet, appreciate the effort here!

> * The task is relying on [inputs with absolute paths](https://github.com/spring-projects/spring-authorization-server/blob/878a75d6ed4a2f848316245d91fc55c774070d81/docs/spring-authorization-server-docs.gradle#L12) and therefore not getting cache hits when the cache is relocated.
> * The whole `docs/src/docs/asciidoc` folder is [silently](https://github.com/asciidoctor/asciidoctor-gradle-plugin/blob/a70bd00b654976ceba099708f0b216063222ba77/base/src/main/groovy/org/asciidoctor/gradle/base/AbstractAsciidoctorBaseTask.groovy#L632) passed as an input to the task and this modifies the inputs depending on `test` task being fetched from build cache or not:

I'm not able to find information about this cache miss in a previous successful build (possibly due to unfamiliarity with Gradle Enterprise). Are you able to link to where you see the cache miss called out in the build?

> This PR:
> 
>     * converts entries from the `attributes` Map with absolute paths to relative paths (`docs-java` requires an absolute path and has to be moved into an asciidoc internal file `attributes.adoc`)

The challenge I had with getting the docs set up originally was being able to edit docs locally in my IDE (with adoc file preview) while also having asciidoctor happy with all of the paths for resolution during a build. Asciidoctor picks a magical relative root for all paths to be resolved that the IDE has no idea about, which is quite frustrating.

The issue I'm finding with this change is that it breaks the relative pathing set up for preview. I'm also not entirely sure what it fixes. Can you explain why this change was necessary? I may not be fully understanding the need.

>     * add an exclusion on `AsciiDocTask` resources

It seems like this is the real issue with cache misses, correct? Could we just relocate the examples folder to be outside the `src/docs/asciidoc` directory? Or is there something else that makes this necessary?

jprinet:
> I'm not able to find information about this cache miss in a previous successful build (possibly due to unfamiliarity with Gradle Enterprise). Are you able to link to where you see the cache miss called out in the build?

A cache miss due to inputs relying on absolute paths happens when relocating your cache, you can demonstrate it by running experiment 3 of the [automated validation scripts](https://github.com/gradle/gradle-enterprise-build-validation-scripts/blob/main/Gradle.md).
The experience is about running 2 builds with the project cloned from 2 different locations on an empty cache. You will then get the following cache miss on the second build:

Screenshot 2023-05-25 at 9 51 37 AM

Please note that the PR also solves [these deprecation warnings](https://ge.spring.io/s/ksqtfiu5amvui/deprecations), as Gradle happens to identify some implicit dependencies between tasks likely due to [this](https://github.com/spring-projects/spring-authorization-server/blob/ef2a302a7c1c6a21edb30cf1ad04b20d9ec50c1a/docs/spring-authorization-server-docs.gradle#L13).

> The challenge I had with getting the docs set up originally was being able to edit docs locally in my IDE (with adoc file preview) while also having asciidoctor happy with all of the paths for resolution during a build. Asciidoctor picks a magical relative root for all paths to be resolved that the IDE has no idea about, which is quite frustrating.

Not sure how to make the change compliant with your use case. Would keeping the `examples-dir` variable [definition](https://github.com/spring-projects/spring-authorization-server/blob/ef2a302a7c1c6a21edb30cf1ad04b20d9ec50c1a/docs/src/docs/asciidoc/guides/how-to-pkce.adoc?plain=1#L5) in the `guides` help there?

> It seems like this is the real issue with cache misses, correct? Could we just relocate the examples folder to be outside the `src/docs/asciidoc` directory? Or is there something else that makes this necessary?

That is a part of the problem but wouldn't be sufficient afaik

jprinet:
Which IDE / plugin are you using @sjohnr?
I have tried to reproduce locally with IntelliJ and the default AsciiDoc plugin, but both default and PR branch are leading to an exception.
Screenshot 2023-05-30 at 5 35 17 PM

jprinet:
I eventually found a way to make the plugin work in my environment, I also pushed a new commit which overrides the attributes in the subfolder.
Could you give it a try @sjohnr ?

sjohnr:
> That's a good news, although I'm not super happy that an extra `guides/attributes.adoc` file is required only to make the IDE plugin happy.

I set up the guides to be a subdirectory of the documentation site, which doesn't seem to be intended by asciidoctor, so it's likely my fault. I think this is only temporary as we will eventually switch to Antora which has a more well-defined project structure and a default location for examples. I believe it has its own IDE plugin as well.

sjohnr:
Thanks @jprinet! This is now merged to `main` as 2f7c4f62d9290b240268161ac3a5cb30ab0bc542.

PR #1233, opened 2023-05-25
Fix NPE on access token in OAuth2AuthorizationCodeAuthenticationProvider

pavelefros:
Currently, (if access tokens are not persisted) line 159 can cause a NullPointerException, as we do not check if the access token is null, before calling `getToken()` on it. This issue occurs when trying to reuse authorization codes.


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
```diff
@@ -154,12 +154,12 @@ public Authentication authenticate(Authentication authentication) throws Authent
 
 		if (!authorizationCode.isActive()) {
 			if (authorizationCode.isInvalidated()) {
-				OAuth2Token token = authorization.getRefreshToken() != null ?
-						authorization.getRefreshToken().getToken() :
-						authorization.getAccessToken().getToken();
+				var token = authorization.getRefreshToken() != null ?
```
jgrandja, line 157: Please use `OAuth2Token` for the variable assignment

jgrandja (changes requested):
Thank you for the fix @pavelefros. Please see the one review comment and also please add a test for the error scenario.

pavelefros:
Hi @jgrandja,
ok, sure, thanks! I will do the necessary and push the changes in the next few days.

pavelefros:
Hi @jgrandja,
I've added the explicit type and added a test.

jgrandja:
Thanks for the updates @pavelefros. This is now merged along with a minor polish commit.

Issue #1250, opened 2023-05-31
OAuth2AuthorizationCodeRequestAuthenticationConverter can not handle multi-valued parameters


martin-lindstrom:
According to section 2.1 of [RFC 8707 - Resource Indicators for OAuth 2.0](https://www.rfc-editor.org/rfc/rfc8707.html) multiple resource values should be sent as the example below:

```
  GET /as/authorization.oauth2?response_type=code
     &client_id=s6BhdRkqt3
     &state=tNwzQ87pC6llebpmac_IDeeq-mCR2wLDYljHUZUAWuI
     &redirect_uri=https%3A%2F%2Fclient.example.org%2Fcb
     &scope=calendar%20contacts
     &resource=https%3A%2F%2Fcal.example.com%2F
     &resource=https%3A%2F%2Fcontacts.example.com%2F HTTP/1.1
  Host: authorization-server.example.com
```

The `OAuth2AuthorizationCodeRequestAuthenticationConverter.convert` method only handles single value parameters:

```
Map<String, Object> additionalParameters = new HashMap<>();
parameters.forEach((key, value) -> {
  if (!key.equals(OAuth2ParameterNames.RESPONSE_TYPE) &&
      !key.equals(OAuth2ParameterNames.CLIENT_ID) &&
      !key.equals(OAuth2ParameterNames.REDIRECT_URI) &&
      !key.equals(OAuth2ParameterNames.SCOPE) &&
      !key.equals(OAuth2ParameterNames.STATE)) {
  additionalParameters.put(key, value.get(0));
}
```
This leads to that only the first resource parameter value is saved in the resulting `OAuth2AuthorizationCodeRequestAuthenticationToken`.

jgrandja:
@martin-lindstrom Resource Indicators for OAuth 2.0 has not been implemented yet. We're already tracking it via gh-279 but it's lower priority at the moment since there is minimal upvotes so far.

Closing this as a duplicate of gh-279.

martin-lindstrom:
It really isn't about whether you support RFC8707 or not. It is a bug in `OAuth2AuthorizationCodeRequestAuthenticationConverter.convert` that only saves the first value of multi-valued parameters, independently of which unknown parameter that is being processed. OK, if you don't implement RFC8707 anytime soon, but IMHO the bug should be fixed.

jgrandja:
@martin-lindstrom Fair point.

Would you be interested in submitting a PR with the fix?

[...]

PR #1252, opened 2023-05-31
Fix to save all values for multi-valued request parameters

martin-lindstrom:
Now, multi-valued parameters are saved as String arrays in the additional parameters map of OAuth2AuthorizationCodeRequestAuthenticationToken.

Fixes gh-1250


martin-lindstrom:
I see know that most of the converters in the `org.springframework.security.oauth2.server.authorization.web.authentication` package have the same bug. I don't know how you want to address this ...

jgrandja (changes requested):
Thanks for the PR @martin-lindstrom. Please see the review comment

> I see know that most of the converters in the `org.springframework.security.oauth2.server.authorization.web.authentication` package have the same bug. I don't know how you want to address this ...

If you can apply the fixes in this PR that would be great.

jgrandja:
Thanks for the updates @martin-lindstrom !

FYI, I had to split up some of the changes as the Device Grant `AuthenticationConverter`'s was merged in `1.1.x` and `main`. The other `AuthenticationConverter`'s were back-ported to `0.4.x`. I also added a minor polish commit.

This is now merged in all the branches. Thanks again!

PR #1263, opened 2023-06-12
updated gitignore to discard gradle and gradlew changes

parvez-amplicomm:


.gitignore
```diff
@@ -9,6 +9,11 @@ target/
 *.log
 *.log.*
 
+buildSrc/gradle/
```
jgrandja, line 12: @parvez-amplicomm Why is this needed? The project never uses these 3 files. I believe this is something in your setup?

parvez-amplicomm: It's related to my local setup

jgrandja: @parvez-amplicomm Please configure your global gitignore file to include those patterns. See [gitignore](https://git-scm.com/docs/gitignore) docs for details.

Issue #1283, opened 2023-06-21
Generating ID token when sid null during refresh_token grant throws IllegalArgumentException


cbilodeauupgrade:
**Describe the bug**
Generating a refresh_token with null `sid` from the OidcIdToken claims throw  `IllegalArgumentException` 

**To Reproduce**
Don't set the `SessionRegistry`  on the `OAuth2AuthorizationCodeAuthenticationProvider` or use a `SessionRegistry` that don't return `SessionInformation` 

**Expected behavior**
Generating a refresh_token with null `sid` from the OidcIdToken claims should not throw an exception.

**Context**
According to the specification (https://openid.net/specs/openid-connect-backchannel-1_0.html#LogoutToken):
> sid
> OPTIONAL. Session ID - String identifier for a Session. This represents a Session of a User Agent or device for a logged-in End-User at an RP. Different sid values are used to identify distinct sessions at an OP. The sid value need only be unique in the context of a particular issuer. Its contents are opaque to the RP. Its syntax is the same as an OAuth 2.0 Client Identifier.

At the moment, the `sid` can be null in the OidcIdToken if the `SessionInformation` was not found but it's throw a  `IllegalArgumentException` when generating a refresh_token.

https://github.com/spring-projects/spring-authorization-server/blob/a74f90d2dfa708bbae5c3890a9a7323902493d64/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtGenerator.java#L137
should be
```java
if (currentIdToken.getClaim("sid") != null) {
  claimsBuilder.claim("sid", currentIdToken.getClaim("sid"));
}
```

jgrandja:
@cbilodeauupgrade Thanks for reporting this. Would you be interested in submitting a PR with the fix?

PR #1289, opened 2023-06-27
Fix generating ID token with null sid when refresh_token grant

cbilodeauupgrade:
fix issue https://github.com/spring-projects/spring-authorization-server/issues/1283

FYI I recreated the pull request since the build was not running on the previous one since I did something wrong


jgrandja:
Thanks for the fix @cbilodeauupgrade ! This is now merged. FYI, I added a minor polish commit.

PR #1316, opened 2023-07-27
Follow RFC to allow client_secret to be empty string

fzyzcjy:
Hi thanks for the authorization server! I am seeing an edge case here: My app originally communicates with Keycloak, and now I am migrating to Spring authorization server. The app has configured an empty client_secret (I do not fill in a strong password, because any "secret" in an app can be reverse engineered so is not a real "secret"). It worked well when sending requests (e.g. login) to Keycloak, but gives errors in Spring Auth Server.

After debugging, it seems that, if I choose a non-empty client secret, it works. I checked https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1 (linked from ClientSecretPostAuthenticationConverter), as well as https://datatracker.ietf.org/doc/html/draft-ietf-oauth-v2-1-08#section-2.4.1 (oauth 2.1). They say it is "required" but not say it is "nonempty". Indeed, oauth 2.0 says, "The client MAY omit the parameter if the client secret is an empty string", so I guess the RFC does allow an empty string.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/ClientSecretPostAuthenticationConverter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/ClientSecretPostAuthenticationConverter.java
@@ -61,9 +61,6 @@ public final class ClientSecretPostAuthenticationConverter implements Authentica
 
 		// client_secret (REQUIRED)
 		String clientSecret = parameters.getFirst(OAuth2ParameterNames.CLIENT_SECRET);
-		if (!StringUtils.hasText(clientSecret)) {
-			return null;
-		}
 
 		if (parameters.get(OAuth2ParameterNames.CLIENT_SECRET).size() != 1) {
 			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_REQUEST);


jgrandja:
@fzyzcjy `ClientSecretBasicAuthenticationConverter` and `ClientSecretPostAuthenticationConverter` both require `client_id` and `client_secret`.

The relevant information in section [2.3.1 Client Password](https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1):

> **Clients in possession of a client password** MAY use the HTTP Basic
   authentication scheme as defined in [[RFC2617](https://datatracker.ietf.org/doc/html/rfc2617)] to authenticate with
   the authorization server

> Alternatively, the authorization server MAY support **including the
   client credentials in the request-body** using the following
   parameters:

Regarding your comment:

> "The client MAY omit the parameter if the client secret is an empty string", so I guess the RFC does allow an empty string.

If the client secret is an empty string, then **the client was not assigned client credentials and only a client identifier**. In this instance, the spec states that the `client_secret` parameter **must be omitted from the request**.

I'm going to close this PR as the current implementation complies with the spec.

fzyzcjy:
@jgrandja Hi thanks for your reply! However, I am still confused. 

> Clients in possession of a client password
> including the client credentials in the request-body

The RFC does not say whether the password is an empty string or must be a non-empty string IMHO. Could you please provide a bit more details why having a password means having a non-empty password?

>    client_secret
         REQUIRED.  The client secret.  The client MAY omit the
         parameter if the client secret is an empty string.
>
> If the client secret is an empty string, then the client was not assigned client credentials and only a client identifier. In this instance, the spec states that the client_secret parameter must be omitted from the request.

I am not sure why client_secret `MUST` be omitted, since the RFC says it `MAY` be omitted.

jgrandja:
@fzyzcjy 

> Could you please provide a bit more details why having a password means having a non-empty password?

Client credentials qualifies as `client_id` and `client_secret`. An empty `client_secret` means the client was not assigned client credentials. There are no authentication systems out there that allow for empty passwords. If there are then they are insecure.

Furthermore, if an authorization server allowed a client to be configured with an empty `client_secret` and allowed the client to authenticate via HTTP Basic or the client credentials in the request-body, then client impersonation can easily be performed since the `client_id` parameter is exposed in the request and can easily be obtained by a malicious user.

> My app originally communicates with Keycloak, and now I am migrating to Spring authorization server. The app has configured an empty client_secret (I do not fill in a strong password, because any "secret" in an app can be reverse engineered so is not a real "secret"). It worked well when sending requests (e.g. login) to Keycloak, but gives errors in Spring Auth Server.

If this worked in Keycloak, then the client configured must have been a [Public Client](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1) and **NOT a Confidential Client**.

Again, the current implementations have been implemented to spec but if you still need to customize the hooks are available. See the [reference doc](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/configuration-model.html#configuring-client-authentication) for further details.

fzyzcjy:
@jgrandja Thank you for the information! I understand the explanation.

> An empty client_secret means the client was not assigned client credentials. There are no authentication systems out there that allow for empty passwords. If there are then they are insecure.

I am not an expert about the RFC, but it seems that I cannot find this sentence... At least 2.3 explicitly mentions an empty string as the secret.

> Furthermore, if an authorization server allowed a client to be configured with an empty client_secret and allowed the client to authenticate via HTTP Basic or the client credentials in the request-body, then client impersonation can easily be performed since the client_id parameter is exposed in the request and can easily be obtained by a malicious user.

I agree it can be easily hacked. However, IMHO, if you set a weak password to the client, then it is still easily hackable even if it not empty. In addition, it seems that the RFC is the thing to follow. 

> If this worked in Keycloak, then the client configured must have been a [Public Client](https://datatracker.ietf.org/doc/html/rfc6749#section-2.1) and NOT a Confidential Client.

Maybe you are right :) Then I hope this great repo supports public client as well.

> Again, the current implementations have been implemented to spec but if you still need to customize the hooks are available. See the [reference doc](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/configuration-model.html#configuring-client-authentication) for further details.

Yes I have already customized it, and I PR just because future people may need this IMHO.

jgrandja:
@fzyzcjy 

> Maybe you are right :) Then I hope this great repo supports public client as well.

Yes, it supports Public Clients. See this [How-to guide](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/guides/how-to-pkce.html) on how to configure.

fzyzcjy:
@jgrandja Thank you! However IIRC I have tried it and it did not work, and I am still confused which part of the RFC forbids empty password.

PR #1317, opened 2023-07-27
Fix NPE during token introspection.

appsandwich:
In our implementation of `OAuth2AuthorizationService` when an access token has been refreshed, the `OAuth2Authorization` can still be retrieved from the data source by passing the stale access token value into `findByToken(...)`. This is because we retain the stale access tokens in our data store, and the relationship to the original authorization is maintained.

This can cause a NPE in the `OAuth2TokenIntrospectionAuthenticationProvider` when inspecting the stale token, as the initial lookup by value (i.e. `findByToken(...)`) will retrieve the `OAuth2Authorization`. However, the call to `authorization.getToken(tokenIntrospectionAuthentication.getToken())` returns null for `authorizedToken`, because the stale token value is no longer the active access token associated with the authorization.

This PR fixes the issue by performing a null check on `authorizedToken` before checking `authorizedToken.isActive()`.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
@@ -98,7 +98,7 @@ public final class OAuth2TokenIntrospectionAuthenticationProvider implements Aut
 
 		OAuth2Authorization.Token<OAuth2Token> authorizedToken =
 				authorization.getToken(tokenIntrospectionAuthentication.getToken());
-		if (!authorizedToken.isActive()) {
+		if (authorizedToken == null || !authorizedToken.isActive()) {
 			if (this.logger.isTraceEnabled()) {
 				this.logger.trace("Did not introspect token since not active");
 			}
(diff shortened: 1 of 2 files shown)


sjohnr:
Hi @appsandwich, thanks for using SAS!

> In our implementation of `OAuth2AuthorizationService` when an access token has been refreshed, the `OAuth2Authorization` can still be retrieved from the data source by passing the stale access token value into `findByToken(...)`. This is because we retain the stale access tokens in our data store, and the relationship to the original authorization is maintained.

If an `OAuth2Authorization` can be looked up based on a particular token value, that token value should be mapped to an `OAuth2Token` by the `OAuth2AuthorizationService`. It can be inactive in various ways, but it should exist.

It sounds like you are maintaining a custom `OAuth2AuthorizationService`, is that correct? While there are different ways to go about protecting from NPEs in a codebase, the framework is not intended to work with the condition you are outlining here.

I'm not particularly keen on introducing this specific null-check because it seems caused by a customization on your end, which IMO is not valid. Instead, the service should not return the `OAuth2Authorization` in this case and should instead return `null` from the `findByToken` method. You can continue managing the data in a custom way, this is not a problem. But the framework's expectation should not be changed to accommodate the customization, and the custom implementation should be changed instead.

Based on the above explanation, I'm going to close this as declined. However, if you want to discuss this further, feel free to do so and we can re-open if needed.

appsandwich:
Hi @sjohnr, thanks so much for the detailed response and clarification.

It does indeed sound like our custom implementation of `OAuth2AuthorizationService` is behaving in an unsupported way. It should be relatively straightforward for us to modify it to return null for stale tokens.

Thanks again for the help, and for all the hard work on this project!

PR #1325, opened 2023-08-02
Fix samples test suite execution and failing tests

### LINKED ISSUE #1324: Samples test suite is not executed as part of build process
**Describe the bug**

The subproject `samples-demo-authorizationserver` test suite is not executed as part of the build and assemble flow of the project. For example running `./gradlew test` within the root project directory does not yield failing tests, while running the individual module test suite in Intellij yields test failures which are accurate -- there are two test cases that are failing for valid reasons. These may have been overlooked because the test suite is not executed properly.

**To Reproduce**

Run `./gradlew test` from root directory -- all test suites _should_ be executed.

**Expected behavior**

The subproject `samples-demo-authorizationserver` should have 2 failing test cases.

ddubson:
This PR:

- Enables JUnit5 test suite in Gradle - https://docs.gradle.org/current/userguide/java_testing.html#using_junit5
- Fixes two failing (outdated) test cases.

Closes #1324

### DIFF
--- a/samples/demo-authorizationserver/samples-demo-authorizationserver.gradle
+++ b/samples/demo-authorizationserver/samples-demo-authorizationserver.gradle
@@ -30,3 +30,7 @@ dependencies {
 	testImplementation "org.junit.jupiter:junit-jupiter"
 	testImplementation "net.sourceforge.htmlunit:htmlunit"
 }
+
+test {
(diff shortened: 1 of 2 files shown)


jgrandja:
Thanks for the fix @ddubson ! This is now merged.

PR #1345, opened 2023-08-26
Fix not to hash secret when Client Secret JWT

### LINKED ISSUE #1344: Should return hashed client_secret when registering with client_secret_jwt
**Describe the bug**
When registering a client with Client Secret JWT, client authentication is not possible because the secret is hashed.

**To Reproduce**
1. Register a client with Client Secret JWT.
2. Authenticate the client using the secret returned by the created client.

**Expected behavior**
Tokens can be issued using Client Secret JWT with the created client.

**Sample**
Look at the test code in the pull request.

**Related pull requests**
#1056 
Created by me. sorry for creating a bug.

uc4w6c:
Fix #1344


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
```diff
@@ -205,7 +205,8 @@ private OidcClientRegistrationAuthenticationToken registerClient(OidcClientRegis
 
 		RegisteredClient registeredClient = this.registeredClientConverter.convert(clientRegistrationAuthentication.getClientRegistration());
 
-		if (StringUtils.hasText(registeredClient.getClientSecret())) {
+		if (StringUtils.hasText(registeredClient.getClientSecret()) &&
+					!ClientAuthenticationMethod.CLIENT_SECRET_JWT.getValue().equals(clientRegistrationAuthentication.getClientRegistration().getTokenEndpointAuthenticationMethod())) {
```
jgrandja, line 209: We should keep the `client_secret` hashed (depending on configured `PasswordEncoder`) in order to maintain confidentiality. A solution to the fix can be as simple as having the client use the hashed `client_secret` as the key when creating the client assertion. So instead of returning the "clear" `client_secret`, return the hashed `client_secret` in the client registration response. Makes sense?

jgrandja (changes requested):
Thanks for submitting the fix @uc4w6c. Please see review comment.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcClientRegistrationAuthenticationProvider.java
uc4w6c: Thank you for your review. Is the comment's content something like this?

client_secret_jwt_hashed


In the case of client_secret_post and client_secret_basic, it seems to be effective because it uses a plain client_secret at the time of the request, and the authorization server hashes the plain client_secret and compares it with the database.
However, in the case of client_secret_jwt, the hash value is signed and verified, so even if it is hashed, I think the hash value essentially has the same meaning as a plain secret.

If you don't want to store plain secrets in the DB, I think it's better to encrypt them instead of hashing them.
But I don't think it's necessary to go that far.

For example, Keycloak currently stores secrets plainly.
https://github.com/keycloak/keycloak/discussions/8455

jgrandja: > Is the comment's content something like this?

Yes. The sequence diagrams exactly depict the flow.

uc4w6c:
@jgrandja 
I have fixed it, please review again.

PR #1348, opened 2023-08-30
Make CodeVerifierAuthenticator public

Vojtech-Sassmann:
* Make the CodeVerifierAuthenticator public, so it can be reused in custom authentication providers.


jgrandja:
@Vojtech-Sassmann `CodeVerifierAuthenticator` is intended for internal use only. Can you please log a new issue providing more details on your use case and why you require that functionality to be exposed. Thanks.

Issue #501, opened 2021-11-17
Support OpenID Connect 1.0 prompt=none parameter


ufkl:
**Expected Behavior**
To support silent authentication on Authorization Code Flow with PKCE.
Reference: https://auth0.com/docs/login/configure-silent-authentication

**Current Behavior**
Single Page Application(SPA) is public client. As it can't hold secret, the SPA need integrate with Authorization Server using Authorization Code Flow with PKCE. When access token is expired, SPA need to redirect users to Authorization Server to begin new Authorization Code Flow, result a bad user experience.

jgrandja:
@ufkl We'll look at adding support for the `prompt` parameter when we start adding further support for OpenID Connect 1.0. At the moment, we have higher priority items we're working on. We will get to this but it will take some time.

asiercamara:
Hello,

I just wanted to ask if there is any update on the implementation of this feature in the project.  The support for the prompt=none parameter would be very useful to enable silent authentication without interrupting the user in SPA applications.

(Maybe related to #211 work)

Thank you for your time and attention.

jgrandja:
@asiercamara No updates so far. Given that this feature currently has 4 upvotes, it's on the lower priority scale. We prioritize features for the next release based on upvotes and other factors.

kuzjka:
Hello,

If providing support only for `prompt=none` parameter is an option, I can suggest #1351 to fix it.

As a workaround, currently we use custom `AuthenticationEntryPoint` for this case with corresponding request filters.

jgrandja:
This is now merged via 19dfcd4

dciarniello:
I understand how `select_account` would be the responsibility of the authentication sub-system, I don't understand how that applies to `login`.  How does the flow get to the authentication system if the authorization doesn't set the authenticated state to false when `prompt=login`.

PR #1351, opened 2023-09-03
Adds support for OIDC 'prompt=none' parameter value

kuzjka:
This change provides limited support for OIDC `prompt` parameter for authorization code requests. If the request contains `prompt` parameter and it's value is `none`, authorization server does not redirect user to login/consent page. Instead, OAuth2 redirects back to client app with OIDC error codes `login_required` or `consent_required`.

Fixes gh-501 (partially)

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
@@ -91,6 +91,10 @@ public final class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilte
 	 */
 	private static final String DEFAULT_AUTHORIZATION_ENDPOINT_URI = "/oauth2/authorize";
 
+	private static final String OIDC_PROMPT_PARAMETER_NAME = "prompt";
+	private static final String OIDC_NO_PROMPT_PARAMETER_VALUE = "none";
+	private static final String OIDC_AUTH_ERROR_URL = "https://openid.net/specs/openid-connect-core-1_0.html#AuthError";
+
 	private final AuthenticationManager authenticationManager;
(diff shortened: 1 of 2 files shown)


kuzjka:
Possible improvement to be done:
- support also `login` and `consent` values for `prompt`. As I understand, they act like flags, while `none` means all flags are not set
- validate parameter value. Only four possible values are defined by the spec[^1]. `none` cannot be used with other values

Questions on implementation:
1. Since `prompt` parameter name is not added yet to `org.springframework.security.oauth2.core.oidc.endpoint.OidcParameterNames`, I have added OIDC-related constants to `OAuth2AuthorizationEndpointFilter`. We might want to keep OAuth2 components to be not aware of OIDC extensions. Probably this OIDC logic is better to be placed in a separate class.
2. I have checked redirect URL and its parameters using `UriComponentsBuilder` in tests, to avoid dependence on the order of parameters in URL. Can we rely on the order here and check the whole URL string? Or is it better to mock `authenticationFailureHandler` and check the exception coming to it?
3. Do I need to add any notes to the documentation in scope of this change? In particular to _How-to: Authenticate using a Single Page Application with PKCE_.

Thank you for your time.

[^1]: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

jgrandja:
@kuzjka Thanks for the PR. However, I'm not sure if you noticed but gh-501 is currently "on-hold" as indicated in the labels.

It's always best to reach out first before submitting a feature PR to see if the feature is planned for a release. At the moment, we have other priority tasks that we're working on and this feature is not yet planned for a release.

Furthermore, when this feature is planned for a release, we would likely require `none`, `login`, and `consent` to be implemented in the PR.

I'm going to close this PR but you can track our [project planning board](https://github.com/orgs/spring-projects/projects/8/views/21) to see when this feature is prioritized for an upcoming release and please reach out then.

kuzjka:
Thank you for the feedback, Joe.
Looking forward to new releases :)

Issue #1369, opened 2023-09-19
Allow extension and/or modification of success/error response handlers within `Filter`s


ddubson:
The purpose of this design charter is to evaluate a strategy for allowing extension and/or modification of success and error response handlers within a set of applicable server endpoints (implemented within `Filter` classes).

The present state of success and error handlers is that they are implemented mostly as private instance methods within respective `Filter` classes. (List of `Filter` classes below). 

**Success response handlers**

Success response handlers should not be fully modifiable as they contain logic critical to a stable operating auth server, but they should be customizable at designated points. The designated points should be decided in advance with the criteria being: *should a developer be able to modify this behavior?* The approach of allowing the developer to inject `Consumer<T>` classes in order to extend a certain logic of a handler may be an option. 

Success response handlers may be extendable at designated points. A developer should be able to inject their own application logic (e.g. for logging purposes) as needed — **after** the critical _default_ success response handler.

**Error response handlers**

Error response handlers are much lighter than success response handlers as they contain less logic and overall have a smaller footprint for any given `Filter` response. 

For example, the following is a subset of `Filter`s contain the same logic for error response handling which may be abstracted into a consistent pattern for a default error response handler:

- `OAuth2TokenIntrospectionEndpointFilter`
- `OAuth2TokenRevocationEndpointFilter`
- `OAuth2TokenEndpointFilter`
- `OAuth2DeviceAuthorizationEndpointFilter`

Error response handlers may be extendable at designated points. A developer should be able to inject their own application logic (i.e. for logging purposes) as needed — **after** the critical _default_ error response handler.

### Potential approaches

Potential approaches for affording **extensions** of response handlers:

- Using private scoped, static functions as default success and error handlers, which may then be used in a class using *Delegation pattern*.
- Disallow overriding default handlers (private methods) as needed.

Potential approaches for affording **modifications** of default response handlers:

- Usage of `Consumer<T>` implementations

### Appendix

The following is a list of all`Filter` classes

- **Applicable (i.e. success/error handlers are present)**
    - `OAuth2AuthorizationEndpointFilter`
    - `OAuth2ClientAuthenticationFilter`
    - `OAuth2DeviceAuthorizationEndpointFilter`
    - `OAuth2DeviceVerificationEndpointFilter`
    - `OAuth2TokenEndpointFilter`
    - `OAuth2TokenIntrospectionEndpointFilter`
    - `OAuth2TokenRevocationEndpointFilter`
[...]

PR #1429, opened 2023-10-30
Add ability to customize the access token response

ddubson:
Fixes gh-925

Referenced in gh-1369

Open questions:

- [x] Does extracting an authentication success handler for OAuth 2 Token Endpoint filter enough to make access token response customizable? Or is there a need for a customization function for adding to `additionalParameters`.
- [x] If the `Authentication` is of incorrect type, [what should the HTTP response be](https://github.com/spring-projects/spring-authorization-server/compare/main...ddubson:issue-1369-pt2?expand=1#diff-669c25197b069fd55e8fc0d7177be305a64d75033a360d00384791fabef1cb64R62-R69)?

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationContext.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationContext.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2020-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.authentication;
+
+import org.springframework.lang.Nullable;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
+import org.springframework.security.oauth2.server.authorization.web.authentication.OAuth2AccessTokenResponseAuthenticationSuccessHandler;
+import org.springframework.util.Assert;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.function.Consumer;
+
+/**
+ * An {@link OAuth2AuthenticationContext} that holds an {@link OAuth2AccessTokenResponse.Builder}
+ * and is used when customizing the building of the {@link OAuth2AccessTokenResponse}.
+ *
+ * @author Dmitriy Dubson
+ * @see OAuth2AuthenticationContext
+ * @see OAuth2AccessTokenResponse
+ * @see OAuth2AccessTokenResponseAuthenticationSuccessHandler#setAccessTokenResponseCustomizer(Consumer)
+ * @since 1.3
+ */
+public final class OAuth2AccessTokenAuthenticationContext implements OAuth2AuthenticationContext {
+	private final Map<Object, Object> context;
+
+	private OAuth2AccessTokenAuthenticationContext(Map<Object, Object> context) {
+		this.context = Collections.unmodifiableMap(new HashMap<>(context));
+	}
+
+	@SuppressWarnings("unchecked")
+	@Nullable
+	@Override
+	public <V> V get(Object key) {
+		return hasKey(key) ? (V) this.context.get(key) : null;
+	}
+
+	@Override
+	public boolean hasKey(Object key) {
+		Assert.notNull(key, "key cannot be null");
+		return this.context.containsKey(key);
+	}
+
+	/**
+	 * Returns the {@link OAuth2AccessTokenResponse.Builder} access token response builder
+	 * @return the {@link OAuth2AccessTokenResponse.Builder}
+	 */
+	public OAuth2AccessTokenResponse.Builder getAccessTokenResponse() {
+		return get(OAuth2AccessTokenResponse.Builder.class);
+	}
+
+	/**
+	 * Constructs a new {@link Builder} with the provided {@link OAuth2AccessTokenAuthenticationToken}.
+	 *
+	 * @param authentication the {@link OAuth2AccessTokenAuthenticationToken}
+	 * @return the {@link Builder}
+	 */
+	public static OAuth2AccessTokenAuthenticationContext.Builder with(OAuth2AccessTokenAuthenticationToken authentication) {
+		return new OAuth2AccessTokenAuthenticationContext.Builder(authentication);
+	}
+
+	/**
+	 * A builder for {@link OAuth2AccessTokenAuthenticationContext}
+	 */
+	public static final class Builder extends AbstractBuilder<OAuth2AccessTokenAuthenticationContext, Builder> {
+		private Builder(OAuth2AccessTokenAuthenticationToken authentication) {
+			super(authentication);
+		}
+
+		/**
+		 * Sets the {@link OAuth2AccessTokenResponse.Builder} access token response builder
+		 * @param accessTokenResponse the {@link OAuth2AccessTokenResponse.Builder}
+		 * @return the {@link Builder} for further configuration
+		 */
+		public Builder accessTokenResponse(OAuth2AccessTokenResponse.Builder accessTokenResponse) {
+			return put(OAuth2AccessTokenResponse.Builder.class, accessTokenResponse);
+		}
+
+		/**
+		 * Builds a new {@link OAuth2AccessTokenAuthenticationContext}.
+		 *
+		 * @return the {@link OAuth2AccessTokenAuthenticationContext}
+		 */
+		public OAuth2AccessTokenAuthenticationContext build() {
+			Assert.notNull(get(OAuth2AccessTokenResponse.Builder.class), "accessTokenResponse cannot be null");
+
+			return new OAuth2AccessTokenAuthenticationContext(getContext());
+		}
+	}
+}
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2023 the original author or authors.
+ * Copyright 2020-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,32 +16,24 @@
 package org.springframework.security.oauth2.server.authorization.web;
 
 import java.io.IOException;
-import java.time.temporal.ChronoUnit;
 import java.util.Arrays;
-import java.util.Map;
 
 import jakarta.servlet.FilterChain;
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServletRequest;
 import jakarta.servlet.http.HttpServletResponse;
-
 import org.springframework.core.log.LogMessage;
 import org.springframework.http.HttpMethod;
-import org.springframework.http.converter.HttpMessageConverter;
-import org.springframework.http.server.ServletServerHttpResponse;
 import org.springframework.security.authentication.AbstractAuthenticationToken;
 import org.springframework.security.authentication.AuthenticationDetailsSource;
 import org.springframework.security.authentication.AuthenticationManager;
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.context.SecurityContextHolder;
-import org.springframework.security.oauth2.core.OAuth2AccessToken;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
-import org.springframework.security.oauth2.core.OAuth2RefreshToken;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
 import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
-import org.springframework.security.oauth2.core.http.converter.OAuth2AccessTokenResponseHttpMessageConverter;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AccessTokenAuthenticationToken;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeAuthenticationProvider;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationGrantAuthenticationToken;
@@ -54,6 +46,7 @@ import org.springframework.security.oauth2.server.authorization.web.authenticati
 import org.springframework.security.oauth2.server.authorization.web.authentication.OAuth2DeviceCodeAuthenticationConverter;
 import org.springframework.security.oauth2.server.authorization.web.authentication.OAuth2ErrorAuthenticationFailureHandler;
 import org.springframework.security.oauth2.server.authorization.web.authentication.OAuth2RefreshTokenAuthenticationConverter;
+import org.springframework.security.oauth2.server.authorization.web.authentication.OAuth2AccessTokenResponseAuthenticationSuccessHandler;
 import org.springframework.security.web.authentication.AuthenticationConverter;
 import org.springframework.security.web.authentication.AuthenticationFailureHandler;
 import org.springframework.security.web.authentication.AuthenticationSuccessHandler;
@@ -61,7 +54,6 @@ import org.springframework.security.web.authentication.WebAuthenticationDetailsS
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
-import org.springframework.util.CollectionUtils;
 import org.springframework.web.filter.OncePerRequestFilter;
 
 /**
@@ -86,6 +78,7 @@ import org.springframework.web.filter.OncePerRequestFilter;
  * @author Joe Grandja
  * @author Madhu Bhat
  * @author Daniel Garnier-Moiroux
+ * @author Dmitriy Dubson
  * @since 0.0.1
  * @see AuthenticationManager
  * @see OAuth2AuthorizationCodeAuthenticationProvider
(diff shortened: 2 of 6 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandler.java
```diff
@@ -0,0 +1,104 @@
...
+	@Override
+	public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, FilterChain chain, Authentication authentication) throws IOException, ServletException {
+		AuthenticationSuccessHandler.super.onAuthenticationSuccess(request, response, chain, authentication);
+	}
+
+	@Override
+	public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
+		if (!(authentication instanceof OAuth2AccessTokenAuthenticationToken accessTokenAuthentication)) {
+			// TODO: determine what should the HTTP response be if authentication type is incorrect.
+			if (this.logger.isWarnEnabled()) {
```
jgrandja, line 64: The `authentication` must be of type `OAuth2AccessTokenAuthenticationToken`, otherwise, this `AuthenticationSuccessHandler` is misconfigured. We should throw an exception in these cases.

jgrandja:
@ddubson 

> is there a need for a customization function for adding to additionalParameters

Maybe. I see you exposed `setAccessTokenHttpResponseConverter()`, which would allow for adding additional parameters. I would suggest writing a test for the use case to see how easy it is. From there, we can decide if another hook is needed to allow for easier customization.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandler.java
ddubson: Gotcha. I've updated the block with an exception thrown, however I'm not 100% on the wording of the error message itself. That may need some polishing.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandler.java
```diff
@@ -0,0 +1,103 @@
...
+
+	@Override
+	public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, FilterChain chain, Authentication authentication) throws IOException, ServletException {
+		AuthenticationSuccessHandler.super.onAuthenticationSuccess(request, response, chain, authentication);
+	}
+
+	@Override
+	public void onAuthenticationSuccess(HttpServletRequest request, HttpServletResponse response, Authentication authentication) throws IOException, ServletException {
+		if (!(authentication instanceof OAuth2AccessTokenAuthenticationToken accessTokenAuthentication)) {
+				this.logger.error(Authentication.class.getSimpleName() + " must be of type " +
```
jgrandja, line 64: We should check `isErrorEnabled()`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandler.java
```diff
@@ -0,0 +1,103 @@
...
+						OAuth2AccessTokenAuthenticationToken.class.getName() +
+						" but was " + authentication.getClass().getName());
+			OAuth2Error error = new OAuth2Error(OAuth2ErrorCodes.SERVER_ERROR, "Unable to process access token authentication type", ERROR_URI);
```
jgrandja, line 67: Update to `new OAuth2Error(OAuth2ErrorCodes.SERVER_ERROR, "Unable to process the access token response.", null)`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandler.java
```diff
@@ -0,0 +1,103 @@
...
+/**
+ * An implementation of an {@link AuthenticationSuccessHandler} used for handling an {@link OAuth2AccessTokenAuthenticationToken}
+ * and returning the {@link OAuth2AccessTokenResponse Access Token Response}.
+ *
+ * @author Dmitriy Dubson
+ * @see AuthenticationSuccessHandler
+ * @see OAuth2AccessTokenResponseHttpMessageConverter
+ * @since 1.3
+ */
+public class OAuth2TokenEndpointAuthenticationSuccessHandler implements AuthenticationSuccessHandler {
```
jgrandja, line 50: Rename to `OAuth2AccessTokenResponseAuthenticationSuccessHandler` and make class `final`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandler.java
```diff
@@ -0,0 +1,103 @@
...
+	private final Log logger = LogFactory.getLog(getClass());
+
+	private HttpMessageConverter<OAuth2AccessTokenResponse> accessTokenHttpResponseConverter =
```
jgrandja, line 53: Rename to `accessTokenResponseConverter` and associated setter

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandlerTests.java
```diff
@@ -0,0 +1,187 @@
...
+
+		@Override
+		protected OAuth2AccessTokenResponse readInternal(Class<? extends OAuth2AccessTokenResponse> clazz, HttpInputMessage inputMessage) throws HttpMessageNotReadableException {
+			return super.readInternal(clazz, inputMessage);
+		}
+
+		@Override
+		protected void writeInternal(OAuth2AccessTokenResponse oAuth2AccessTokenResponse, HttpOutputMessage outputMessage) throws HttpMessageNotWritableException {
+			Map<String, Object> allParams = new HashMap<>(oAuth2AccessTokenResponse.getAdditionalParameters());
+			allParams.putAll(Map.of("custom_key", "custom_value", "custom_key_2", "custom_value_2"));
```
jgrandja, line 167: I'm not sure this approach will work for adding additional parameters. What if the requirement is to obtain specific information from `OAuth2Authorization` and use that for additional parameters? See the sample in this [comment](https://github.com/spring-projects/spring-authorization-server/issues/925#issuecomment-1291178579).

jgrandja (changes requested):
Thanks for the updates @ddubson. Please see review comments.

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2TokenEndpointAuthenticationSuccessHandlerTests.java
ddubson: I see your point about the authorization information injection use case. I updated the implementation to reflect this. Let me know if it makes sense.

jgrandja: @ddubson I took a look at the updated test and I don't think the `HttpMessageConverter` should have the responsibility of looking up `OAuth2Authorization` for additional information. This logic (if needed) should live in the `AuthenticationSuccessHandler`.

I'd like to propose the same pattern we've been using in the following implementations:

- `OAuth2AuthorizationConsentAuthenticationProvider.setAuthorizationConsentCustomizer()`
- `OAuth2AuthorizationCodeRequestAuthenticationProvider.setAuthenticationValidator()`
- `OidcUserInfoAuthenticationProvider.setUserInfoMapper()`

The proposal is to add:

`OAuth2AccessTokenResponseAuthenticationSuccessHandler.setAccessTokenResponseCustomizer(Consumer<OAuth2AccessTokenAuthenticationContext> authenticationContext)`

The `OAuth2AccessTokenAuthenticationContext` would contain `OAuth2AccessTokenAuthenticationToken` and `OAuth2AccessTokenResponse.Builder` for further customization.

Let me know what you think.

Also, I'm off for the holidays and returning Jan 8.
Enjoy your holidays and we'll chat when I'm back.

ddubson: Thanks for the heads up about the context object approach being used. I wasn't aware of it until now and it makes sense. From my cursory glance at the places you mentioned, it seems to be what is needed in this feature as well. Will update this thread when changes are ready for review.

ddubson: @jgrandja I've applied the context object pattern to this PR. Let me know your thoughts.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationContext.java
```diff
@@ -0,0 +1,106 @@
...
+		}
+
+		/**
+		 * Builds a new {@link OAuth2AccessTokenAuthenticationContext}.
+		 *
+		 * @return the {@link OAuth2AccessTokenAuthenticationContext}
+		 */
+		public OAuth2AccessTokenAuthenticationContext build() {
+			Assert.notNull(get(OAuth2AccessTokenResponse.Builder.class), "accessTokenResponse cannot be null");
+			Assert.notNull(get(OAuth2AccessTokenAuthenticationToken.class), "accessTokenAuthenticationToken cannot be null");
```
jgrandja, line 101: This will never be `null` since the `Builder` cannot be instantiated without `OAuth2AccessTokenAuthenticationToken`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilter.java
```diff
@@ -16,32 +16,24 @@
 package org.springframework.security.oauth2.server.authorization.web;
```
jgrandja, line 17: Update copyright year

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationContextTest.java
```diff
@@ -0,0 +1,70 @@
...
+/**
+ * Tests for {@link OAuth2AccessTokenAuthenticationContext}
+ *
+ * @author Dmitriy Dubson
+ */
+public class OAuth2AccessTokenAuthenticationContextTest {
+	private final RegisteredClient registeredClient = TestRegisteredClients.registeredClient().build();
+	private final OAuth2Authorization authorization = TestOAuth2Authorizations.authorization(this.registeredClient).build();
+	private final Authentication principal = this.authorization.getAttribute(Principal.class.getName());
+	OAuth2AccessTokenAuthenticationToken accessTokenAuthenticationToken = new OAuth2AccessTokenAuthenticationToken(registeredClient, principal,
```
jgrandja, line 41: Make `private final`

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AccessTokenAuthenticationContextTest.java
```diff
@@ -0,0 +1,70 @@
...
+	public void withWhenAuthenticationNullThenThrowIllegalArgumentException() {
+		assertThatThrownBy(() -> OAuth2AccessTokenAuthenticationContext.with(null))
+				.isInstanceOf(IllegalArgumentException.class)
+				.hasMessage("authentication cannot be null");
+	}
+
+	@Test
+	public void setWhenValueNullThenThrowIllegalArgumentException() {
+		OAuth2AccessTokenAuthenticationContext.Builder builder =
+				OAuth2AccessTokenAuthenticationContext.with(accessTokenAuthenticationToken);
```
jgrandja, line 54: Prefix `this.` for `accessTokenAuthenticationToken` and all other member variable references in this test class

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2AccessTokenResponseAuthenticationSuccessHandlerTests.java
```diff
@@ -0,0 +1,195 @@
...
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.verifyNoInteractions;
+
+/**
+ * Tests for {@link OAuth2AccessTokenResponseAuthenticationSuccessHandler}.
+ *
+ * @author Dmitriy Dubson
+ */
+public class OAuth2AccessTokenResponseAuthenticationSuccessHandlerTests {
+	RegisteredClient testClient = TestRegisteredClients.registeredClient().build();
```
jgrandja, line 66: Make `private final` and rename to `registeredClient`

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2AccessTokenResponseAuthenticationSuccessHandlerTests.java
```diff
@@ -0,0 +1,195 @@
...
+		Instant issuedAt = Instant.now();
+		Instant expiresAt = issuedAt.plusSeconds(300);
+		OAuth2Authorization testAuthorization = TestOAuth2Authorizations.authorization(testClient).build();
+		Map<String, Object> additionalParameters = Collections.singletonMap("param1", "value1");
+		Authentication authentication = new OAuth2AccessTokenAuthenticationToken(testClient, clientPrincipal,
+				testAuthorization.getAccessToken().getToken(), testAuthorization.getRefreshToken().getToken(),
+				additionalParameters);
+
+		HttpMessageConverter<OAuth2AccessTokenResponse> responseConverter = mock(HttpMessageConverter.class);
+		authenticationSuccessHandler.setAccessTokenResponseConverter(responseConverter);
```
jgrandja, line 105: Prefix `this.` for `authenticationSuccessHandler ` and all other member variable references in this test class

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2AccessTokenResponseAuthenticationSuccessHandlerTests.java
```diff
@@ -0,0 +1,195 @@
...
+	public void onAuthenticationSuccessAccessTokenResponseIsCustomizedViaAccessTokenResponseCustomizer() throws ServletException, IOException {
+		MockHttpServletRequest request = new MockHttpServletRequest();
+		MockHttpServletResponse response = new MockHttpServletResponse();
+		OAuth2AuthorizationService authorizationService = new InMemoryOAuth2AuthorizationService();
+		OAuth2Authorization testAuthorization = TestOAuth2Authorizations.authorization(testClient).build();
+		authorizationService.save(testAuthorization);
+
+		Instant issuedAt = Instant.now();
+		Instant expiresAt = issuedAt.plusSeconds(300);
+		var accessToken = testAuthorization.getAccessToken().getToken();
```
jgrandja, line 152: Let's avoid the use of `var`

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2AccessTokenResponseAuthenticationSuccessHandlerTests.java
```diff
@@ -0,0 +1,195 @@
...
+	public void setAccessTokenResponseCustomizerWhenNullThenThrowIllegalArgumentException() {
+		// @formatter:off
+		assertThatThrownBy(() -> this.authenticationSuccessHandler.setAccessTokenResponseCustomizer(null))
+				.isInstanceOf(IllegalArgumentException.class)
+				.hasMessage("accessTokenResponseCustomizer cannot be null");
+		// @formatter:on
+	}
+
+	@Test
+	public void onAuthenticationSuccessWritesAccessTokenToHttpResponse() throws ServletException, IOException {
```
jgrandja, line 92: Please change the test method names to be consistent with the naming convention used throughout all tests

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2AccessTokenResponseAuthenticationSuccessHandler.java
```diff
@@ -0,0 +1,125 @@
...
+		ServletServerHttpResponse httpResponse = new ServletServerHttpResponse(response);
+		this.accessTokenResponseConverter.write(accessTokenResponse, null, httpResponse);
+	}
+
+	/**
+	 * Sets the {@link HttpMessageConverter} used for converting an {@link OAuth2AccessTokenResponse} to an HTTP response.
+	 *
+	 * @param accessTokenResponseConverter the {@link HttpMessageConverter} used for converting an {@link OAuth2AccessTokenResponse} to an HTTP response
+	 */
+	public void setAccessTokenResponseConverter(HttpMessageConverter<OAuth2AccessTokenResponse> accessTokenResponseConverter) {
```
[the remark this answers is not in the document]
ddubson: I'm for removing the Converter setter, however, I'm bumping into an issue as to how to properly test `onAuthenticationSuccess` method without being able to observe the Converter invocation (since there would be no way to override or inject a Converter, at least for testing purposes). Specifically [here](https://github.com/spring-projects/spring-authorization-server/pull/1429/files#diff-0ae510445a9f315af67faf0c1a8abc096ffe4f0c69528514e13316d1b55ef78cR112) and one other test.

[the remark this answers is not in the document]
jgrandja: You can read the token response using this approach:

https://github.com/spring-projects/spring-authorization-server/blob/01441f917c131e0c91255fc0030703d5e6b81bbd/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/web/OAuth2TokenEndpointFilterTests.java#L282

And then assert on the values in `OAuth2AccessTokenResponse`.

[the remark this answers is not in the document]
ddubson: Ah great, thanks for the heads up! I applied the change and it works nicely.

jgrandja:
Thanks for the updates @ddubson. This is now merged!

Issue #1430, opened 2023-10-30
Allow for a configurable strategy for granting refresh_token


jgrandja:
We should consider allowing for a configurable strategy for granting `refresh_token`.

The solution should take into consideration for allowing `refresh_token` to be granted to public clients, but would be the responsibility of the consuming application to configure the strategy.

Related gh-297

stefanocke:
@jgrandja , I was able to **get** refresh tokens for a public client as shown in your test case.
However, I was not able to **use** them afterwards: 
I call the token endpoint with the refresh token and wiht grant_type "refresh_token", but I get an InsufficientAuthenticationException ("Full authentication is required to access this resource"). 

During debugging I have seen so far, that `OAuth2TokenEndpointFilter` and `OAuth2RefreshTokenAuthenticationConverter` are not even called at all.
My guess is that the reason is the missing `OAuth2ClientAuthenticationToken`, since none of the AuthenticationConverters applies here, neither `ClientSecretBasicAuthenticationConverter`, nor `ClientSecretPostAuthenticationConverter`. And especially not `PublicClientAuthenticationConverter` , since it is only for PKCE. But this is not applicable when sending a `refresh_token` to the token endpoint, but only for `authorization_code` grant. Likewise, an according `AuthenticationProvider` is missing, since `PublicClientAuthenticationProvider` only works for PKCE again.
So, the `AuthorizationFilter` fails due to missing authentication, and `OAuth2TokenEndpointFilter`, which is afterwards, is never reached.
There are extension points for both, AuthenticationConverters and AuthenticationProviders. So, would it be the right way to write a converter that looks up the client by the refresh token and then a provider that authenticates the **public** client by that?
I guess that could work, but I am a bit afraid to open some new security holes here.

**Update:**
I haved solved it as pointed out above, by a custom `AuthenticationConverter` and `AuthenticationProvider`, which are configured like this:
```
authorizationServerConfigurer.clientAuthentication(ca -> {
            ca.authenticationConverters(converters -> 
                     converters.add(new PublicClientRefreshTokenAuthenticationConverter()));
            ca.authenticationProviders(providers -> PublicClientRefreshTokenSupportingAuthenticationProvider
                    .createAndInsert(providers, registeredClientRepository, authorizationService));
        });
``` 

[...]

Issue #1422, opened 2023-10-23
How support requesting refresh_token with the offline_access scope


nverbos-godaddy:
**Expected Behavior**
want to add support for the offline_access scope as described in the [openid-connect rfc](https://openid.net/specs/openid-connect-core-1_0.html#OfflineAccess). When the offline_access scope is requested, then a refresh_token is issued.

**Current Behavior**
Currently the spring-authorization-server project issues a refresh token when a `RegisteredClient` contains `AuthorizationGrantType.REFRESH_TOKEN`. I would like to change this behavior so that the the client must request the offline_access scope in order for a refresh token to be issued.

**Context**
What is the best way to add support for this? Initially, I tried copying all of the code from `OAuth2AuthorizationCodeAuthenticationProvider` into my own custom implementation and edited [the conditional statement](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java#L216) that determines whether or not refresh token should be issued. However, I would like to avoid copying and overriding this for maintainability reasons. Is there a way to customize this for our implementation? Is this a feature that we could add to directly to the spring-authorization-server project?

jgrandja:
@nverbos-godaddy 

> I would like to change this behavior so that the the client must request the `offline_access` scope in order for a refresh token to be issued

Can you provide more details on your use case? Why do you need to change this behaviour?

> I tried copying all of the code from OAuth2AuthorizationCodeAuthenticationProvider into my own custom implementation and edited [the conditional statement](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java#L216) that determines whether or not refresh token should be issued

In order to support the `offline_access` scope within the framework, there are a few other requirements that need to be considered, for example:

> MUST ensure that the `prompt` parameter contains `consent`

FYI, the `prompt` parameter is currently not supported either gh-501

nverbos-godaddy:
> Can you provide more details on your use case? Why do you need to change this behaviour?


[...]

PR #1432, opened 2023-11-02
Allow configurable refresh token strategy for authorization_code grant

jgrandja:
Closes gh-1430

Related gh-1422

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
@@ -38,7 +38,6 @@ import org.springframework.security.core.session.SessionInformation;
 import org.springframework.security.core.session.SessionRegistry;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.ClaimAccessor;
-import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.security.oauth2.core.OAuth2AccessToken;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
@@ -213,24 +212,23 @@ public final class OAuth2AuthorizationCodeAuthenticationProvider implements Auth
 
 		// ----- Refresh token -----
 		OAuth2RefreshToken refreshToken = null;
-		if (registeredClient.getAuthorizationGrantTypes().contains(AuthorizationGrantType.REFRESH_TOKEN) &&
-				// Do not issue refresh token to public client
-				!clientPrincipal.getClientAuthenticationMethod().equals(ClientAuthenticationMethod.NONE)) {
-
+		if (registeredClient.getAuthorizationGrantTypes().contains(AuthorizationGrantType.REFRESH_TOKEN)) {
 			tokenContext = tokenContextBuilder.tokenType(OAuth2TokenType.REFRESH_TOKEN).build();
 			OAuth2Token generatedRefreshToken = this.tokenGenerator.generate(tokenContext);
-			if (!(generatedRefreshToken instanceof OAuth2RefreshToken)) {
-				OAuth2Error error = new OAuth2Error(OAuth2ErrorCodes.SERVER_ERROR,
-						"The token generator failed to generate the refresh token.", ERROR_URI);
-				throw new OAuth2AuthenticationException(error);
-			}
+			if (generatedRefreshToken != null) {
+				if (!(generatedRefreshToken instanceof OAuth2RefreshToken)) {
+					OAuth2Error error = new OAuth2Error(OAuth2ErrorCodes.SERVER_ERROR,
+							"The token generator failed to generate a valid refresh token.", ERROR_URI);
+					throw new OAuth2AuthenticationException(error);
+				}
 
-			if (this.logger.isTraceEnabled()) {
-				this.logger.trace("Generated refresh token");
-			}
+				if (this.logger.isTraceEnabled()) {
+					this.logger.trace("Generated refresh token");
+				}
 
-			refreshToken = (OAuth2RefreshToken) generatedRefreshToken;
-			authorizationBuilder.refreshToken(refreshToken);
+				refreshToken = (OAuth2RefreshToken) generatedRefreshToken;
+				authorizationBuilder.refreshToken(refreshToken);
+			}
 		}
 
 		// ----- ID token -----
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2RefreshTokenGenerator.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2RefreshTokenGenerator.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 the original author or authors.
+ * Copyright 2020-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,8 +21,11 @@ import java.util.Base64;
 import org.springframework.lang.Nullable;
 import org.springframework.security.crypto.keygen.Base64StringKeyGenerator;
 import org.springframework.security.crypto.keygen.StringKeyGenerator;
+import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
 import org.springframework.security.oauth2.core.OAuth2RefreshToken;
 import org.springframework.security.oauth2.server.authorization.OAuth2TokenType;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientAuthenticationToken;
 
 /**
  * An {@link OAuth2TokenGenerator} that generates an {@link OAuth2RefreshToken}.
@@ -42,9 +45,22 @@ public final class OAuth2RefreshTokenGenerator implements OAuth2TokenGenerator<O
 		if (!OAuth2TokenType.REFRESH_TOKEN.equals(context.getTokenType())) {
 			return null;
 		}
+		if (isPublicClientForAuthorizationCodeGrant(context)) {
+			// Do not issue refresh token to public client
+			return null;
+		}
+
 		Instant issuedAt = Instant.now();
 		Instant expiresAt = issuedAt.plus(context.getRegisteredClient().getTokenSettings().getRefreshTokenTimeToLive());
 		return new OAuth2RefreshToken(this.refreshTokenGenerator.generateKey(), issuedAt, expiresAt);
+	}
+
+	private static boolean isPublicClientForAuthorizationCodeGrant(OAuth2TokenContext context) {
+		if (AuthorizationGrantType.AUTHORIZATION_CODE.equals(context.getAuthorizationGrantType()) &&
+				(context.getAuthorizationGrant().getPrincipal() instanceof OAuth2ClientAuthenticationToken clientPrincipal)) {
+			return clientPrincipal.getClientAuthenticationMethod().equals(ClientAuthenticationMethod.NONE);
+		}
+		return false;
 	}
 
 }
--- a/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProviderTests.java
+++ b/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProviderTests.java
@@ -42,6 +42,7 @@ import org.springframework.security.core.session.SessionInformation;
 import org.springframework.security.core.session.SessionRegistry;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
 import org.springframework.security.oauth2.core.ClientAuthenticationMethod;
+import org.springframework.security.oauth2.core.OAuth2AccessToken;
 import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2ErrorCodes;
 import org.springframework.security.oauth2.core.OAuth2Token;
@@ -371,7 +372,7 @@ public class OAuth2AuthorizationCodeAuthenticationProviderTests {
 	}
 
 	@Test
-	public void authenticateWhenRefreshTokenNotGeneratedThenThrowOAuth2AuthenticationException() {
+	public void authenticateWhenInvalidRefreshTokenGeneratedThenThrowOAuth2AuthenticationException() {
 		RegisteredClient registeredClient = TestRegisteredClients.registeredClient().build();
 		OAuth2Authorization authorization = TestOAuth2Authorizations.authorization(registeredClient).build();
 		when(this.authorizationService.findByToken(eq(AUTHORIZATION_CODE), eq(AUTHORIZATION_CODE_TOKEN_TYPE)))
@@ -389,7 +390,8 @@ public class OAuth2AuthorizationCodeAuthenticationProviderTests {
 		doAnswer(answer -> {
 			OAuth2TokenContext context = answer.getArgument(0);
 			if (OAuth2TokenType.REFRESH_TOKEN.equals(context.getTokenType())) {
-				return null;
+				return new OAuth2AccessToken(
+						OAuth2AccessToken.TokenType.BEARER, "access-token", Instant.now(), Instant.now().plusSeconds(300));
 			} else {
 				return answer.callRealMethod();
 			}
@@ -400,7 +402,7 @@ public class OAuth2AuthorizationCodeAuthenticationProviderTests {
 				.extracting(ex -> ((OAuth2AuthenticationException) ex).getError())
 				.satisfies(error -> {
 					assertThat(error.getErrorCode()).isEqualTo(OAuth2ErrorCodes.SERVER_ERROR);
-					assertThat(error.getDescription()).contains("The token generator failed to generate the refresh token.");
+					assertThat(error.getDescription()).contains("The token generator failed to generate a valid refresh token.");
 				});
 	}
 
--- a/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationCodeGrantTests.java
+++ b/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationCodeGrantTests.java
@@ -55,6 +55,7 @@ import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabase;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseBuilder;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseType;
+import org.springframework.lang.Nullable;
 import org.springframework.mock.http.client.MockClientHttpResponse;
 import org.springframework.mock.web.MockHttpServletResponse;
 import org.springframework.security.authentication.AuthenticationProvider;
@@ -65,9 +66,12 @@ import org.springframework.security.config.annotation.web.configuration.EnableWe
 import org.springframework.security.core.Authentication;
 import org.springframework.security.core.GrantedAuthority;
 import org.springframework.security.core.authority.SimpleGrantedAuthority;
+import org.springframework.security.crypto.keygen.Base64StringKeyGenerator;
+import org.springframework.security.crypto.keygen.StringKeyGenerator;
 import org.springframework.security.crypto.password.NoOpPasswordEncoder;
 import org.springframework.security.crypto.password.PasswordEncoder;
 import org.springframework.security.oauth2.core.AuthorizationGrantType;
+import org.springframework.security.oauth2.core.OAuth2RefreshToken;
 import org.springframework.security.oauth2.core.OAuth2Token;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AccessTokenResponse;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
@@ -422,6 +426,54 @@ public class OAuth2AuthorizationCodeGrantTests {
 		assertThat(accessTokenAuthorization.getAccessToken()).isNotNull();
 
 		OAuth2Authorization.Token<OAuth2AuthorizationCode> authorizationCodeToken = accessTokenAuthorization.getToken(OAuth2AuthorizationCode.class);
+		assertThat(authorizationCodeToken).isNotNull();
+		assertThat(authorizationCodeToken.getMetadata().get(OAuth2Authorization.Token.INVALIDATED_METADATA_NAME)).isEqualTo(true);
+	}
+
+	// gh-1430
+	@Test
+	public void requestWhenPublicClientWithPkceAndCustomRefreshTokenGeneratorThenReturnRefreshToken() throws Exception {
+		this.spring.register(AuthorizationServerConfigurationWithCustomRefreshTokenGenerator.class).autowire();
+
+		RegisteredClient registeredClient = TestRegisteredClients.registeredPublicClient()
+				.authorizationGrantType(AuthorizationGrantType.REFRESH_TOKEN)
+				.build();
+		this.registeredClientRepository.save(registeredClient);
+
+		MvcResult mvcResult = this.mvc.perform(get(DEFAULT_AUTHORIZATION_ENDPOINT_URI)
+				.params(getAuthorizationRequestParameters(registeredClient))
+				.param(PkceParameterNames.CODE_CHALLENGE, S256_CODE_CHALLENGE)
+				.param(PkceParameterNames.CODE_CHALLENGE_METHOD, "S256")
+				.with(user("user")))
+				.andExpect(status().is3xxRedirection())
+				.andReturn();
+		String redirectedUrl = mvcResult.getResponse().getRedirectedUrl();
+		assertThat(redirectedUrl).matches("https://example.com\\?code=.{15,}&state=" + STATE_URL_ENCODED);
+
+		String authorizationCode = extractParameterFromRedirectUri(redirectedUrl, "code");
+		OAuth2Authorization authorizationCodeAuthorization = this.authorizationService.findByToken(authorizationCode, AUTHORIZATION_CODE_TOKEN_TYPE);
+		assertThat(authorizationCodeAuthorization).isNotNull();
+		assertThat(authorizationCodeAuthorization.getAuthorizationGrantType()).isEqualTo(AuthorizationGrantType.AUTHORIZATION_CODE);
+
+		this.mvc.perform(post(DEFAULT_TOKEN_ENDPOINT_URI)
+				.params(getTokenRequestParameters(registeredClient, authorizationCodeAuthorization))
+				.param(OAuth2ParameterNames.CLIENT_ID, registeredClient.getClientId())
+				.param(PkceParameterNames.CODE_VERIFIER, S256_CODE_VERIFIER))
+				.andExpect(header().string(HttpHeaders.CACHE_CONTROL, containsString("no-store")))
+				.andExpect(header().string(HttpHeaders.PRAGMA, containsString("no-cache")))
+				.andExpect(status().isOk())
+				.andExpect(jsonPath("$.access_token").isNotEmpty())
+				.andExpect(jsonPath("$.token_type").isNotEmpty())
+				.andExpect(jsonPath("$.expires_in").isNotEmpty())
+				.andExpect(jsonPath("$.refresh_token").isNotEmpty())
+				.andExpect(jsonPath("$.scope").isNotEmpty());
+
+		OAuth2Authorization authorization = this.authorizationService.findById(authorizationCodeAuthorization.getId());
+		assertThat(authorization).isNotNull();
+		assertThat(authorization.getAccessToken()).isNotNull();
+		assertThat(authorization.getRefreshToken()).isNotNull();
+
+		OAuth2Authorization.Token<OAuth2AuthorizationCode> authorizationCodeToken = authorization.getToken(OAuth2AuthorizationCode.class);
 		assertThat(authorizationCodeToken).isNotNull();
 		assertThat(authorizationCodeToken.getMetadata().get(OAuth2Authorization.Token.INVALIDATED_METADATA_NAME)).isEqualTo(true);
 	}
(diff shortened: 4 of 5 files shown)


jgrandja:
Merge via 71d923575a94ed4a06e9fa4bf3b3540e6a1309f7

armandAkop:
@jgrandja When this lands in 1.2.0, will we need to have a custom [refresh token generator] setup in our apps similar to (https://github.com/spring-projects/spring-authorization-server/pull/1432/files#diff-ca4b9c945a7879f3295084d0dc482d46f195021de6f1c120e3bb6491cc25f387R732)?

gaborbsd:
@jgrandja Thank you so much for adding this feature, now I'm able to generate refresh tokens for my public client by defining the CustomRefreshTokenGenerator as seen in the tests. However, now there is a "contradiction." The authorization_code grant works without client authentication just with PKCE but refresh_token grant does not work with client authentication. This means that just for the sake of token refresh, I need to send a client secret to the auth server, which in the case of a public client makes no sense. I know that the recommendation (and the spring-authorization-server stance) is not using refresh tokens with public clients since they mean a potential point of attack stealing them from the browser but actually, server-side rendered webapps typically store session cookies and session hijacking is a similar potential point of attack. I also see that oauth2 is a bit different because session hijacking would affect the security of the webapp itself (=client) whereas stealing a refresh token would affect the resource server. However, in my case, I have a microservices application with oauth2-based SSO so in my case, it could be seen as one system. We used to have the old Spring Boot Oauth2 Authorization Server with password grant but now we are migrating to spring-authorization-server. So I'd prefer using refresh tokens without client authentication. Could you please help me if it is possible with configuration? Or would that require further changes regarding how refresh tokens are handled? Thank you very much in advance.

jgrandja:
@gaborbsd 

I added a test that shows how to customize the client authentication configuration for allowing public clients to perform the `refresh_token` grant flow. See this commit faad0be15367b9d138be6774356adb6e77c5d386

denbakal:
Hi @jgrandja 

Is it possible to apply the custom client authentication configuration implemented in your test above when OpenID Connect 1.0 is enabled? 

**Configuration details of the ClientAuthenticationMethod & AuthorizationGrantType values:**
` .clientAuthenticationMethod(ClientAuthenticationMethod.NONE)`
`.authorizationGrantType(AuthorizationGrantType.AUTHORIZATION_CODE)`
`.authorizationGrantType(AuthorizationGrantType.REFRESH_TOKEN)`

**Scope:**
`.scope(OidcScopes.OPENID)`
`.scope(OidcScopes.PROFILE)`

**Configuration details of the ClientSettings values:**
`.requireAuthorizationConsent(false)`
`.requireProofKey(true)`

Thanks in advance.

arianox:
Does anyone have it working and using it productively?

AfeefRazick:
Yes, I got it working looking at the tests provided by jgrandja in these 2 commits [test1](https://github.com/spring-projects/spring-authorization-server/pull/1432/commits/45f292378d68d569ee7db5f92275493148ade6d7) and [test2](https://github.com/spring-projects/spring-authorization-server/commit/faad0be15367b9d138be6774356adb6e77c5d386#diff-5022e1d6e779e54073770beac732a2c23a6b4601c5679071cc66a2295afbe0e4R366]) 

By default spring provides a refresh token generator bean which prevents refresh tokens from being issued when client authentication method is NONE (which is the case for public clients with PKCE). 

The test shows how you can provide a token generator bean for refresh tokens, thus allowing you to omit the condition in the default bean.

```
@Bean
		public SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http) throws Exception {
			OAuth2AuthorizationServerConfigurer authorizationServerConfigurer =
					new OAuth2AuthorizationServerConfigurer();
			http.apply(authorizationServerConfigurer);

			authorizationServerConfigurer
					.tokenGenerator(tokenGenerator())
					.oidc(Customizer.withDefaults());	// Enable OpenID Connect 1.0

			RequestMatcher endpointsMatcher = authorizationServerConfigurer.getEndpointsMatcher();

			http
					.securityMatcher(endpointsMatcher)
					.authorizeHttpRequests(authorize ->
							authorize.anyRequest().authenticated()
					)
					.csrf(csrf -> csrf.ignoringRequestMatchers(endpointsMatcher));

			return http.build();
		}
		// @formatter:on

		@Bean
		OAuth2TokenGenerator<?> tokenGenerator() {
			JwtGenerator jwtGenerator = new JwtGenerator(new NimbusJwtEncoder(jwkSource()));
			jwtGenerator.setJwtCustomizer(jwtCustomizer());
			OAuth2TokenGenerator<OAuth2RefreshToken> refreshTokenGenerator = new CustomRefreshTokenGenerator();
			return new DelegatingOAuth2TokenGenerator(jwtGenerator, refreshTokenGenerator);
		}

		private static final class CustomRefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
			private final OAuth2RefreshTokenGenerator delegate = new OAuth2RefreshTokenGenerator();

			@Nullable
			@Override
			public OAuth2RefreshToken generate(OAuth2TokenContext context) {
				if (context.getAuthorizedScopes().contains(OidcScopes.OPENID) &&
						!context.getAuthorizedScopes().contains("offline_access")) {
					return null;
				}
				return this.delegate.generate(context);
			}

		}
```

The above test (first commit I linked above) will allow you to generate a refresh token but you will still not be able to use the refresh token without actually providing the client secret, which defeats the purpose... the second test provides the complete solution enabling you to use the refresh token with no client authentication method

```
	@EnableWebSecurity
	@Configuration(proxyBeanMethods = false)
	static class AuthorizationServerConfigurationWithPublicClientAuthentication extends AuthorizationServerConfiguration {
		// @formatter:off
		@Bean
		SecurityFilterChain authorizationServerSecurityFilterChain(
				HttpSecurity http, RegisteredClientRepository registeredClientRepository) throws Exception {
			OAuth2AuthorizationServerConfigurer authorizationServerConfigurer =
					new OAuth2AuthorizationServerConfigurer();
			authorizationServerConfigurer
					.clientAuthentication(clientAuthentication ->
							clientAuthentication
									.authenticationConverter(
											new PublicClientRefreshTokenAuthenticationConverter())
									.authenticationProvider(
											new PublicClientRefreshTokenAuthenticationProvider(registeredClientRepository))
					);
			RequestMatcher endpointsMatcher = authorizationServerConfigurer.getEndpointsMatcher();
			http
					.securityMatcher(endpointsMatcher)
					.authorizeHttpRequests(authorize ->
							authorize.anyRequest().authenticated()
					)
					.csrf(csrf -> csrf.ignoringRequestMatchers(endpointsMatcher))
					.apply(authorizationServerConfigurer);
			return http.build();
		}
		// @formatter:on
	}
	@Transient
	private static final class PublicClientRefreshTokenAuthenticationToken extends OAuth2ClientAuthenticationToken {
		private PublicClientRefreshTokenAuthenticationToken(String clientId) {
			super(clientId, ClientAuthenticationMethod.NONE, null, null);
		}
		private PublicClientRefreshTokenAuthenticationToken(RegisteredClient registeredClient) {
			super(registeredClient, ClientAuthenticationMethod.NONE, null);
		}
	}
	private static final class PublicClientRefreshTokenAuthenticationConverter implements AuthenticationConverter {
		@Nullable
		@Override
		public Authentication convert(HttpServletRequest request) {
			// grant_type (REQUIRED)
			String grantType = request.getParameter(OAuth2ParameterNames.GRANT_TYPE);
			if (!AuthorizationGrantType.REFRESH_TOKEN.getValue().equals(grantType)) {
				return null;
			}
			// client_id (REQUIRED)
			String clientId = request.getParameter(OAuth2ParameterNames.CLIENT_ID);
			if (!StringUtils.hasText(clientId)) {
				return null;
			}
			return new PublicClientRefreshTokenAuthenticationToken(clientId);
		}
	}
	private static final class PublicClientRefreshTokenAuthenticationProvider implements AuthenticationProvider {
		private final RegisteredClientRepository registeredClientRepository;
		private PublicClientRefreshTokenAuthenticationProvider(RegisteredClientRepository registeredClientRepository) {
			Assert.notNull(registeredClientRepository, "registeredClientRepository cannot be null");
			this.registeredClientRepository = registeredClientRepository;
		}
		@Override
		public Authentication authenticate(Authentication authentication) throws AuthenticationException {
			PublicClientRefreshTokenAuthenticationToken publicClientAuthentication =
					(PublicClientRefreshTokenAuthenticationToken) authentication;
			if (!ClientAuthenticationMethod.NONE.equals(publicClientAuthentication.getClientAuthenticationMethod())) {
				return null;
			}
			String clientId = publicClientAuthentication.getPrincipal().toString();
			RegisteredClient registeredClient = this.registeredClientRepository.findByClientId(clientId);
			if (registeredClient == null) {
				throwInvalidClient(OAuth2ParameterNames.CLIENT_ID);
			}
			if (!registeredClient.getClientAuthenticationMethods().contains(
					publicClientAuthentication.getClientAuthenticationMethod())) {
				throwInvalidClient("authentication_method");
			}
			return new PublicClientRefreshTokenAuthenticationToken(registeredClient);
		}
		@Override
		public boolean supports(Class<?> authentication) {
			return PublicClientRefreshTokenAuthenticationToken.class.isAssignableFrom(authentication);
		}
		private static void throwInvalidClient(String parameterName) {
			OAuth2Error error = new OAuth2Error(
					OAuth2ErrorCodes.INVALID_CLIENT,
					"Public client authentication failed: " + parameterName,
					null
			);
			throw new OAuth2AuthenticationException(error);
		}
	}
```

Suvink:
Hey @AfeefRazick, thanks for the code sample!

When generating the refresh token, this snippet creates a new instance of `OAuth2RefreshTokenGenerator` and delegates the token generation to it.
```java
   private static final class CustomRefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
        private final OAuth2RefreshTokenGenerator delegate = new OAuth2RefreshTokenGenerator();

        @Nullable
        @Override
        public OAuth2RefreshToken generate(OAuth2TokenContext context) {
            if (context.getAuthorizedScopes().contains(OidcScopes.OPENID) &&
                    !context.getAuthorizedScopes().contains("offline_access")) {
                return null;
            }
            return this.delegate.generate(context);
        }

    }
```
The check for the public client is in this `OAuth2RefreshTokenGenerator` (See [file](https://github.com/spring-projects/spring-authorization-server/blob/b76300b674bd70519d87309cf61fdba792fbed7e/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2RefreshTokenGenerator.java#L49)) and when we delegate token generation to this generator, it will always return `null`.

Instead of delegating, shouldn't we generate a token and return a `OAuth2RefreshToken`?

AfeefRazick:
> Hey @AfeefRazick, thanks for the code sample!
> 
> When generating the refresh token, this snippet creates a new instance of `OAuth2RefreshTokenGenerator` and delegates the token generation to it.
> ```java
>    private static final class CustomRefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
>         private final OAuth2RefreshTokenGenerator delegate = new OAuth2RefreshTokenGenerator();
> 
>         @Nullable
>         @Override
>         public OAuth2RefreshToken generate(OAuth2TokenContext context) {
>             if (context.getAuthorizedScopes().contains(OidcScopes.OPENID) &&
>                     !context.getAuthorizedScopes().contains("offline_access")) {
>                 return null;
>             }
>             return this.delegate.generate(context);
>         }
> 
>     }
> ```
> The check for the public client is in this `OAuth2RefreshTokenGenerator` (See [file](https://github.com/spring-projects/spring-authorization-server/blob/b76300b674bd70519d87309cf61fdba792fbed7e/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2RefreshTokenGenerator.java#L49)) and when we delegate token generation to this generator, it will always return `null`.
> 
> Instead of delegating, shouldn't we generate a token and return a `OAuth2RefreshToken`?
> 

Hi @Suvink .

In the code snippet provided

```
private static final class CustomRefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
        private final OAuth2RefreshTokenGenerator delegate = new OAuth2RefreshTokenGenerator();

        @Nullable
        @Override
        public OAuth2RefreshToken generate(OAuth2TokenContext context) {
            if (context.getAuthorizedScopes().contains(OidcScopes.OPENID) &&
                    !context.getAuthorizedScopes().contains("offline_access")) {
                return null;
            }
            return this.delegate.generate(context);
        }

    }
```

We are overriding the default refresh token generator. In our version of the refresh token generator we don't have the condition  "if pkce return null" that is in the default refresh token generator you linked.
 Therefore the refresh token we get from our custom generator is never null

AfeefRazick:
> Hey @AfeefRazick, thanks for the code sample!
> 
> When generating the refresh token, this snippet creates a new instance of `OAuth2RefreshTokenGenerator` and delegates the token generation to it.
> ```java
>    private static final class CustomRefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
>         private final OAuth2RefreshTokenGenerator delegate = new OAuth2RefreshTokenGenerator();
> 
>         @Nullable
>         @Override
>         public OAuth2RefreshToken generate(OAuth2TokenContext context) {
>             if (context.getAuthorizedScopes().contains(OidcScopes.OPENID) &&
>                     !context.getAuthorizedScopes().contains("offline_access")) {
>                 return null;
>             }
>             return this.delegate.generate(context);
>         }
> 
>     }
> ```
> The check for the public client is in this `OAuth2RefreshTokenGenerator` (See [file](https://github.com/spring-projects/spring-authorization-server/blob/b76300b674bd70519d87309cf61fdba792fbed7e/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/OAuth2RefreshTokenGenerator.java#L49)) and when we delegate token generation to this generator, it will always return `null`.
> 
> Instead of delegating, shouldn't we generate a token and return a `OAuth2RefreshToken`?
> 

Hi @Suvink .

In the code snippet provided

```
private static final class CustomRefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
        private final OAuth2RefreshTokenGenerator delegate = new OAuth2RefreshTokenGenerator();

        @Nullable
        @Override
        public OAuth2RefreshToken generate(OAuth2TokenContext context) {
            if (context.getAuthorizedScopes().contains(OidcScopes.OPENID) &&
                    !context.getAuthorizedScopes().contains("offline_access")) {
                return null;
            }
            return this.delegate.generate(context);
        }

    }
```

We are overriding the default refresh token generator. In our version of the refresh token generator we don't have the condition  "if pkce return null" that is in the default refresh token generator you linked.
 Therefore the refresh token we get from our custom generator is never null

> Instead of delegating, shouldn't we generate a token and return a OAuth2RefreshToken?

We are generating a token and returning it... via overriding the default generator bean. Spring follows the delegation mechanism for token generation.

If I have mentioned anything wrong, please let me know.

Suvink:
> "if pkce return null" 

I assume you're referring to this condition in the default RefreshTokenGenerator?
```java
if (isPublicClientForAuthorizationCodeGrant(context)) {
    // Do not issue refresh token to public client
     return null;
}
```

> We are generating a token and returning it... via overriding the default generator bean. Spring follows the delegation mechanism for token generation.

I still don't understand this part. I added some debug points and followed through the classes. The delegation mechanism works well. It first hits the default RefreshTokenGenerator and it returns `null`. Then it tries to get the token from the `CustomRefreshTokenGenerator `. Thereafter it checks for the `openid` and `offline_access` scopes and delegates the token generation. 

In the code snippet in my previous comment, we are creating a new instance of the default token generator and delegating the generation to it within the overridden generate method.
```java
private final OAuth2RefreshTokenGenerator delegate = new OAuth2RefreshTokenGenerator();
```
This default token generator has the check for the public client and will return `null`.

Following is the implementation I did for the `CustomRefreshTokenGenerator`.

```java
public class RefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {

    private final StringKeyGenerator REFRESH_TOKEN_GENERATOR = new Base64StringKeyGenerator(
            Base64.getUrlEncoder().withoutPadding(), 96);

    @Nullable
    @Override
    public OAuth2RefreshToken generate(OAuth2TokenContext context) {

        if (context.getAuthorizedScopes().contains(OidcScopes.OPENID) &&
                context.getAuthorizedScopes().contains("offline_access)) {
            return null;
        }

        Instant issuedAt = Instant.now();
        Instant expiresAt = issuedAt.plus(context.getRegisteredClient().getTokenSettings().getRefreshTokenTimeToLive());
        return new OAuth2RefreshToken(this.REFRESH_TOKEN_GENERATOR.generateKey(), issuedAt, expiresAt);
    }

}
```

WDYT?

AfeefRazick:
Ah yes, good catch. What you just provided is identical to my implementation which follows this [commit](https://github.com/spring-projects/spring-authorization-server/pull/1432/files) changes on line 951 - 986. 

```
private static final class CustomRefreshTokenGenerator implements OAuth2TokenGenerator<OAuth2RefreshToken> {
			private final StringKeyGenerator refreshTokenGenerator =
					new Base64StringKeyGenerator(Base64.getUrlEncoder().withoutPadding(), 96);

			@Nullable
			@Override
			public OAuth2RefreshToken generate(OAuth2TokenContext context) {
				if (!OAuth2TokenType.REFRESH_TOKEN.equals(context.getTokenType())) {
					return null;
				}
				Instant issuedAt = Instant.now();
				Instant expiresAt = issuedAt.plus(context.getRegisteredClient().getTokenSettings().getRefreshTokenTimeToLive());
				return new OAuth2RefreshToken(this.refreshTokenGenerator.generateKey(), issuedAt, expiresAt);
			}

		}
```

It seems I have made a mistake in my comment, I was unable to share my actual implementation so I resorted to the commit above but added the wrong snippet from that commit 😅. Apologies for the confusion.

I hope you were able to get the usage of the refresh token part working as well.

Suvink:
No worries. This is working fine now.

armandAkop:
@AfeefRazick I have been following this thread and the solution worked fine, but I ran into one issue where standard confidential clients aren't able to use the refresh token to generate a new access token.

Here is my implementation:

```kotlin
package com.myapp.authorization.configuration

import jakarta.servlet.http.HttpServletRequest
import org.springframework.security.authentication.AuthenticationProvider
import org.springframework.security.core.Authentication
import org.springframework.security.core.Transient
import org.springframework.security.crypto.keygen.Base64StringKeyGenerator
import org.springframework.security.crypto.keygen.StringKeyGenerator
import org.springframework.security.oauth2.core.AuthorizationGrantType
import org.springframework.security.oauth2.core.ClientAuthenticationMethod
import org.springframework.security.oauth2.core.OAuth2AuthenticationException
import org.springframework.security.oauth2.core.OAuth2Error
import org.springframework.security.oauth2.core.OAuth2ErrorCodes
import org.springframework.security.oauth2.core.OAuth2RefreshToken
import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames
import org.springframework.security.oauth2.server.authorization.OAuth2TokenType
import org.springframework.security.oauth2.server.authorization.authentication.OAuth2ClientAuthenticationToken
import org.springframework.security.oauth2.server.authorization.client.RegisteredClient
import org.springframework.security.oauth2.server.authorization.client.RegisteredClientRepository
import org.springframework.security.oauth2.server.authorization.token.OAuth2TokenContext
import org.springframework.security.oauth2.server.authorization.token.OAuth2TokenGenerator
import org.springframework.security.web.authentication.AuthenticationConverter
import java.time.Instant
import java.util.*

/***
 * Source: https://github.com/spring-projects/spring-authorization-server/pull/1432
 *
 * Spring authorization server doesn't support issuing refresh tokens for public clients.
 * To support this feature, we have to implement a custom authentication converter, authentication
 * provider and refresh token generator [CustomRefreshTokenGenerator]
 */

class PublicClientRefreshTokenAuthenticationConverter : AuthenticationConverter {
    override fun convert(request: HttpServletRequest): Authentication? {
        // grant_type (REQUIRED)
        val grantType = request.getParameter(OAuth2ParameterNames.GRANT_TYPE)
        if (AuthorizationGrantType.REFRESH_TOKEN.value != grantType) {
            return null
        }
        // client_id (REQUIRED)
        val clientId = request.getParameter(OAuth2ParameterNames.CLIENT_ID)
        if (clientId.isNullOrBlank()) {
            return null
        }

        return PublicClientRefreshTokenAuthenticationToken(clientId)
    }
}

@Transient
class PublicClientRefreshTokenAuthenticationToken : OAuth2ClientAuthenticationToken {

    constructor(clientId: String) : super(clientId, ClientAuthenticationMethod.NONE, null, null)

    constructor(registeredClient: RegisteredClient) : super(registeredClient, ClientAuthenticationMethod.NONE, null)
}

class PublicClientRefreshTokenAuthenticationProvider(private val registeredClientRepository: RegisteredClientRepository) :
    AuthenticationProvider {

    override fun authenticate(authentication: Authentication): Authentication? {
        val publicClientAuthentication: PublicClientRefreshTokenAuthenticationToken =
            authentication as PublicClientRefreshTokenAuthenticationToken

        if (!ClientAuthenticationMethod.NONE.equals(publicClientAuthentication.clientAuthenticationMethod)) {
            return null
        }

        val clientId: String = publicClientAuthentication.principal.toString()
        val registeredClient = registeredClientRepository.findByClientId(clientId)
        if (registeredClient == null) {
            throwInvalidClient(OAuth2ParameterNames.CLIENT_ID)
        }

        if (!registeredClient!!.clientAuthenticationMethods.contains(
                publicClientAuthentication.clientAuthenticationMethod,
            )
        ) {
            throwInvalidClient("authentication_method")
        }

        return PublicClientRefreshTokenAuthenticationToken(registeredClient)
    }

    override fun supports(authentication: Class<*>?): Boolean {
        return PublicClientRefreshTokenAuthenticationToken::class.java.isAssignableFrom(authentication)
    }

    companion object {
        private fun throwInvalidClient(parameterName: String) {
            val error = OAuth2Error(
                OAuth2ErrorCodes.INVALID_CLIENT,
                "Public client authentication failed: $parameterName",
                null,
            )
            throw OAuth2AuthenticationException(error)
        }
    }
}

/**
 * Custom refresh token generator that overrides [org.springframework.security.oauth2.server.authorization.token.OAuth2RefreshTokenGenerator]
 * to allow generating a refresh token for public clients
 */
class CustomRefreshTokenGenerator : OAuth2TokenGenerator<OAuth2RefreshToken?> {
    private val refreshTokenGenerator: StringKeyGenerator =
        Base64StringKeyGenerator(Base64.getUrlEncoder().withoutPadding(), 96)

    override fun generate(context: OAuth2TokenContext): OAuth2RefreshToken? {
        if (!OAuth2TokenType.REFRESH_TOKEN.equals(context.tokenType)) {
            return null
        }

        val issuedAt = Instant.now()
        val expiresAt = issuedAt.plus(context.registeredClient.tokenSettings.refreshTokenTimeToLive)

        return OAuth2RefreshToken(this.refreshTokenGenerator.generateKey(), issuedAt, expiresAt)
    }
}

```

This seems to break the existing functionality where confidential clients could exchange a refresh token for an access token. This code from `PublicClientRefreshTokenAuthenticationProvider` is executed and this exception is thrown:
```kotlin
        if (!registeredClient!!.clientAuthenticationMethods.contains(
                publicClientAuthentication.clientAuthenticationMethod,
            )
        ) {
            throwInvalidClient("authentication_method")
        }
```

Any help with this? Isn't there another authentication convert that is supposed to handle this case before PublicClientRefreshTokenAuthenticationConverter kicks in?

AfeefRazick:
Hi, I am unable to confirm this right now, but you are probably right about having to add another refresh token generator.

Something like the default OAuth2RefreshTokenGenerator implementation by spring, where we remove some of those error throws with `return null`

You will also likely have to remove that throw method in the PublicClientRefreshTokenGenerator with 
`return null` as well, allowing it to pass to the following token generators.

This is an assumption, I cannot confirm as of now.

AfeefRazick:
Hi, I am unable to confirm this right now, but you are probably right about having to add another converter before this converter.

Either springs default authentication converter (I assume ClientSecretPostAuthenticationConverter, or one of the converters listed here, maybe 
https://docs.spring.io/spring-authorization-server/docs/current/api/org/springframework/security/oauth2/server/authorization/web/OAuth2ClientAuthenticationFilter.html), if that implementation has any error throws 
preventing refresh token conversion for public clients, you will have to remove them.

Another issue is that the PublicClientAuthenticarionConverter we have here is assuming any refresh token request is intended for it, when it should only look at refresh token requests without a client secret. So if you can implement a check for "if client secret is null". That may also work, allowing it to pass to the following converter implementations.

This is an assumption, I can not confirm as of now.

armandAkop:
I updated my converter implementation to check if a client_secret is present in the request, we consider it a confidential client and return null, ensuring the other converts get a chance to authenticate, and it seems to work with a few of my test cases. Here is the code

```kotlin
    override fun convert(request: HttpServletRequest): Authentication? {
        // grant_type (REQUIRED)
        val grantType = request.getParameter(OAuth2ParameterNames.GRANT_TYPE)
        if (AuthorizationGrantType.REFRESH_TOKEN.value != grantType) {
            return null
        }

        // client_id (REQUIRED)
        val clientId = request.getParameter(OAuth2ParameterNames.CLIENT_ID)
        if (clientId.isNullOrBlank()) {
            return null
        }

        // client_secret (must be excluded to be considered "public" client)
        val clientSecret = request.getParameter(OAuth2ParameterNames.CLIENT_SECRET)
        if (!clientSecret.isNullOrBlank()) {
            return null
        }

        return PublicClientRefreshTokenAuthenticationToken(clientId)
    }
```

AfeefRazick:
I'm glad that previous solution helped.

About the revoking, I'm honestly not sure myself. However, this is where I can point you to.

It seems that for the revocation, there is a different chain of revocation authentication converters as well as providers.  And it can be configured here. [OAuth2TokenRevocationEndpointConfigurer](https://docs.spring.io/spring-authorization-server/reference/protocol-endpoints.html#oauth2-token-revocation-endpoint)

It seems that this is the (or one of many) default converter. [OAuth2TokenRevocationAuthenticationConverter](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server%2Fsrc%2Fmain%2Fjava%2Forg%2Fspringframework%2Fsecurity%2Foauth2%2Fserver%2Fauthorization%2Fweb%2Fauthentication%2FOAuth2TokenRevocationAuthenticationConverter.java)

shinokamparos:
Has anyone got a sucessful implementation of token revocation?

Suvink:
> Has anyone got a sucessful implementation of token revocation?

Isn't this supported OOTB in Spring Authz Server?

armandAkop:
@shinokamparos Here is what an implementation could look like, would be good to get some eyes on this from other folks more familiar with spring-auth-server architecture. Note: this is in Kotlin and a lot of the code was borrowed from other authentication converters/providers in the codebase.


```kotlin
class PublicClientOAuth2RefreshTokenRevocationConverter : AuthenticationConverter {
    override fun convert(request: HttpServletRequest): Authentication? {
        val parameters = getFormParameters(request)

        /**
         * We can determine this is a refresh token revocation for a public client with the
         * following:
         * - tokenTypeHint must be "refresh_token"
         * - client_secret must NOT be present
         *
         *   otherwise return null to allow remaining converters to process request
         */
        val tokenTypeHint = parameters.getFirst("token_type_hint")
        val clientSecret = parameters.getFirst("client_secret")
        if (tokenTypeHint != "refresh_token" || !clientSecret.isNullOrBlank()) {
            return null
        }

        val token = parameters.getFirst("token")
        if (token.isNullOrBlank()) {
            throwError(OAuth2ErrorCodes.INVALID_REQUEST, "token")
        }

        val clientId = parameters.getFirst("client_id")
        if (clientId.isNullOrBlank()) {
            throwError(OAuth2ErrorCodes.INVALID_REQUEST, "client_id")
        }

        return PublicClientRefreshTokenRevocationAuthenticationToken(
            clientId!!,
            refreshToken = token!!,
        )
    }

    private fun throwError(errorCode: String, parameterName: String) {
        val error =
            OAuth2Error(
                errorCode,
                "OAuth 2.0 Token Revocation Parameter: $parameterName",
                "https://datatracker.ietf.org/doc/html/rfc7009#section-2.1",
            )
        throw OAuth2AuthenticationException(error)
    }

    private fun getFormParameters(
        request: HttpServletRequest
    ): LinkedMultiValueMap<String?, String?> {
        val parameterMap = request.parameterMap
        val parameters: LinkedMultiValueMap<String?, String?> =
            LinkedMultiValueMap<String?, String?>()
        parameterMap.forEach { (key: String?, values: Array<String?>?) ->
            val queryString =
                if (StringUtils.hasText(request.queryString)) request.queryString else ""
            if (!queryString.contains(key) && values.isNotEmpty()) {
                for (value in values) {
                    parameters.add(key, value)
                }
            }
        }
        return parameters
    }
}

class PublicClientOAuth2RefreshTokenRevocationProvider(
    val registeredClientRepository: RegisteredClientRepository,
    val authorizationService: OAuth2AuthorizationService,
) : AuthenticationProvider {
    override fun authenticate(authentication: Authentication?): Authentication {
        val clientAuthentication =
            authentication as PublicClientRefreshTokenRevocationAuthenticationToken
        val clientId = clientAuthentication.principal.toString()

        val registeredClient = registeredClientRepository.findByClientId(clientId)
        if (registeredClient == null) {
            throwInvalidClient("client_id")
        }
        registeredClient!!

        if (
            !registeredClient.clientAuthenticationMethods.contains(
                clientAuthentication.clientAuthenticationMethod
            )
        ) {
            throwInvalidClient("authentication_method")
        }

        val authorization =
            authorizationService.findByToken(
                clientAuthentication.refreshToken,
                OAuth2TokenType.REFRESH_TOKEN,
            )

        if (authorization == null) {
            throw OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT)
        }

        // Attempting to refresh another client's refresh token
        if (!registeredClient.id.equals(authorization.registeredClientId)) {
            throw OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_CLIENT)
        }

        return OAuth2ClientAuthenticationToken(
            registeredClient,
            ClientAuthenticationMethod.NONE,
            null,
        )
    }

    override fun supports(authentication: Class<*>?): Boolean {
        return PublicClientRefreshTokenRevocationAuthenticationToken::class
            .java
            .isAssignableFrom(authentication)
    }

    private fun throwInvalidClient(parameterName: String) {
        val error =
            OAuth2Error(
                OAuth2ErrorCodes.INVALID_CLIENT,
                "Client authentication failed: " + parameterName,
                "https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1",
            )
        throw OAuth2AuthenticationException(error)
    }
}

@Transient
class PublicClientRefreshTokenRevocationAuthenticationToken(
    clientId: String,
    var refreshToken: String,
) : OAuth2ClientAuthenticationToken(clientId, ClientAuthenticationMethod.NONE, null, null)

// Register converter/provider in authorization server configurer
authorizationServerConfigurer
.clientAuthentication {
  it.authenticationConverter(PublicClientOAuth2RefreshTokenRevocationConverter())
    .authenticationProvider(
        PublicClientOAuth2RefreshTokenRevocationProvider(
            registeredClientRepository,
            authorizationService,
        )
    )
}
```

The key here is to have `PublicClientOAuth2RefreshTokenRevocationProvider` return an authenticated `OAuth2ClientAuthenticationToken` which [OAuth2TokenRevocationAuthenticationProvider](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenRevocationAuthenticationProvider.java) picks up and performs the revoke later in the chain.

I ended up diving into the code and setting up a breakpoints to see how the flow works, and this was the solution I came up with which seems to work.

yuezk:
> This seems to break the existing functionality where confidential clients could exchange a refresh token for an access token. This code from PublicClientRefreshTokenAuthenticationProvider is executed and this exception is thrown:
>
> Any help with this? Isn't there another authentication converter that is supposed to handle this case before PublicClientRefreshTokenAuthenticationConverter kicks in?

Hi @armandAkop, I know you have resolved your problem. But the reason you encountered the issue above is that the `PublicClientRefreshTokenAuthenticationConverter` is added at the head of the client authentication converter list, which will be matched first and returned when calling the token endpoint.

So instead of calling `it.authenticationConverter(PublicClientRefreshTokenAuthenticationConverter())`, you should call ` it.authenticationConverters(converters -> converters.add(new PublicClientRefreshTokenAuthenticationConverter()))`.

The latter is an authentication converter customizer bean; We can add the new converter to the end of the converter list.

Below is my configuration:

```java
              .clientAuthentication(ca -> {
                  ca.authenticationConverters(converters -> converters.add(new PublicClientRefreshTokenAuthenticationConverter()));
                  ca.authenticationProviders(providers -> providers.add(new PublicClientRefreshTokenAuthenticationProvider(registeredClientRepository)));
              });
```

Issue #1449, opened 2023-11-17
Add parameter name to JwtClientAssertionAuthenticationConverter error messages


felix-hellman:
**Expected Behavior**
When client assertions fail validation, the exception message contains the parameter name.

**Current Behavior**
When client assertions fail validation the exception message is empty. 

**Context**
Since the validation is for internal use and does not expose what parameter has failed validation, it is required to wrap the existing converter and re-run the validation adding the failed parameter name. Adding this will save time when troubleshooting what the client is sending incorrectly without adding additional logging since the exception is already thrown.

jgrandja:
@felix-hellman Thanks for the PR associated to this issue. However, the changes in the PR would make the `OAuth2Error` inconsistent with the other client authentication converters `ClientSecretBasicAuthenticationConverter`, `ClientSecretPostAuthenticationConverter` and `PublicClientAuthenticationConverter`.

Furthermore, I'm not sure enhancing the error message to include the parameter name is the long term solution we're looking. I think we'll need to address this in gh-1240 and implement a holistic solution.

I'm going to close this issue and associated PR but feel free to add any additional comments to gh-1240.

PR #1450, opened 2023-11-17
Adds paramater name to JwtClientAssertion validation error Fixes gh-1449

felix-hellman:
The validation could be re-written to validate all parameters and add the name of all invalid parameters, however, this change is more minimal as it only requires a small refactor of the existing code.


jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-authorization-server/issues/1449#issuecomment-1826353509)

Issue #541, opened 2021-12-17
How-to: Handle errors and customize the OAuth 2.0 Error response


jgrandja:
Publish a guide on How-to: Handle errors and customize the OAuth 2.0 Error response

Related gh-499

jgrandja:
Sample demonstrating how to customize the error response parameters using `OAuth2ErrorAuthenticationFailureHandler`:

```java
private AuthenticationFailureHandler authenticationFailureHandler = createAuthenticationFailureHandler();

private static AuthenticationFailureHandler createAuthenticationFailureHandler() {
	OAuth2ErrorHttpMessageConverter errorResponseConverter = new OAuth2ErrorHttpMessageConverter();
	errorResponseConverter.setErrorParametersConverter(error -> {
		Map<String, String> parameters = new HashMap<>();
		parameters.put(OAuth2ParameterNames.ERROR, error.getErrorCode());
		return parameters;
	});

	OAuth2ErrorAuthenticationFailureHandler authenticationFailureHandler = new OAuth2ErrorAuthenticationFailureHandler();
	authenticationFailureHandler.setErrorResponseConverter(errorResponseConverter);

	return authenticationFailureHandler;
}
```

jgrandja:
Sample demonstrating how to customize the HTTP status for the error response using `OAuth2ErrorAuthenticationFailureHandler`:

```java
private AuthenticationFailureHandler authenticationFailureHandler = createAuthenticationFailureHandler();

private static AuthenticationFailureHandler createAuthenticationFailureHandler() {
	OAuth2ErrorHttpMessageConverter errorResponseConverter = new OAuth2ErrorHttpMessageConverter() {
		@Override
		protected void writeInternal(OAuth2Error error, HttpOutputMessage outputMessage)
				throws HttpMessageNotWritableException {

			HttpServletResponse servletResponse = ((ServletServerHttpResponse) outputMessage).getServletResponse();
			if (OAuth2ErrorCodes.INVALID_CLIENT.equals(error.getErrorCode())) {
				servletResponse.setStatus(HttpStatus.UNAUTHORIZED.value());
			} else {
				servletResponse.setStatus(HttpStatus.BAD_REQUEST.value());
			}
			super.writeInternal(error, outputMessage);
		}

	};

	OAuth2ErrorAuthenticationFailureHandler authenticationFailureHandler = new OAuth2ErrorAuthenticationFailureHandler();
[...]

PR #1465, opened 2023-11-28
Allow customization of error and HTTP response within OAuth2AuthenticationFailureHandler

ddubson:
This is an initial proof-of-concept for the last segment of gh-1369, as stated in the original:

> For the second iteration, let's see if we can come up with a hook in the new `AuthenticationFailureHandler` that will allow for reuse and customization for these additional Filter's:
> 
> `OAuth2ClientAuthenticationFilter`
> `OidcUserInfoEndpointFilter`
> `OidcClientRegistrationEndpointFilter`

There are currently failing existing test cases, and no additional new test cases, as this draft PR is meant for early review of the approach.

Once the rough sketch is approved for this implementation, the TODO list is as follows:

* [ ] Fix failing existing test cases
* [ ] Add test coverage for the newly implemented functionality
* [ ] Polish and document newly added functions or blocks

Fixes gh-1369

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2ErrorAuthenticationFailureHandler.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2ErrorAuthenticationFailureHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2023 the original author or authors.
+ * Copyright 2020-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,9 @@
 package org.springframework.security.oauth2.server.authorization.web.authentication;
 
 import java.io.IOException;
+import java.util.Map;
+import java.util.function.BiConsumer;
+import java.util.function.Function;
 
 import jakarta.servlet.ServletException;
 import jakarta.servlet.http.HttpServletRequest;
@@ -24,6 +27,7 @@ import jakarta.servlet.http.HttpServletResponse;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import org.springframework.core.convert.converter.Converter;
 import org.springframework.http.HttpStatus;
 import org.springframework.http.converter.HttpMessageConverter;
 import org.springframework.http.server.ServletServerHttpResponse;
@@ -46,14 +50,15 @@ import org.springframework.util.Assert;
 public final class OAuth2ErrorAuthenticationFailureHandler implements AuthenticationFailureHandler {
 	private final Log logger = LogFactory.getLog(getClass());
 	private HttpMessageConverter<OAuth2Error> errorResponseConverter = new OAuth2ErrorHttpMessageConverter();
+	private BiConsumer<OAuth2AuthenticationException, ServletServerHttpResponse> httpResponseCustomizer = (exception, httpResponse) -> httpResponse.setStatusCode(HttpStatus.BAD_REQUEST);
 
 	@Override
 	public void onAuthenticationFailure(HttpServletRequest request, HttpServletResponse response,
 			AuthenticationException authenticationException) throws IOException, ServletException {
-		ServletServerHttpResponse httpResponse = new ServletServerHttpResponse(response);
-		httpResponse.setStatusCode(HttpStatus.BAD_REQUEST);
+		if (authenticationException instanceof OAuth2AuthenticationException exception) {
+			ServletServerHttpResponse httpResponse = new ServletServerHttpResponse(response);
+			httpResponseCustomizer.accept(exception, httpResponse);
 
-		if (authenticationException instanceof OAuth2AuthenticationException) {
 			OAuth2Error error = ((OAuth2AuthenticationException) authenticationException).getError();
 			this.errorResponseConverter.write(error, null, httpResponse);
 		} else {
@@ -75,4 +80,7 @@ public final class OAuth2ErrorAuthenticationFailureHandler implements Authentica
 		this.errorResponseConverter = errorResponseConverter;
 	}
 
+	public void setHttpResponseCustomizer(BiConsumer<OAuth2AuthenticationException, ServletServerHttpResponse> httpResponseCustomizer) {
+		this.httpResponseCustomizer = httpResponseCustomizer;
+	}
 }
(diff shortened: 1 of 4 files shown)


ddubson:
@jgrandja -- I'd like to get your input as to the approach for this implementation, let me know if you need further clarification on any part of this draft.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2ErrorAuthenticationFailureHandler.java
```diff
@@ -47,14 +49,18 @@ public final class OAuth2ErrorAuthenticationFailureHandler implements Authentica
 	private final Log logger = LogFactory.getLog(getClass());
 	private HttpMessageConverter<OAuth2Error> errorResponseConverter = new OAuth2ErrorHttpMessageConverter();
 
+	private BiConsumer<OAuth2AuthenticationException, ServletServerHttpResponse> httpResponseCustomizer = (exception, httpResponse) -> httpResponse.setStatusCode(HttpStatus.BAD_REQUEST);
+
+	private Function<OAuth2AuthenticationException, OAuth2Error> errorCustomizer = OAuth2AuthenticationException::getError;
```
jgrandja, line 54: I don't think this is needed as I can customize the error parameters by providing a custom `Converter` to `OAuth2ErrorHttpMessageConverter.setErrorParametersConverter()`.

For example, I can customize `OAuth2ClientAuthenticationFilter.authenticationFailureHandler` to use `OAuth2ErrorAuthenticationFailureHandler` but with a custom `OAuth2ErrorHttpMessageConverter.setErrorParametersConverter()` that only returns the `OAuth2Error.errorCode`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2ErrorAuthenticationFailureHandler.java
```diff
@@ -47,14 +49,18 @@ public final class OAuth2ErrorAuthenticationFailureHandler implements Authentica
 	private final Log logger = LogFactory.getLog(getClass());
 	private HttpMessageConverter<OAuth2Error> errorResponseConverter = new OAuth2ErrorHttpMessageConverter();
 
+	private BiConsumer<OAuth2AuthenticationException, ServletServerHttpResponse> httpResponseCustomizer = (exception, httpResponse) -> httpResponse.setStatusCode(HttpStatus.BAD_REQUEST);
```
jgrandja, line 52: I'm not sure how I feel about using a `BiConsumer`. Currently, the framework does not make use of `BiConsumer`. I'll have to think about this for a bit.

jgrandja (changes requested):
Thanks for the PR @ddubson. Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2ErrorAuthenticationFailureHandler.java
ddubson: Will a Context class work in this situation? Now that I've learned about that pattern, I can see it fitting in here, however I'm not sure if the project uses it in handling error conditions as this warrants.

jgrandja: I don't think it makes sense to use a context class here.

I've been giving this some more thought and I'm starting to wonder if any changes are needed after all.

We have the ability to customize the error response parameters as mentioned in this [comment](https://github.com/spring-projects/spring-authorization-server/pull/1465#discussion_r1420240417), so the only other customization required at this point is to change the status code. However, this can be done using this:


```java
private AuthenticationFailureHandler authenticationFailureHandler = createAuthenticationFailureHandler();

private static AuthenticationFailureHandler createAuthenticationFailureHandler() {
	OAuth2ErrorHttpMessageConverter errorResponseConverter = new OAuth2ErrorHttpMessageConverter() {
		@Override
		protected void writeInternal(OAuth2Error error, HttpOutputMessage outputMessage)
				throws HttpMessageNotWritableException {

			HttpServletResponse servletResponse = ((ServletServerHttpResponse) outputMessage).getServletResponse();
			if (OAuth2ErrorCodes.INVALID_CLIENT.equals(error.getErrorCode())) {
				servletResponse.setStatus(HttpStatus.UNAUTHORIZED.value());
			} else {
				servletResponse.setStatus(HttpStatus.BAD_REQUEST.value());
			}
			super.writeInternal(error, outputMessage);
		}

	};
	errorResponseConverter.setErrorParametersConverter(error -> {
		Map<String, String> parameters = new HashMap<>();
		parameters.put(OAuth2ParameterNames.ERROR, error.getErrorCode());
		return parameters;

	});

	OAuth2ErrorAuthenticationFailureHandler authenticationFailureHandler = new OAuth2ErrorAuthenticationFailureHandler();
	authenticationFailureHandler.setErrorResponseConverter(errorResponseConverter);

	return authenticationFailureHandler;
}
```

This `AuthenticationFailureHandler` would work for `OAuth2ClientAuthenticationFilter`. I know it's not pretty though.

I'm just having 2nd thoughts on introducing a new API just so we can customize the status code. Furthermore, no one from the community has asked for this specific customization. I realize this is my ask in the hopes of reusing `OAuth2ErrorAuthenticationFailureHandler` for the other 3x `Filter`'s.

What are your thoughts?

ddubson: If there's little to no demand from the developer/consumer side, I think a good course of action is to document the approach you mentioned above as a way to do the error customization somewhere within the official docs -- that may very well be useful to those that need it in the future, without spending time and effort for an API change.

Additionally, a test case demonstrating this customization approach could also be useful in tandem with the docs change.

jgrandja: > I think a good course of action is to document the approach you mentioned above as a way to do the error customization somewhere within the official docs -- that may very well be useful to those that need it in the future, without spending time and effort for an API change.

Fully agree and good idea on adding the sample customization to the reference. For now, I added the sample code to gh-541 as it should be added as part of that How-to guide.

I'm going to close this PR as we'll hold off on any API changes at this point. Thank you for spending your time on coming up with the right approach 👍

Issue #1240, opened 2023-05-27
Consider enhancing OAuth2Error with specific error codes


Yneth:
**Expected Behavior**
I want to have the option to override OAuth2 Error codes that are thrown from the `spring-authorization-server` source code.

**Current Behavior**
I am not able to override default OAuth2 Error codes.

**Context**
In our project, we want to provide more meaningful error codes to clients, as sometimes the standard OAuth2 error codes are too vague, such as `invalid_grant`. In our client's code, we want to have specific error codes for each problem so that they can handle each case differently.

We can currently map error codes by parsing the `error_description`, but we want to avoid that.

It would be amazing if we had either more 
(A) detailed error codes or
(B) an additional internal list of error codes that would allow for more precise usage.

I understand it could violate RFC but we are ok with that.

**Example**
Option A:
instead of just `invalid_request` say `missing_client_id`

Option B:
send both `invalid_request` and `missing_client_id`

jgrandja:
@Yneth The [Protocol Endpoints](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/protocol-endpoints.html) provide hooks into customizing the error response (including error code) by configuring a custom `AuthenticationFailureHandler`. For example, the [Token Endpoint](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/protocol-endpoints.html#oauth2-token-endpoint) provides `authorizationServerConfigurer.tokenEndpoint.errorResponseHandler()` for customizing the [OAuth2Error response](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2).

FYI, we're also planning on writing a How-to guide in gh-541.

I'll close this as customizing the error response is available.

Yneth:
I will try to explain my issue better as in my case using `AuthenticationFailureHandler` is not enough.
Let's consider the following lines of code:
* https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java#L147
* https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java#L152

in those two cases, it would be impossible to enrich the error code to something more granular, like `invalid_client_id` or `invalid_redirect_uri`.
[...]

PR #1467, opened 2023-12-01
Improve logging

leshalv:
https://github.com/spring-projects/spring-authorization-server/issues/1240


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeAuthenticationProvider.java
```diff
@@ -143,12 +143,17 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+			throw new OAuth2AuthenticationException(error);
 		}
 
 		if (StringUtils.hasText(authorizationRequest.getRedirectUri()) &&
 				!authorizationRequest.getRedirectUri().equals(authorizationCodeAuthentication.getRedirectUri())) {
-			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
+			if (this.logger.isWarnEnabled()) {
+				this.logger.warn(LogMessage.format("Invalidated redirect_uri used by registered client '%s'", registeredClient.getId()));
+			}
+			OAuth2Error error = new OAuth2Error(OAuth2ErrorCodes.INVALID_GRANT,"The redirect_uri does not match the redirection URI used in the authorization request.",ERROR_URI);
```
jgrandja, line 155: @leshalv We do not want to reveal too much information to the caller as it can provide hints to perform other operations in order to gain access. It is intentional to return only a general error code.

At some point, we may look at addressing this in gh-1240.

I'm going to close this PR for reasons explained above.

leshalv: When docking integration, it is easy to have some problems, when troubleshooting problems, it is difficult to know the specific problem, here is my reference to okta back to do, which should not affect major security

leshalv: Or we could consider the PR, including its log input, masking this detail in the exception handler, which can be customized by a custom exception handler if the developer wants to return it.
@jgrandja

leshalv: At present, I look at the implementation, only the server exception is returned details, in fact, we can do not return details of the INVALID_GRANT exception in the exception handler, the decision is left to the user-defined processor.
@jgrandja

leshalv: We are currently developing IAM systems based on springsecurity, and there are a lot of issues around exceptions, because there is no log print there, it is difficult to know what is going on. I have also looked at Okta, Auth0 and other SSO platforms, and I find that everyone has different wrong responses. I can also override this implementation, but considering the connection with the upstream, including subsequent upgrades, I think this piece should be synchronized to the upstream, perhaps we can optimize this piece through other ways, such as in the exception processor mask, the user through the custom processor open.
@jgrandja

jgrandja: @leshalv I would be open to enhancing the `trace` logging if that works for you?

leshalv: > @leshalv I would be open to enhancing the `trace` logging if that works for you?

Strengthening `trace` was a good decision.
Can we consider the scheme I said we described in exception handler masking? 
Because we've done the same thing on other processors.

leshalv:
@jgrandja I made logging enhancements in some key places.

jgrandja:
@leshalv Apologies I didn't have time to review the changes.

I'm off for the holidays and back Jan 8. I will review it the week I'm back.

Enjoy the holidays!

jgrandja:
Thanks for the updates @leshalv. This is now merged. FYI, I added a polish commit with some minor updates.

PR #1471, opened 2023-12-07
Fix Spring Authorization Server Reference hyperlink on README.adoc

OSKUDA:
The hyperlink for the 'Spring Authorization Server Reference' on README.adoc was incorrect and points to an non existing resource. It redirects the user to 'https://docs.spring.io/spring-authorization-server/docs/current/reference/html/' instead of 'https://docs.spring.io/spring-authorization-server/reference/'

Have updated the README.adoc with the correct url.


jgrandja:
Thanks for the update @OSKUDA! This is now merged.

PR #1490, opened 2024-01-02
Build directories are excluded from the checkstyleNohttp task

erichaagdev:
This change addresses a build deprecation warning causing caching to be disabled for the `checkstyleNohttp`. The deprecation warning is caused by [the `checkstyleNohttp` task using the output files of other tasks, without creating an implicit or explicit dependency between them](https://ge.solutions-team.gradle.com/s/5kw3tind56qii/deprecations). As a result, [Gradle disables caching in order to ensure correctness of the build](https://ge.solutions-team.gradle.com/s/5kw3tind56qii/timeline?details=z56sfccjdrmfs).

With this change, the `checkstyleNohttp` task will ignore all build directories. This [resolves the deprecation warnings](https://ge.solutions-team.gradle.com/s/drhuxrwqyxb7c/deprecations) and [allows the task to be cached](https://ge.solutions-team.gradle.com/s/drhuxrwqyxb7c/timeline?details=z56sfccjdrmfs).


sjohnr:
Thanks @erichaagdev! This is merged into `1.1.x` (and merged forward) as f6b07f5649fe5cec0507d0448733b389939a0960, so we can attempt to upgrade our OSS supported 1.x branches to Gradle 8.

Issue #1454, opened 2023-11-20
Allow the ability to deactivate the device authorization grant


ddittmar:
**Expected Behavior**

I would like to have the ability to deactivate certain flows (or endpoints).

**Current Behavior**

There is no way to deactivate certain flows (or endpoints)

**Context**

Specifically I want to deactivate the "device authorization flow". I don't want my server to have the ability.

The "OpenID Connect 1.0 Client Registration endpoint" is disabled by default. Why can't I deactive other endpoints?

jgrandja:
@ddittmar 

> I want to deactivate the "device authorization flow". I don't want my server to have the ability.

Can you please provide a detailed reason why you need to deactivate the device authorization endpoint?

As an FYI, if there are no `RegisterClient`'s that are configured with `.authorizationGrantType(AuthorizationGrantType.DEVICE_CODE)` then the endpoint will never process the flow anyway.

> I would like to have the ability to deactivate certain flows (or endpoints).

Which other flows? Please be specific and provide a detailed reason.

ddittmar:
@jgrandja sorry for the late reply ... 

You're right that the flow is never processed if the `RegisterdClient` does not have the `AuthorizationGrantType.DEVICE_CODE` configured... but then the endpoint is open and will never do anything... additionally the filter is in the securty chain without any need...

I just thought it would be good to deactive the flow completly because (in my case) there is no need for a "device authorization flow" and I think (especally) the device flow is a more exotic case.

jgrandja:
Thanks for the explanation @ddittmar. I updated the issue title to be more specific.

gregecho:
@jgrandja ,

Can I work on this enhancement?

jgrandja:
@gregecho Thanks for your interest. It's a lower priority item at the moment but if you would like to work on it please go ahead. Just a heads up that I'm heading out for the holidays and will be back Jan 8.

jgrandja:
@ddittmar Just circling back to this and I have a temporary workaround until we come up with a solution.

Assuming the `@Bean` name for the Authorization Server `SecurityFilterChain` is `authorizationServerSecurityFilterChain`, registering the following `BeanPostProcessor` will remove the `device_code` grant `Filter`'s:

```java
@Bean
public BeanPostProcessor authorizationServerSecurityFilterChainPostProcessor() {
	return new BeanPostProcessor() {
		public Object postProcessAfterInitialization(Object bean, String beanName) throws BeansException {
[...]

PR #1498, opened 2024-01-08
Disable device authorization flow by default

gregecho:
Fix gh-1454, Disable device authorization grant by default. Such as, deviceAuthorizationEndpoint and deviceVerificationEndpoint

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
@@ -403,7 +403,6 @@ public final class OAuth2AuthorizationServerConfigurer
 		configurers.put(OAuth2TokenEndpointConfigurer.class, new OAuth2TokenEndpointConfigurer(this::postProcess));
 		configurers.put(OAuth2TokenIntrospectionEndpointConfigurer.class, new OAuth2TokenIntrospectionEndpointConfigurer(this::postProcess));
 		configurers.put(OAuth2TokenRevocationEndpointConfigurer.class, new OAuth2TokenRevocationEndpointConfigurer(this::postProcess));
-		//configurers.put(OAuth2DeviceVerificationEndpointConfigurer.class, new OAuth2DeviceVerificationEndpointConfigurer(this::postProcess));
 		return configurers;
 	}
 
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
@@ -222,40 +222,26 @@ public final class OAuth2AuthorizationServerConfigurer
 	}
 
 	/**
-	 * Configures the OAuth 2.0 Device Authorization Endpoint (disabled by default).
+	 * Configures the OAuth 2.0 Device Authorization Endpoint.
 	 *
 	 * @param deviceAuthorizationEndpointCustomizer the {@link Customizer} providing access to the {@link OAuth2DeviceAuthorizationEndpointConfigurer}
 	 * @return the {@link OAuth2AuthorizationServerConfigurer} for further configuration
 	 * @since 1.1
 	 */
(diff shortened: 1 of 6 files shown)


gregecho:
@jgrandja ,
Having a PR for disable gh-1454, would you please help to review? Let me know if you have any questions.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
```diff
@@ -359,8 +403,6 @@ private Map<Class<? extends AbstractOAuth2Configurer>, AbstractOAuth2Configurer>
 		configurers.put(OAuth2TokenEndpointConfigurer.class, new OAuth2TokenEndpointConfigurer(this::postProcess));
 		configurers.put(OAuth2TokenIntrospectionEndpointConfigurer.class, new OAuth2TokenIntrospectionEndpointConfigurer(this::postProcess));
 		configurers.put(OAuth2TokenRevocationEndpointConfigurer.class, new OAuth2TokenRevocationEndpointConfigurer(this::postProcess));
-		configurers.put(OAuth2DeviceAuthorizationEndpointConfigurer.class, new OAuth2DeviceAuthorizationEndpointConfigurer(this::postProcess));
```
jgrandja, line 362: This change breaks (backward) compatibility since it's not disabled by default. Instead of disabling by default, there needs to be a configurable way of disabling the feature.

gregecho: Make sense.

gregecho: Do we need to provider a configuration(e.g. disableDeviceAuthorizationFlow) and default value is true? Any setting we can refer to?

gregecho: @jgrandja ，
Adding a configuration `deviceAuthorizationEndpoint.enableDeviceAuthorizationEndpoint(true);` for  enable/disable device authorization flow. The default value of `deviceAuthorizationEndpoint.enableDeviceAuthorizationEndpoint ` is true which enable by default for backward compatibility.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2DeviceAuthorizationEndpointConfigurer.java
```diff
@@ -161,6 +167,11 @@ public OAuth2DeviceAuthorizationEndpointConfigurer verificationUri(String verifi
 		return this;
 	}
 
+	public OAuth2DeviceAuthorizationEndpointConfigurer enableDeviceAuthorizationEndpoint(boolean enableDeviceAuthorizationEndpoint) {
```
jgrandja, line 170: I'm not keen on exposing it this way. As well, we would need to provide the same to `OAuth2DeviceVerificationEndpointConfigurer` to be consistent.

I also think we should come up with a solution that allows other endpoints to be disabled if needed. I don't think a "flag" setting on all the configurers makes sense.

At this point, I don't have any suggestions but I'll give it some thought. Please give it some further thought s well.

gregecho: @jgrandja , sorry about that. I didn't have a chance to discovery the alternative solution. Any ideas?

jgrandja: @gregecho I haven't had time either to come up with a solution. I'm going to close this PR for now. When you have time and come up with an alternative solution, please propose it here first and then I can either re-open this PR or you can submit a new one. Thanks.

dlehammer:
Hi guys,

Thought I would add my 2c regarding.

> .. I don't think a "flag" setting on all the configurers makes sense.

I second the notion regarding boolean flags, as the readability and granularity usually suffers greatly along this path.

Instead this got me thinking of something along the lines of an "implicit" flag.

Consider the scenario where a `..Configurer` has one or more `customizer` field(s) representing ex. the device-flow or similar chunks of functionality 🤔 

Then the value `null` could be considered disabled ~ `false`, and if populated the configuration is applied as-is.

(I also suspect this approach would enable a path for aligning the `.well-known` content with the runtime configuration)

Issue #1475, opened 2023-12-14
Minimal configuration doesn't work with Spring Boot 3.2.0


Corke123:
**Describe the bug**

Setting up Spring Authorization Server with minimal configuration as per documentation from [here](https://docs.spring.io/spring-boot/docs/current/reference/html/web.html#web.security.oauth2.authorization-server) doesn't work with Spring OAuth Client. 
The login request fails with message: 

> No AuthenticationProvider found for org.springframework.security.authentication.UsernamePasswordAuthenticationToken

This works fine with Spring Boot version 3.1.3.

Stacktrace:
```
org.springframework.security.authentication.ProviderNotFoundException: No AuthenticationProvider found for org.springframework.security.authentication.UsernamePasswordAuthenticationToken
	at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:234) ~[spring-security-core-6.2.0.jar:6.2.0]
	at org.springframework.security.web.authentication.UsernamePasswordAuthenticationFilter.attemptAuthentication(UsernamePasswordAuthenticationFilter.java:85) ~[spring-security-web-6.2.0.jar:6.2.0]
	at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:231) ~[spring-security-web-6.2.0.jar:6.2.0]
	at org.springframework.security.web.authentication.AbstractAuthenticationProcessingFilter.doFilter(AbstractAuthenticationProcessingFilter.java:221) ~[spring-security-web-6.2.0.jar:6.2.0]
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:374) ~[spring-security-web-6.2.0.jar:6.2.0]
	at org.springframework.security.web.authentication.logout.LogoutFilter.doFilter(LogoutFilter.java:107) ~[spring-security-web-6.2.0.jar:6.2.0]
	at org.springframework.security.web.authentication.logout.LogoutFilter.doFilter(LogoutFilter.java:93) ~[spring-security-web-6.2.0.jar:6.2.0]
	at org.springframework.security.web.FilterChainProxy$VirtualFilterChain.doFilter(FilterChainProxy.java:374) ~[spring-security-web-6.2.0.jar:6.2.0]
[...]

Issue #1186, opened 2023-04-25
ref-doc: Update Getting Started with Spring Boot starter


sjohnr:
In preparation for the 1.1 release, the Getting Started page can be updated to use `spring-boot-starter-oauth2-authorization-server`.

Related gh-673

PR #1519, opened 2024-01-19
Update application.yml in Getting Started guide

rozagerardo:
Update `application.yml` in Getting Started guide so it aligns with the `UserDetailsService` `@Bean` defined in the _"Defining Required Components"_ section.

Issue gh-1186

### DIFF
--- a/docs/src/main/java/sample/gettingstarted/application.yml
+++ b/docs/src/main/java/sample/gettingstarted/application.yml
@@ -7,6 +7,9 @@ logging:
 
 spring:
   security:
+    user:
+      name: user
+      password: password
     oauth2:
       authorizationserver:
         client:


jgrandja:
Thanks for the PR @rozagerardo.

Just a heads up that I updated the PR description and associated commit message as the original description stated that this fixed gh-1475, which is not the case. We'll need to wait until Boot `3.2.2` is released to validate the fix.

rozagerardo:
Cool @jgrandja. A quick question, though, if this doesn't "fix" https://github.com/spring-projects/spring-authorization-server/issues/1475, then how will we resolve that issue?

Anyway, Boot `3.2.2` is GA now. The minimal configuration seems to be working with this version.

Here is a small sample to check it out (I included an OAuth2 Client and a RS to check the whole flow, just in case):

https://github.com/rozagerardo/samples/tree/spring/auth-server-minimal-config

Also, as I was working on the setup, I realized we're using the `{noop}` prefix for the `client-secret`, and I was wondering if we should follow the same approach for the user `password` we're including at this stage, mainly for consistency (it works either way). Let me know if I should modify/re-create the PR with that small change :)

jgrandja:
@rozagerardo 

> A quick question, though, if this doesn't "fix" https://github.com/spring-projects/spring-authorization-server/issues/1475, then how will we resolve that issue?

See this [comment](https://github.com/spring-projects/spring-authorization-server/issues/1475#issuecomment-1916822828)

jgrandja:
@rozagerardo 

> I was wondering if we should follow the same approach for the user `password` we're including at this stage, mainly for consistency (it works either way)

I think we are fine to leave as-is. FYI, `UserDetailsServiceAutoConfiguration` automatically prefixes the password with `{noop}`

PR #1525, opened 2024-01-26
Implement OAuth 2.0 Token Exchange

sjohnr:


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
+		}
+
+		Authentication subjectPrincipal = subjectAuthorization.getAttribute(Principal.class.getName());
+		if (subjectPrincipal == null) {
+			// As per https://datatracker.ietf.org/doc/html/rfc8693#section-1.1
+			// We require a principal to be available via the subject_token for
+			// impersonation or delegation use cases.
+			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
+		}
```
sjohnr, line 181: Based on "A.1: Impersonation Token Exchange Example", the `sub` of the issued `access_token` should be the same `sub` of the `subject_token`. The way the `JwtGenerator` currently works requires a`principal` for this, so this validation ensures we have a user principal.

This would mean that `client_credentials` is not a valid grant type for the `subject_token`, because there is no user principal for that grant type. Only tokens originally issued with `authorization_code` or `device_code` would be allowed.

I wonder if there are use cases for exchanging tokens issued with the `client_credentials` grant? If so, we would have to look up the associated `RegisteredClient` and construct some kind of `Authentication` to represent that original client's principal. Seems like something we would wait to implement?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+			}
+
+			OAuth2Authorization.Token<OAuth2Token> actorToken = actorAuthorization.getToken(
+					tokenExchangeAuthentication.getActorToken());
+			if (!actorToken.isActive()) {
+				// As per https://tools.ietf.org/html/rfc6749#section-5.2
+				// invalid_grant: The provided authorization grant (e.g., authorization code,
+				// resource owner credentials) or refresh token is invalid, expired, revoked [...].
+				throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
+			}
```
sjohnr, line 203: We also look up an existing authorization by the `actor_token` (if provided). Same validation as above would apply.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+
+			actorPrincipal = actorAuthorization.getAttribute(Principal.class.getName());
+			if (actorPrincipal == null) {
+				// As per https://datatracker.ietf.org/doc/html/rfc8693#section-1.1
+				// We require a principal to be available via the actor_token for
+				// the delegation use case.
+				throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
+			}
```
sjohnr, line 211: The actor would also require a user principal, and the same discussion as above applies.

Since the requesting client can authenticate using their own credentials, the `actor_token` would represent a separate identity from the client, correct? In other words, delegation in this context is one user delegating to another user, not a user delegating to a client (which is normal OAuth2).

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+		}
+
+		if (this.logger.isTraceEnabled()) {
+			this.logger.trace("Validated token request parameters");
+		}
+
+		if (CollectionUtils.isEmpty(authorizedScopes)) {
+			// TODO: Validate that original authorizedScopes are also configured for this client?
+			authorizedScopes = new LinkedHashSet<>(subjectAuthorization.getAuthorizedScopes());
+		}
```
sjohnr, line 231: Scopes are optional in the request. In A.1 the scopes are automatically carried over from `subject_token` to issued `access_token`. Do we need to validate that the scopes are allowed by the requesting client to carry them over in this case?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+				.authorizedScopes(authorizedScopes)
+				.tokenType(OAuth2TokenType.ACCESS_TOKEN)
+				.authorizationGrantType(TOKEN_EXCHANGE)
+				.authorizationGrant(tokenExchangeAuthentication);
+		// @formatter:on
+		if (actorPrincipal != null) {
+			// NOTE: Passing attributes requires corresponding changes in JwtGenerator.
+			// TODO: Consider a "composite" principal combining the subject and actor principals?
+			tokenContextBuilder.put(Authentication.class.getName().concat(".ACTOR"), actorPrincipal);
+		}
```
sjohnr, line 248: I'm trying to work out a way to pass the actor (if any) so it can be used by the `TokenGenerator`. There's a fairly tight coupling between the extra information passed by this `AuthenticationProvider` and the additional claims that should end up in the `access_token`.

There's already a bit of specific code for certain token types in the `JwtGenerator`, but do we want grant type specific functionality in there?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+				generatedAccessToken.getExpiresAt(), tokenContext.getAuthorizedScopes());
+
+		// @formatter:off
+		OAuth2Authorization.Builder authorizationBuilder = OAuth2Authorization.withRegisteredClient(registeredClient)
+				.principalName(subjectAuthorization.getPrincipalName())
+				.authorizationGrantType(TOKEN_EXCHANGE)
+				.authorizedScopes(authorizedScopes)
+				.attribute(SUBJECT_AUTHORIZATION_ID, subjectAuthorization.getId())
+				.attribute(SUBJECT_TOKEN_TYPE, tokenExchangeAuthentication.getSubjectTokenType())
+				.attribute(Principal.class.getName(), subjectPrincipal);
```
sjohnr, line 274: I feel we need some way to store information about the original subject. I'm open to suggestions for a preferred way to store this information or if this ok for now. The principal is important though, as we need to carry that forward if this issued token is later exchanged for yet another token.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+		// @formatter:on
+		if (actorAuthorization != null) {
+			authorizationBuilder.attribute(ACTOR_AUTHORIZATION_ID, actorAuthorization.getId());
+			authorizationBuilder.attribute(ACTOR_TOKEN_TYPE, tokenExchangeAuthentication.getActorTokenType());
```
sjohnr, line 278: We need some way to convey information about the actor. I'm open to suggestions for a preferred way to store this information. I believe we'll need it to be able to construct a chain of actors for the `act` claim (not yet implemented in this PR).

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+
+		Map<String, Object> additionalParameters = new HashMap<>();
+		additionalParameters.put(ISSUED_TOKEN_TYPE, tokenExchangeAuthentication.getRequestedTokenType());
+
+		if (this.logger.isTraceEnabled()) {
+			this.logger.trace("Authenticated token request");
+		}
+
+		return new OAuth2AccessTokenAuthenticationToken(
+				registeredClient, clientPrincipal, accessToken, null, additionalParameters);
```
sjohnr, line 305: We never issue a `refresh_token` for now, hence `null` here.

sjohnr (commented):
@jgrandja I've opened this as a draft to gather feedback about specific things in the `AuthenticationProvider`, and the `AuthenticationConverter` and `Authentication` are implemented as well. I wanted to get feedback at this point before I go much further to avoid any rework, as the rest of the implementation would build on what I have so far.

So far, you can make a valid token exchange request with a subject or subject+actor, get an access token in response, and see an authorization in the DB with attributes. The new/updated token claims have not been implemented yet.

You can ignore the rest of the PR for now as it's still very much work-in-progress and local testing.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
```diff
@@ -0,0 +1,313 @@
...
+		}
+
+		OAuth2Authorization.Token<OAuth2Token> subjectToken = subjectAuthorization.getToken(
+				tokenExchangeAuthentication.getSubjectToken());
+		if (!subjectToken.isActive()) {
+			// As per https://tools.ietf.org/html/rfc6749#section-5.2
+			// invalid_grant: The provided authorization grant (e.g., authorization code,
+			// resource owner credentials) or refresh token is invalid, expired, revoked [...].
+			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
+		}
```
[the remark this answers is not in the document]
jgrandja: Yes, let's validate to ensure the correctness of the request parameters

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
jgrandja: > I wonder if there are use cases for exchanging tokens issued with the `client_credentials` grant?

Most definitely there are use cases. For example, a backend-service obtains a token using `client_credentials` grant and then calls a downstream service. The downstream service then might do a token exchange before it calls the next downstream service, but with downgraded scopes.

We need to account for this flow.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
jgrandja: Yes same validation.

But when I look at `actorAuthorization.getAttribute(Principal.class.getName())`, this will be an issue for a `client_credentials` authorization, since it would be `null`. We still can get the `Principal` using `client_id`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
jgrandja: > Do we need to validate that the scopes are allowed by the requesting client to carry them over in this case?

I think we do. Moreover, I'm not sure it's a good idea to carry-over the same scopes as-is as this may turn out to be a vulnerability through privilege escalation. My gut feeling is that we don't carry it over but we need to decide the most secure approach to take.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
jgrandja: `SessionInformation` is passed in via the `OAuth2TokenContext`. Can we do something similar?

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
jgrandja: I'm not sure if we should store each item in their own attribute. Maybe we create a simple "holder" class and it's composed of all the items, which is then stored in an attribute of the authorization.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
jgrandja: See [comment](https://github.com/spring-projects/spring-authorization-server/pull/1525#discussion_r1471820793)

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
sjohnr: > We still can get the Principal using client_id.

Do you mean instantiate a `OAuth2ClientAuthenticationToken`?

jgrandja: No, because `OAuth2ClientAuthenticationToken` is strictly used to represent an authenticated client principal. I think this point is mute anyway based on our discussion of not supporting subject tokens that were obtained using `client_credentials`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenExchangeAuthenticationProvider.java
sjohnr: For now, I've added validation for carried-over-scopes.

Milesy:
@sjohnr - Is it possible to get some very basic documentation so I can have an experiment with the new code locally until the release is ready?

sjohnr:
@Milesy I plan to add documentation into RC1. In the meantime, you can clone this project and run the sample applications to see token exchange in action to learn how it works and inspect the code/configuration to learn how to configure it. Even after the reference documentation is in place, the [demo-authorizationserver sample](https://github.com/spring-projects/spring-authorization-server/tree/main/samples/demo-authorizationserver/src/main/java/sample) is still one of the best places to learn about supported features.

You simply need to [add the grant type](https://github.com/spring-projects/spring-authorization-server/blob/d1515680f293f98ab768367f484168c67b718525/samples/demo-authorizationserver/src/main/java/sample/config/AuthorizationServerConfig.java#L165) for the registered client. You can also take a look at a [new token customizer](https://github.com/spring-projects/spring-authorization-server/blob/d1515680f293f98ab768367f484168c67b718525/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2TokenExchangeTokenCustomizers.java#L55-L83) specific to token-exchange. If you need to customize anything about the token beyond the basics, you can [register a token customizer](https://docs.spring.io/spring-authorization-server/reference/core-model-components.html#oauth2-token-customizer) which gives access to the entire token request through `OAuth2TokenExchangeAuthenticationToken`.

Milesy:
@sjohnr thank you - and apologies for questions here I am not sure where else I can add them given how new this stuff is. I have got everything configured but I am failing on:

`static OAuth2ClientAuthenticationToken getAuthenticatedClientElseThrowInvalidClient(Authentication authentication) {
        OAuth2ClientAuthenticationToken clientPrincipal = null;
        if (OAuth2ClientAuthenticationToken.class.isAssignableFrom(authentication.getPrincipal().getClass())) {
            clientPrincipal = (OAuth2ClientAuthenticationToken)authentication.getPrincipal();
        }

        if (clientPrincipal != null && clientPrincipal.isAuthenticated()) {
            return clientPrincipal;
        } else {
            throw new OAuth2AuthenticationException("invalid_client");
        }
    }`

I have configured the resource server with a JWT decoder, which is the authentication token used within our corporate network. But it means the class type is:

class org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken

And the method throws invalid_client back out of the request.

Milesy:
@sjohnr - in my use case, the token the user has is not issued by this server. The problem I am trying to solve is where we have multiple IDPs coming into our gateway, and we want to exchange that token for a common generic token that can then be used internally. So both the Authorization: header and the subject_token is the same. It is not possible to use any sort of password or client secret based authentication.

sjohnr:
@Milesy this issue tracker is not a support forum, so I would ask you to please direct specific questions about your implementation to [Stack Overflow](https://stackoverflow.com/questions/tagged/spring-authorization-server). Having said that, this token exchange implementation is not a general purpose Security Token Service (STS) and is not designed to exchange arbitrary tokens, so the error you are receiving is expected.

Issue #1019, opened 2022-12-21
How-to: Implement core services with JSON-backed store


jgrandja:
We should provide a guide that demonstrates how to implement `RegisteredClientRepository`, `OAuth2AuthorizationService` and `OAuth2AuthorizationConsentService` using a JSON-backed store (e.g. Redis).

Related gh-558

wapkch:
Since I'm doing migrations from Spring Security OAuth to Spring Authorization Server, I try to implement a RedisOAuth2AuthorizationService.  The demo can be found [here](https://github.com/wapkch/spring-authorization-server-demo), tests can be run with [AuthorizationServerTests](https://github.com/wapkch/spring-authorization-server-demo/blob/main/src/test/java/sample/AuthorizationServerTests.java). I haven't used this in production yet, so just as a sample for guys who need it urgently. 

```
public class OAuth2AuthorizationDeserializer extends JsonDeserializer<OAuth2Authorization> {

    private final RegisteredClientRepository registeredClientRepository;

    private static final TypeReference<Set<String>> SET_TYPE_REFERENCE = new TypeReference<>() {
    };

    private static final TypeReference<Map<String, Object>> MAP_TYPE_REFERENCE = new TypeReference<>() {
    };

    private static final TypeReference<AuthorizationGrantType> GRANT_TYPE_TYPE_REFERENCE = new TypeReference<>() {
    };

    public OAuth2AuthorizationDeserializer(RegisteredClientRepository registeredClientRepository) {
        this.registeredClientRepository = registeredClientRepository;
    }

    @Override
    public OAuth2Authorization deserialize(JsonParser jp, DeserializationContext ctxt) throws IOException {
        ObjectMapper mapper = (ObjectMapper) jp.getCodec();
        JsonNode jsonNode = mapper.readTree(jp);

        Set<String> authorizedScopes = mapper.convertValue(jsonNode.get("authorizedScopes"), SET_TYPE_REFERENCE);
        Map<String, Object> attributes = mapper.convertValue(jsonNode.get("attributes"), MAP_TYPE_REFERENCE);
        Map<String, Object> tokens = mapper.convertValue(jsonNode.get("tokens"), MAP_TYPE_REFERENCE);
        AuthorizationGrantType grantType = mapper.convertValue(jsonNode.get("authorizationGrantType"), GRANT_TYPE_TYPE_REFERENCE);

        String id = readJsonNode(jsonNode, "id").asText();
[...]

PR #1527, opened 2024-01-29
Revelation Builder constructor

leshalv:


leshalv:
Deserialize and assemble OAuth2Authorization in a user-defined mode

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2Authorization.java
```diff
@@ -391,7 +391,7 @@ public static class Builder implements Serializable {
 		private Map<Class<? extends OAuth2Token>, Token<?>> tokens = new HashMap<>();
 		private final Map<String, Object> attributes = new HashMap<>();
 
-		protected Builder(String registeredClientId) {
+		public Builder(String registeredClientId) {
```
jgrandja, line 394: @leshalv 

The `Builder` is not intended to be directly instantiated. The factory methods `withRegisteredClient()` and `from()` should be used instead.

> Deserialize and assemble OAuth2Authorization in a user-defined mode

I don't quite understand this statement. Please provide more specific details.

leshalv: When customizing redis storage, deserialization is required. RegisteredClient is required to build an OAuth2Authorization object, which requires encapsulation of RegisteredClient. In fact, when building OAuth2Authorization, only registeredClientId is enough.

jgrandja: @leshalv `OAuth2Authorization` and `RegisteredClient` are core domain classes within the framework and are not intended to represent an Entity or DTO, that can more easily be serialized. If you're looking to serialize these objects into a Redis store then you need to create an Entity/DTO representation of those objects so it can be serialized.

There is a discussion and some examples provided in gh-558.

We are also planning on providing a How-to guide in gh-1019 to demonstrate how to implement it. You can also look at the existing [How-to: Implement core services with JPA](https://docs.spring.io/spring-authorization-server/reference/guides/how-to-jpa.html) to see how the Entity representations can look like and instead of persisting to a sql db, you can serialize to JSON for Redis storage.

I'm going to close this PR as we don't want to expose the `Builder` constructor as the factory methods should be used instead.

Issue #850, opened 2022-08-10
Check client secret not expired in ClientSecretAuthenticationProvider


doctormacky:
**Describe the bug**
When check the client id and secret which load from database, the jwt code are always be generated even the client id is expred.

**To Reproduce**
set a expiration date for client id and secret, then try to call /oauth2/token with the predefined client id and secret. 

**Expected behavior**
return exception like 'the client id is expired'

**Sample**

A link to a GitHub repository with a [minimal, reproducible sample](https://stackoverflow.com/help/minimal-reproducible-example).

Reports that include a sample will take priority over reports that do not.
At times, we may require a sample, so it is good to try and include a sample up front.

I know we can customize the ClientSecretAuthenticationProvider by ourselves, but as a build-in/default one, We should check this attribute (expirate date) by default.

jgrandja:
Thanks for the request @doctormacky. This would be a good enhancement. Are you interested in submitting a PR?

doctormacky:
@jgrandja  I am glad to help on that,  which branch should I work on? I noted that this project is a crazy one and moving very fast.

jgrandja:
@doctormacky Please base the PR on `0.4.x`. Thank you!

doctormacky:
@jgrandja I already submit a PR, please review [it](https://github.com/spring-projects/spring-authorization-server/pull/862). thanks.

PR #1532, opened 2024-02-05
fix a bug

ansatsing:
check the client expiration time in advance to prevent endless redirection jumps


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java
```diff
@@ -109,6 +111,12 @@ public Authentication authenticate(Authentication authentication) throws Authent
 					authorizationCodeRequestAuthentication, null);
 		}
 
+		if (registeredClient.getClientSecretExpiresAt() != null &&
```
jgrandja, line 114: @ansatsing This check has already been implemented in gh-850

ansatsing: I know the checkpoint you mentioned. But the earlier the system checks, the less meaningless processing it will do.

PR #1545, opened 2024-02-20
Explain the importance of requireProofKey

weltonrodrigo:
The importance of making PKCE required is explained at https://www.ietf.org/archive/id/draft-ietf-oauth-security-topics-22.html#name-pkce-downgrade-attack


jgrandja:
Thanks for the update @weltonrodrigo. This is now merged.

PR #1551, opened 2024-03-01
:bug: Configures OAuth 2.0 Resource Server support.

xuxiaowei-com-cn:


samples/demo-authorizationserver/src/main/java/sample/config/AuthorizationServerConfig.java
```diff
@@ -121,9 +121,7 @@ public SecurityFilterChain authorizationServerSecurityFilterChain(
 					new LoginUrlAuthenticationEntryPoint("/login"),
 					new MediaTypeRequestMatcher(MediaType.TEXT_HTML)
 				)
-			)
-			.oauth2ResourceServer(oauth2ResourceServer ->
```
jgrandja, line 125: This should not be removed as `oauth2ResourceServer()` should be configured with the `authorizationServerSecurityFilterChain()` NOT the `defaultSecurityFilterChain()`.

Issue #1541, opened 2024-02-20
Provide more flexibility on when to display consent page


petrdvorak:
**Expected Behavior**

When logging in via OAuth 2.1 dance, the consent is only requested once. For more security/compliance-sensitive applications, it would be nice to have a simple way to enforce the consent screen on every login so that the consent is re-granted.

We would like to be able to configure, i.e.:

```java
final RegisteredClient oidcClient = RegisteredClient.withId(UUID.randomUUID().toString())
                // ... 
                // Consent.ALWAYS / Consent.ONCE / Consent.NEVER
                .clientSettings(ClientSettings.builder().requireAuthorizationConsent(Consent.ALWAYS).build())
                .build();
```

**Current Behavior**

Once consent is granted to given scopes, the consent screen is not shown.

We are able to configure:

```java
final RegisteredClient oidcClient = RegisteredClient.withId(UUID.randomUUID().toString())
                // ...
                .clientSettings(ClientSettings.builder().requireAuthorizationConsent(true).build())
                .build();
```

**Context**

We are mostly trying to have a consistent user flow, where the sequence of screens is the same for each login. The additional reason is that we would like the user to re-confirm the consent granted to the third-party app.

Related gh-1363

MrJovanovic13:
I've been eager to contribute for some time now and this looks like something I can tackle.

@jgrandja If this is something that you want to support, I will gladly work on it.

jgrandja:
@petrdvorak 

> When logging in via OAuth 2.1 dance, the consent is only requested once. For more security/compliance-sensitive applications, it would be nice to have a simple way to enforce the consent screen on every login so that the consent is re-granted.

I believe you are looking for support for the OIDC parameter `prompt=consent`. Please see this [comment](https://github.com/spring-projects/spring-authorization-server/issues/501#issuecomment-989172918). Can you confirm?

Do you also need similar support for the standard OAuth 2.0 `authorization_code` grant flow? Or you just need it for OIDC (login) flow?

petrdvorak:
@jgrandja This does not seem to be what I am looking for, as this seems to be the client-side control only. (... or, am I missing how this could actually ensure the consent was granted again? Removing the parameter would skip the consent screen, so I do not see why this is even in the specification...)

[...]

PR #1552, opened 2024-03-02
Provide more flexibility on when to display consent page

MrJovanovic13:
This PR fixes [Issue-1541](https://github.com/spring-projects/spring-authorization-server/issues/1541#issue-2143352464)

Some things worth mentioning.

1. The Predicate naming is too similar to the private method name, do you think we should find a different name?
2. The private method is not static anymore since I am checking the predicate inside it and therefore i had to remove the static modifier and I've moved it above the other static methods in the file.
3. I've added the logic in such a way that if predicate exists it will immediately exit according to its evaluation, without going through further checks(such as if authorization consent is enabled in client settings). I figured if someone sets the predicate they want it to have full control instead of still going through additional checks which are currently in place(scopes, previous consent etc.)

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationContext.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationContext.java
@@ -21,6 +21,8 @@ import java.util.Map;
 import java.util.function.Consumer;
 
 import org.springframework.lang.Nullable;
+import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationRequest;
+import org.springframework.security.oauth2.server.authorization.OAuth2AuthorizationConsent;
 import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
 import org.springframework.util.Assert;
 
@@ -64,6 +66,27 @@ public final class OAuth2AuthorizationCodeRequestAuthenticationContext implement
 	}
 
 	/**
+	 * Returns the {@link OAuth2AuthorizationRequest oauth2 authorization request}.
+	 *
+	 * @return the {@link OAuth2AuthorizationRequest}
+	 */
+	@Nullable
+	public OAuth2AuthorizationRequest getOAuth2AuthorizationRequest() {
+		return get(OAuth2AuthorizationRequest.class);
+	}
+
+	/**
+	 * Returns the {@link OAuth2AuthorizationConsent oauth2 authorization consent}.
+	 *
+	 * @return the {@link OAuth2AuthorizationConsent}
+	 */
+	@Nullable
+	public OAuth2AuthorizationConsent getOAuth2AuthorizationConsent() {
+		return get(OAuth2AuthorizationConsent.class);
+	}
+
+
+	/**
 	 * Constructs a new {@link Builder} with the provided {@link OAuth2AuthorizationCodeRequestAuthenticationToken}.
 	 *
 	 * @param authentication the {@link OAuth2AuthorizationCodeRequestAuthenticationToken}
@@ -90,6 +113,28 @@ public final class OAuth2AuthorizationCodeRequestAuthenticationContext implement
 		 */
 		public Builder registeredClient(RegisteredClient registeredClient) {
 			return put(RegisteredClient.class, registeredClient);
+		}
+
+		/**
+		 * Sets the {@link OAuth2AuthorizationRequest oauth2 authorization request}.
+		 *
+		 * @param authorizationRequest the {@link OAuth2AuthorizationRequest}
+		 * @return the {@link Builder} for further configuration
+		 * @since 1.3.0
+		 */
+		public Builder authorizationRequest(OAuth2AuthorizationRequest authorizationRequest) {
+			return put(OAuth2AuthorizationRequest.class, authorizationRequest);
+		}
+
+		/**
+		 * Sets the {@link OAuth2AuthorizationConsent oauth2 authorization consent}.
+		 *
+		 * @param authorizationConsent the {@link OAuth2AuthorizationConsent}
+		 * @return the {@link Builder} for further configuration
+		 * @since 1.3.0
+		 */
+		public Builder authorizationConsent(OAuth2AuthorizationConsent authorizationConsent) {
+			return put(OAuth2AuthorizationConsent.class, authorizationConsent);
 		}
 
 		/**
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java
@@ -19,6 +19,7 @@ import java.security.Principal;
 import java.util.Base64;
 import java.util.Set;
 import java.util.function.Consumer;
+import java.util.function.Predicate;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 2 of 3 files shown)


jgrandja:
Thanks @MrJovanovic13. I will review this soon.

I'm trying to finish up gh-101 before we release `1.3.0-M3`.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java
```diff
@@ -264,7 +266,49 @@ public void setAuthenticationValidator(Consumer<OAuth2AuthorizationCodeRequestAu
...
+	/**
+	 * Sets the {@link Predicate} used to determine if authorization consent is required.
+	 *
+	 * <p>
+	 * The {@link OAuth2AuthorizationCodeRequestAuthenticationContext} gives the predicate access to the {@link OAuth2AuthorizationCodeRequestAuthenticationToken},
+	 * as well as, the following context attribute:
+	 * {@link OAuth2AuthorizationCodeRequestAuthenticationContext#getRegisteredClient()} containing {@link RegisteredClient} used to make the request.
+	 *
+	 * @param requiresAuthorizationConsent the {@link Predicate} that determines if authorization consent is required.
+	 * @since 1.2.3
```
jgrandja, line 278: Update to `1.3`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java
```diff
@@ -295,26 +339,6 @@ private static OAuth2TokenContext createAuthorizationCodeTokenContext(
 		return tokenContextBuilder.build();
 	}
 
-	private static boolean requireAuthorizationConsent(RegisteredClient registeredClient,
```
jgrandja, line 298: We should keep this method and associated logic as-is but change the signature to:

```java
private boolean requireAuthorizationConsent(OAuth2AuthorizationCodeRequestAuthenticationContext context)
```

We would also need to supply `OAuth2AuthorizationCodeRequestAuthenticationContext` with `OAuth2AuthorizationRequest` and `OAuth2AuthorizationConsent`.

This method would then be the default for:

```java
private Predicate<OAuth2AuthorizationCodeRequestAuthenticationContext> requiresAuthorizationConsent;
```

Applications could then override this via `setRequiresAuthorizationConsent()`.

jgrandja (changes requested):
Thanks for the PR @MrJovanovic13.

Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java
MrJovanovic13: Sounds good. 

Would you like if `OAuth2AuthorizationRequest` and `OAuth2AuthorizationConsent` got treated the same way as `RegisteredClient` by exposing `getRegisteredClient()` and a builder method `registeredClient(RegisteredClient registeredClient)` inside the `OAuth2AuthorizationCodeRequestAuthenticationContext`? 

Or do you want just to use regular `.put()` and `.get()` for their supply & access?

MrJovanovic13: I have pushed a commit with a refactor according to your comment.

I attempted to expose getter & builders methods in the `OAuth2AuthorizationCodeRequestAuthenticationContext` for `OAuth2AuthorizationRequest` and `OAuth2AuthorizationConsent` instead of the `.put()` and `.get()` approach. However that ended up breaking a lot of tests. If It's important to go in that direction, I will put some more effort into making that work.

MrJovanovic13:
Sorry for the slight delay, had a busy week.

I can squash the commits if you guys prefer that.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationContext.java
```diff
@@ -92,6 +129,26 @@ public Builder registeredClient(RegisteredClient registeredClient) {
 			return put(RegisteredClient.class, registeredClient);
 		}
 
+		/**
+		 * Sets the {@link OAuth2AuthorizationRequest oauth2 authorization request}.
+		 *
+		 * @param oAuth2AuthorizationRequest the {@link OAuth2AuthorizationRequest}
+		 * @return the {@link Builder} for further configuration
+		 */
+		public Builder oAuth2AuthorizationRequest(OAuth2AuthorizationRequest oAuth2AuthorizationRequest) {
```
jgrandja, line 138: Rename to `authorizationRequest`

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationContext.java
```diff
@@ -92,6 +129,26 @@ public Builder registeredClient(RegisteredClient registeredClient) {
...
+			return put(OAuth2AuthorizationRequest.class, oAuth2AuthorizationRequest);
+		}
+
+		/**
+		 * Sets the {@link OAuth2AuthorizationConsent oauth2 authorization consent}.
+		 *
+		 * @param oAuth2AuthorizationConsent the {@link OAuth2AuthorizationConsent}
+		 * @return the {@link Builder} for further configuration
+		 */
+		public Builder oAuth2AuthorizationConsent(OAuth2AuthorizationConsent oAuth2AuthorizationConsent) {
```
jgrandja, line 148: Rename to `authorizationConsent`

jgrandja (changes requested):
Thanks for the updates @MrJovanovic13. Please see review comments.

Also, please squash commits. Thanks.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationContext.java
```diff
@@ -78,6 +101,20 @@ public static Builder with(OAuth2AuthorizationCodeRequestAuthenticationToken aut
...
 
+		/**
+		 * Associates an attribute.
+		 *
+		 * @param key the key for the attribute
+		 * @param value the value of the attribute
+		 * @return the {@link Builder} for further configuration
+		 */
+		@Override
+		public Builder put(Object key, Object value) {
```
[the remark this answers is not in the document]
MrJovanovic13: > This can be removed since it's inherited by `AbstractBuilder`

I overrode this method as opposed to using the one in AbstractBuilder because the one in the AbstractBuilder does not accept null values and `authorizationConsent` can be null.

The alternative is to remove the not null assertion inside the AbstractBuilder, but it's inherited 5 times so I didn't want to cause potential issues in other builders.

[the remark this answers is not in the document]
jgrandja: If `authorizationConsent` or `authorizationRequest` is `null` it should not be passed into the `builder` so it's not an issue.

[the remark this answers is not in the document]
MrJovanovic13: > If `authorizationConsent` or `authorizationRequest` is `null` it should not be passed into the `builder` so it's not an issue.

The `authorizationConsent` from `authorizationConsentService.findById()` is `@Nullable`.

So I added a null check which won't pass it into the builder if it's null. I got rid of the custom `put()` method.
Let me know if the solution is suitable.

jgrandja:
Thanks for the updates @MrJovanovic13.

FYI, I added a polish commit to get this merged before tomorrow's release.

Thanks again!

PR #1260, opened 2023-06-10
using token type hint in OAuth2TokenIntrospectionAuthenticationProvider

paz-raon:
Issue gh-1255

I hope this PR will help resolve the issue.

PR #1580, opened 2024-04-01
Using token_type_hint when find tokens in OAuth2TokenRevocationAuthenticationProvider

sangjaeoh:
Hello.

I called oauth2/revoke with a token_type_hint to revoke the token, but I found that the token type hint was not being used.

I looked through the documentation, found that access_token and refresh_token can be provided as token_type_hint.
(https://tools.ietf.org/html/rfc7009#section-2.1)

Thank you. 😃


jgrandja:
Thanks for the PR @sangjaeoh. But this is a duplicate of gh-1260. Please see this [comment](https://github.com/spring-projects/spring-authorization-server/pull/1260#issuecomment-1587720297) for details.

PR #1609, opened 2024-05-07
Update findByToken in JpaOAuth2AuthorizationService to handle exceptions

Andrew-Kang-G:
Compared to findById, which uses a primary key, findByToken has the potential to cause errors, such as returning duplicate results. Although this likelihood is low, considering many concurrent situations, it's better to handle unexpected exceptions to prevent login failures.

### DIFF
--- a/docs/src/main/java/sample/jpa/service/authorization/JpaOAuth2AuthorizationService.java
+++ b/docs/src/main/java/sample/jpa/service/authorization/JpaOAuth2AuthorizationService.java
@@ -15,26 +15,15 @@
  */
 package sample.jpa.service.authorization;
 
-import java.time.Instant;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.function.Consumer;
-
 import com.fasterxml.jackson.core.type.TypeReference;
 import com.fasterxml.jackson.databind.Module;
 import com.fasterxml.jackson.databind.ObjectMapper;
-import sample.jpa.entity.authorization.Authorization;
-import sample.jpa.repository.authorization.AuthorizationRepository;
-
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.core.log.LogMessage;
 import org.springframework.dao.DataRetrievalFailureException;
 import org.springframework.security.jackson2.SecurityJackson2Modules;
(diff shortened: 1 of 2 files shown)


docs/src/main/java/sample/jpa/service/authorization/JpaOAuth2AuthorizationService.java
```diff
@@ -90,29 +90,64 @@ public OAuth2Authorization findByToken(String token, OAuth2TokenType tokenType)
...
+	private void handleException(String token, OAuth2TokenType tokenType, Exception e) {
+
+		if (this.logger.isWarnEnabled()) {
+			this.logger.warn(LogMessage.format("Failed in running findByToken (token: '%s', error details: '%s')", token, e.getMessage()));
+		}
+
+		String value = tokenType != null ? tokenType.getValue() : null;
+
+		if (OAuth2ParameterNames.STATE.equals(value)) {
+			this.authorizationRepository.deleteAllByState(token);
```
jgrandja, line 131: @Andrew-Kang-G 

We cannot assume the error condition is directly related to "duplicate results":

> findByToken has the potential to cause errors, such as returning duplicate results

Also keep in mind that this is sample code that we have provided in a How-to guide, and the expectation is that the user will take the code and enhance it to fit the requirements for their application.

I'm going to close this as I don't think we should introduce the `delete**` operations as we want to keep the sample a simple starting point for users.

Andrew-Kang-G: Thank you for the answer. @jgrandja 

For normal operations, duplicate results don't occur. However, when considering a scenario where a user is using multiple devices, their access tokens should be managed separately. In such cases, the access token values could potentially be the same, as they are encoded blobs consisting of the user's information. This means there is a possibility of collisions. If the library's standard implementation ensures that the same token is never generated at any time, I agree. 

| principal_name | registered_client_id | access_token_value |customized_my_value sent from each mobile device|
|-------|-------|-------|-------|
| customer1 |app_customer| SameBlobRRR|mobile-Android-Chris...|
| customer1 |app_customer| SameBlobRRR|mobile-Android-Jane...|
| customer2 |app_customer| DifferentBlobRRR|mobile-Android-DDD |

Issue #1629, opened 2024-05-28
AOT hints prevent compilation when using `JdbcOAuth2AuthorizationService` or `JdbcRegisteredClientRepository`


nwholloway:
If you have a bean of type `org.springframework.security.oauth2.server.authorization.JdbcOAuth2AuthorizationService` or `org.springframework.security.oauth2.server.authorization.client.JdbcRegisteredClientRepository` then the compilation fails in `processAot` with the error:

```
java.lang.ClassNotFoundException:
  org.springframework.security.oauth2.server.authorization.jackson2.OAuth2ActorAuthenticationTokenMixin
```

To reproduce use `Spring Initializr` with dependencies "OAuth2 Authorization Server", "Spring Data JDBC" and "GraalVM Native Support", and just add the bean

```
@Bean
public JdbcRegisteredClientRepository registeredClientRepository(JdbcOperations jdbcOperations) {
    return new JdbcRegisteredClientRepository(jdbcOperations);
}
```

I have created an example [repository](https://github.com/nwholloway/spring-authorization-bug).

My investigation leads me to believe this this is caused by the AOT hint [OAuth2AuthorizationServerBeanRegistrationAotProcessor](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/aot/hint/OAuth2AuthorizationServerBeanRegistrationAotProcessor.java#L153)

The hint was introduced by 25a785de498b, but was not updated when the class was renamed by 2f1f45bc0152 (#60).

jgrandja:
Closing as duplicate of gh-1630

PR #1630, opened 2024-05-28
Fix AOT hints for OAuth 2.0 Token Exchange

nwholloway:
This is a proposed fix for issue #1629.

It just updates the classes to match the renaming performed by 2f1f45bc0152b6b8b414134cd2506ab01ece423d.


ahassanio:
I faced this issue while building the image!
what happen if i downgrade the version? and which version will be satisfied?!

nwholloway:
> I faced this issue while building the image! what happen if i downgrade the version? and which version will be satisfied?!

Spring Authorization Server 1.2.4 is not affected by this problem.  You should be able to downgrade to that version if you don't need the new features of 1.3.0.

jgrandja:
Thanks for submitting this fix @nwholloway ! This is now merged.

Issue #1636, opened 2024-06-03
Return more appropriate http status codes in OAuth2ErrorAuthenticationFailureHandler


MatthiasWinzeler:
Hi folks

Many thanks for this great project, it helps tremendously implementing a modern identity solution!

I noticed that the [OAuth2ErrorAuthenticationFailureHandler](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2ErrorAuthenticationFailureHandler.java) always [return an error 400 (bad request) to clients](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2ErrorAuthenticationFailureHandler.java#L56).

While this might be a sensible default, there are a lot of [OAuth2ErrorCodes](https://github.com/spring-projects/spring-security/blob/main/oauth2/oauth2-core/src/main/java/org/springframework/security/oauth2/core/OAuth2ErrorCodes.java) that would have more appropriate HTTP status codes, for example:

```
ACCESS_DENIED could be a 403
SERVER_ERROR could be a 500
TEMPORARILY_UNAVAILABLE could be a 503
INVALID_CLIENT could be a 401
...
```

I could not find anything related to http status codes in the OAuth2 specs, but it seems to me that the Ory Fosite project has a [good mapping of those OAuth2 error codes to http status codes](https://github.com/ory/fosite/blob/master/errors.go#L61).

It sounds like cosmetic, but we encountered real world issues where some clients need to distinguish between those errors (for example when silent refreshing in a iframe they handle 401/403s differently). Would it make sense to return the more appropriate error codes?

jgrandja:
@MatthiasWinzeler 

> Many thanks for this great project, it helps tremendously implementing a modern identity solution!

We are so happy to hear this !!! 

Regarding customizing the HTTP status, please take a look at this [comment](https://github.com/spring-projects/spring-authorization-server/pull/1465#discussion_r1462347098) for more context. Does this help?

MatthiasWinzeler:
@jgrandja many thanks - the approach in the comment looks good, we can then use this as a workaround.

However, do you think it would still make sense to return different return codes for some errors by default? i.e. the ones that are clearly not `400`s, like internal server error or the 401/403s?

jgrandja:
[...]

PR #1643, opened 2024-06-13
return more specific http status codes in OAuth2ErrorAuthenticationFailureHandler

MatthiasWinzeler:
Currently, `OAuth2ErrorAuthenticationFailureHandler` always returns `400 Bad Request` for all OAuth2 errors. There are some OAuth2 error codes that have clearly more matching http status codes, for example `"unauthorized_client"` could return a `401 Unauthorized` etc.

Adresses https://github.com/spring-projects/spring-authorization-server/issues/1636.


jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-authorization-server/issues/1636#issuecomment-2176718859)

Issue #1666, opened 2024-07-01
OAuth2AuthorizationRequestDeserializer fails on multi-valued request parameters


stefanocke:
**Describe the bug**
https://github.com/spring-projects/spring-authorization-server/issues/1250 introduced to store mutli-valued OAuth2 Authorization Request parameters as string arrays.
This seems to lead to the following exception when serializing and de-serializing the request, like done by JdbcOAuth2AuthorizationService.
```
Caused by: java.lang.IllegalArgumentException: The class with [Ljava.lang.String; and name of [Ljava.lang.String; is not in the allowlist. If you believe this class is safe to deserialize, please provide an explicit mapping using Jackson annotations or by providing a Mixin. If the serialization is only done by a trusted source, you can also enable default typing. See https://github.com/spring-projects/spring-security/issues/4370 for details
	at org.springframework.security.jackson2.SecurityJackson2Modules$AllowlistTypeIdResolver.typeFromId(SecurityJackson2Modules.java:285)
	at com.fasterxml.jackson.databind.jsontype.impl.TypeDeserializerBase._findDeserializer(TypeDeserializerBase.java:159)
	at com.fasterxml.jackson.databind.jsontype.impl.AsArrayTypeDeserializer._deserialize(AsArrayTypeDeserializer.java:97)
	at com.fasterxml.jackson.databind.jsontype.impl.AsArrayTypeDeserializer.deserializeTypedFromArray(AsArrayTypeDeserializer.java:53)
	at com.fasterxml.jackson.databind.jsontype.impl.AsPropertyTypeDeserializer.deserializeTypedFromAny(AsPropertyTypeDeserializer.java:211)
	at com.fasterxml.jackson.databind.deser.std.UntypedObjectDeserializerNR.deserializeWithType(UntypedObjectDeserializerNR.java:115)
	at com.fasterxml.jackson.databind.deser.impl.TypeWrappedDeserializer.deserialize(TypeWrappedDeserializer.java:74)
	at com.fasterxml.jackson.databind.deser.DefaultDeserializationContext.readRootValue(DefaultDeserializationContext.java:323)
	at com.fasterxml.jackson.databind.ObjectMapper._readValue(ObjectMapper.java:4706)
	at com.fasterxml.jackson.databind.ObjectMapper.readValue(ObjectMapper.java:2879)
[...]

PR #1676, opened 2024-07-22
Fix json deserialization of multi-valued authorization request parameters

MrJovanovic13:
I have tested this mixin on a sample spring authorization server app, by registering the mixin on a custom JdbcOAuth2AuthorizationService bean. 

Authorization flow now works correctly without throwing on multi valued parameters. String[] is correctly serialized & deserialized.

Closes gh-1666


jgrandja:
Thanks for the fix @MrJovanovic13 ! This is now merged.

Issue #1671, opened 2024-07-12
Empty auth_code parameter results in 500 on /token


aijazkeerio:
**Describe the bug**
Empty auth_code parameter results in 500 on /token
This is caused by https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java#L141

```
	private static boolean authorizationCodeGrant(Map<String, Object> parameters) {
		// @formatter:off
		return AuthorizationGrantType.AUTHORIZATION_CODE.getValue().equals(
				parameters.get(OAuth2ParameterNames.GRANT_TYPE)) &&
				parameters.get(OAuth2ParameterNames.CODE) != null;
		// @formatter:on
	}
``` 
returns true even though code has no value.

**To Reproduce**
Create a request for /token and add param `code` with no value/string

**Expected behavior**
Empty auth_code parameter results in 400 on /token

**Sample**
At the point of https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java#L141 
it should check if param has some value for example
```	
private static boolean authorizationCodeGrant(Map<String, Object> parameters) {
		// @formatter:off
		return AuthorizationGrantType.AUTHORIZATION_CODE.getValue().equals(
				parameters.get(OAuth2ParameterNames.GRANT_TYPE)) &&
				parameters.get(OAuth2ParameterNames.CODE) != null &&
				!parameters.get(OAuth2ParameterNames.CODE).toString().isBlank();
		// @formatter:on
	}
```

jgrandja:
Thanks for reporting this @aijazkeerio. I've confirmed this is a bug.

Would you be interested in submitting a fix for this?

jgrandja:
Closing in favour of gh-1680

PR #1680, opened 2024-07-26
Fix empty code parameter in CodeVerifierAuthenticator

aijazkeerio:
Fixes https://github.com/spring-projects/spring-authorization-server/issues/1671

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
@@ -140,7 +140,8 @@ final class CodeVerifierAuthenticator {
 		// @formatter:off
 		return AuthorizationGrantType.AUTHORIZATION_CODE.getValue().equals(
 				parameters.get(OAuth2ParameterNames.GRANT_TYPE)) &&
-				parameters.get(OAuth2ParameterNames.CODE) != null;
+				parameters.get(OAuth2ParameterNames.CODE) != null &&
+				!parameters.get(OAuth2ParameterNames.CODE).toString().isBlank();
 		// @formatter:on
 	}
 
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
@@ -137,12 +137,16 @@ final class CodeVerifierAuthenticator {
 	}
 
 	private static boolean authorizationCodeGrant(Map<String, Object> parameters) {
-		// @formatter:off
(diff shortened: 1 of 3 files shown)


9527286768 (approved):

9527286768 (approved):

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
```diff
@@ -140,7 +140,8 @@ private static boolean authorizationCodeGrant(Map<String, Object> parameters) {
 		// @formatter:off
 		return AuthorizationGrantType.AUTHORIZATION_CODE.getValue().equals(
 				parameters.get(OAuth2ParameterNames.GRANT_TYPE)) &&
-				parameters.get(OAuth2ParameterNames.CODE) != null;
+				parameters.get(OAuth2ParameterNames.CODE) != null &&
```
jgrandja, line 143: If the grant type is `authorization_code` and the `code` is empty, it should not return `false` and instead should error.

Try this instead:

```java
if (!AuthorizationGrantType.AUTHORIZATION_CODE.getValue().equals(parameters.get(OAuth2ParameterNames.GRANT_TYPE))) {
	return false;
}
if (!StringUtils.hasText((String) parameters.get(OAuth2ParameterNames.CODE))) {
	throwInvalidGrant(OAuth2ParameterNames.CODE);
}
return true;
```

oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/authentication/ClientSecretAuthenticationProviderTests.java
```diff
@@ -282,6 +283,29 @@ public void authenticateWhenAuthorizationCodeGrantAndValidCredentialsThenAuthent
 		assertThat(authenticationResult.getRegisteredClient()).isEqualTo(registeredClient);
 	}
 
+	@Test
+	public void authenticateWhenAuthorizationCodeGrantAndPkceAndValidCodeVerifierAndMissingCodeThenAuthenticated() {
```
jgrandja, line 287: Please remove this test and add `OAuth2AuthorizationCodeGrantTests.requestWhenPublicClientWithPkceAndEmptyCodeThenBadRequest()`. You can use `requestWhenPublicClientWithPkceThenReturnAccessTokenResponse()` as a template and adjust it.

jgrandja (changes requested):
Thanks for the PR @aijazkeerio. Please see review comments.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/CodeVerifierAuthenticator.java
aijazkeerio: Please review new changes.

jgrandja:
Thanks for the updates @aijazkeerio ! This is now merged.

FYI, I added a polish commit that was mostly formatting changes.

PR #1713, opened 2024-09-11
Update OAuth2TokenIntrospectionAuthenticationProvider.java to add username into tokenclaim

george510257:
OAuth2TokenIntrospectionAuthenticationProvider add username to tokenClaims

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
@@ -111,7 +111,7 @@ public final class OAuth2TokenIntrospectionAuthenticationProvider implements Aut
 
 		RegisteredClient authorizedClient = this.registeredClientRepository
 			.findById(authorization.getRegisteredClientId());
-		OAuth2TokenIntrospection tokenClaims = withActiveTokenClaims(authorizedToken, authorizedClient);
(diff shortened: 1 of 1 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2TokenIntrospectionAuthenticationProvider.java
```diff
@@ -141,6 +141,7 @@ private static OAuth2TokenIntrospection withActiveTokenClaims(
 		tokenClaims.clientId(authorizedClient.getClientId());
 
 		// TODO Set "username"
+		tokenClaims.username(authorization.getPrincipalName());
```
jgrandja, line 144: As per [Section 2.2](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2).  Introspection Response

> username
      OPTIONAL.  Human-readable identifier for the resource owner who
      authorized this token.

The `authorization.getPrincipalName()` might return an actual name or email address depending on the  `Authentication` instance. Furthermore, `Human-readable identifier` is quite vague because it could be a GUID or email address since they both are unique and represent identifiers.

Also, for the `client_credentials` grant, `authorization.getPrincipalName()` returns the `client_id`, which would be incorrect to set as `username`.

Lastly, since `username` is OPTIONAL, I think it's better for the application to provide a custom [OAuth2TokenCustomizer](https://docs.spring.io/spring-authorization-server/reference/core-model-components.html#oauth2-token-customizer) that adds the `username` claim, if needed.

jgrandja:
@george510257 I'm going to close this PR based on the [comment](https://github.com/spring-projects/spring-authorization-server/pull/1713#discussion_r1764709523).

Issue #1693, opened 2024-08-16
Customize `post_logout_redirect_uri` validation in OIDC RP-initiated logout


Kehrlann:
## Context

In local development scenarios, some auth servers relax some specified constraints. For example, in Tanzu, we have use-cases where we remove `redirect_uri` validation in `OAuth2AuthorizationCodeRequestAuthenticationProvider` through the `OAuth2AuthorizationCoderRequestAuthenticationProvider#setAuthenticationValidator` hook.

## Feature request

We want to turn off `post_logout_redirect_uri` validation in `OidcLogoutAuthenticationProvider`, but it is [currently hardcoded](https://github.com/spring-projects/spring-authorization-server/blob/f885df4343b4961732157238082ed5407bc5632c/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationProvider.java#L129-L133).

We would like to have a similar `#setAuthenticationValidator` in `OidcLogoutAuthenticationProvider`. We are interested in the redirect uri validation, but maybe this extends to the client identity (e.g. `aud` claim).

## Workaround

For our workaround, we have custom `OidcLogoutAuthenticationProvider` that wraps around the original, and modifies the client registration (not great). We rely on our custom `RegisteredClientRepository` and `RegisteredClient` implementations:

```java
class SsoOidcLogoutAuthenticationProvider implements AuthenticationProvider {

	private final OidcLogoutAuthenticationProvider delegate;

	private final OAuth2AuthorizationService authorizationService;

	private final SsoRegisteredClientRepository registeredClientRepository;

	public SsoOidcLogoutAuthenticationProvider(OidcLogoutAuthenticationProvider delegate,
			OAuth2AuthorizationService authorizationService, SsoRegisteredClientRepository registeredClientRepository) {
		this.delegate = delegate;
		this.authorizationService = authorizationService;
		this.registeredClientRepository = registeredClientRepository;
	}

	@Override
	public Authentication authenticate(Authentication authentication) throws AuthenticationException {
		OidcLogoutAuthenticationToken initialAuth = (OidcLogoutAuthenticationToken) authentication;

		OAuth2Authorization authorization = this.authorizationService.findByToken(initialAuth.getIdTokenHint(),
				new OAuth2TokenType(OidcParameterNames.ID_TOKEN));
		// ... handle error
		SsoRegisteredClient registeredClient = this.registeredClientRepository
[...]

PR #1723, opened 2024-09-17
Support custom validation in OidcLogoutAuthenticationProvider

Kehrlann:
- Similar to custom validation in `OAuth2AuthorizationCodeRequestAuthenticationProvider`
    - Similarly, not exposed in the Configurer
- Closes gh-1693

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationContext.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationContext.java
@@ -1,0 +1,106 @@
+/*
+ * Copyright 2020-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.security.oauth2.server.authorization.oidc.authentication;
+
+import java.util.Map;
+import java.util.function.Consumer;
+
+import org.springframework.lang.Nullable;
+import org.springframework.security.core.Authentication;
+import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthenticationContext;
+import org.springframework.security.oauth2.server.authorization.client.RegisteredClient;
+import org.springframework.util.Assert;
+
+/**
+ * An {@link OAuth2AuthenticationContext} that holds an
+ * {@link OidcLogoutAuthenticationToken} and additional information and is used when
+ * validating the OpenID Connect RP-Initiated Logout Request parameters.
+ *
+ * @author Daniel Garnier-Moiroux
+ * @since 1.4
+ * @see OAuth2AuthenticationContext
+ * @see OidcLogoutAuthenticationToken
+ * @see OidcLogoutAuthenticationProvider#setAuthenticationValidator(Consumer)
+ */
+public final class OidcLogoutAuthenticationContext implements OAuth2AuthenticationContext {
+
+	private final Map<Object, Object> context;
+
+	private OidcLogoutAuthenticationContext(Map<Object, Object> context) {
+		this.context = context;
+	}
+
+	@SuppressWarnings("unchecked")
+	@Nullable
+	@Override
+	public <V> V get(Object key) {
+		return hasKey(key) ? (V) this.context.get(key) : null;
+	}
+
+	@Override
+	public boolean hasKey(Object key) {
+		Assert.notNull(key, "key cannot be null");
+		return this.context.containsKey(key);
+	}
+
+	/**
+	 * Returns the {@link RegisteredClient registered client}.
+	 * @return the {@link RegisteredClient}
+	 */
+	public RegisteredClient getRegisteredClient() {
+		return get(RegisteredClient.class);
+	}
+
+	/**
+	 * Constructs a new {@link Builder} with the provided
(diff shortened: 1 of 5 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationContext.java
```diff
@@ -0,0 +1,142 @@
+/*
+ * Copyright 2020-2023 the original author or authors.
```
jgrandja, line 2: Update copyright year

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationProvider.java
```diff
@@ -21,6 +21,7 @@
 import java.security.Principal;
 import java.util.Base64;
```
jgrandja, line 22: Update copyright year

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationValidator.java
```diff
@@ -0,0 +1,94 @@
+/*
+ * Copyright 2020-2023 the original author or authors.
```
jgrandja, line 2: Update copyright year

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationContext.java
```diff
@@ -0,0 +1,142 @@
...
+	 */
+	public RegisteredClient getRegisteredClient() {
+		return get(RegisteredClient.class);
+	}
+
+	/**
+	 * Returns the {@link OAuth2Authorization authorization request}.
+	 * @return the {@link OAuth2Authorization}
+	 */
+	public OAuth2Authorization getAuthorizationRequest() {
```
jgrandja, line 73: This is not used so can be removed

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationValidator.java
```diff
@@ -0,0 +1,94 @@
...
+ * @see OidcLogoutAuthenticationContext
+ * @see OidcLogoutAuthenticationToken
+ * @see OidcLogoutAuthenticationProvider#setAuthenticationValidator(Consumer)
+ */
+public final class OidcLogoutAuthenticationValidator implements Consumer<OidcLogoutAuthenticationContext> {
+
+	/**
+	 * The default validator for {@link OidcIdToken#getAudience()}.
+	 */
+	public static final Consumer<OidcLogoutAuthenticationContext> DEFAULT_AUDIENCE_VALIDATOR = OidcLogoutAuthenticationValidator::validateAudience;
```
jgrandja, line 53: I think we should reset this change since this issue is to address `post_logout_redirect_uri` customization. Is this needed here?

jgrandja (changes requested):
Thanks for the PR @Kehrlann !

In addition to addressing the feedback, please add documentation to the reference similar to [Customizing Authorization Request Validation](https://docs.spring.io/spring-authorization-server/reference/protocol-endpoints.html#oauth2-authorization-endpoint-customizing-authorization-request-validation) and [Customizing Client Credentials Grant Request Validation](https://docs.spring.io/spring-authorization-server/reference/protocol-endpoints.html#oauth2-token-endpoint-customizing-client-credentials-grant-request-validation).

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationContext.java
Kehrlann: I think we should leave it here. 

Since this is the `OidcLogoutAuthenticationContext`, and the `OAuth2Authorization` is part of the the logout context and is loaded from a repository, it may prove useful for further customization checks.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationValidator.java
Kehrlann: I wanted to give more a bit flexibility in logout validation, and showcase how validators can be composed, similar to what we do in the redirect uri validation.

Happy to remove it though - In I can't think of a valid use-case where the `id_token` doesn't have the `client_id` in the `aud` claim.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationContext.java
jgrandja: >  it may prove useful for users who want to perform further checks

Possibly, but we haven't seen a use case for this as of yet and the preference is to only add public API's when there is a valid use case reported. We can add this later if users need it.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/oidc/authentication/OidcLogoutAuthenticationValidator.java
jgrandja: > I can't think of a valid use-case where the `id_token` doesn't have the `client_id` in the `aud` claim

I can't either and [the spec doesn't specify](https://openid.net/specs/openid-connect-rpinitiated-1_0.html#RPLogout) that it can be another value. Please reset this change and we can enhance later if users need it.

jgrandja:
Thanks for the updates @Kehrlann ! This is now merged.

Issue #1707, opened 2024-08-29
Support Configuring Authorization Server in HttpSecurity.with(authorizationServer(), ...)


rwinch:
**Expected Behavior**

It would be nice if all of the Authorization Server configuration options could be made/discoverable from a single entry point & on an existing `SecurityFilterChain`.

For example, it would be very nice if adding a default Authorization Server to an existing application could be done with a single invocation of the `HttpSecurity` DSL using something like this:

```java
http
    .with(authorizationServer(), Customizer.withDefaults())
    // ...
```

If a user wanted to make changes to the defaults, then they would do so using the `Customizer`. For example, if OIDC was not enabled by default, then the user would be able to use:

```java
http
    .with(authorizationServer(), authz -> authz
        .oidc(Customizer.withDefaults())
   )
   // ...
```

**NOTE**:  I'm not saying that OIDC should  or should not be a default, but I'm just using this as an example of how a user would be able to change the defaults.

Using this approach:
* Should feel familiar to Spring Security users
* It will make all of the features available via autocomplete after a single static import
* Should allow users to enable a feature (authorization server) rather than needing to know about the implementation details such as exception handling.

**Current Behavior**

Currently the [recommendation](https://docs.spring.io/spring-authorization-server/reference/getting-started.html#defining-required-components) is:

```java
@Bean 
@Order(1)
public SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http)
		throws Exception {
	OAuth2AuthorizationServerConfiguration.applyDefaultSecurity(http);
	http.getConfigurer(OAuth2AuthorizationServerConfigurer.class)
		.oidc(Customizer.withDefaults());	// Enable OpenID Connect 1.0
	http
		// Redirect to the login page when not authenticated from the
		// authorization endpoint
		.exceptionHandling((exceptions) -> exceptions
			.defaultAuthenticationEntryPointFor(
				new LoginUrlAuthenticationEntryPoint("/login"),
				new MediaTypeRequestMatcher(MediaType.TEXT_HTML)
			)
		)
		// Accept access tokens for User Info and/or Client Registration
		.oauth2ResourceServer((resourceServer) -> resourceServer
			.jwt(Customizer.withDefaults()));
		return http.build();
}
```  

The disadvantages of this approach are:

* It is less concise
* Having multiple SecurityFilterChain is less familiar and adds additional complexity to understanding/debugging
[...]

PR #1725, opened 2024-09-19
Simplify configuring authorization server using HttpSecurity.with()

jgrandja:
This PR attempts to simplify the [default configuration](https://docs.spring.io/spring-authorization-server/reference/configuration-model.html#default-configuration) for an authorization server using `HttpSecurity.with()`.

We would greatly appreciate if our community users can **provide their upvote** if you are in favour of:

* the [new proposed way](https://github.com/spring-projects/spring-authorization-server/pull/1725#issuecomment-2368101027) of configuring
* keeping with the [current way](https://github.com/spring-projects/spring-authorization-server/pull/1725#issuecomment-2368104944) of configuring (as documented in the [Getting Started](https://docs.spring.io/spring-authorization-server/reference/getting-started.html#defining-required-components))

Issue gh-1707

### DIFF
--- a/docs/modules/ROOT/pages/configuration-model.adoc
+++ b/docs/modules/ROOT/pages/configuration-model.adoc
@@ -8,9 +8,6 @@
 
 `OAuth2AuthorizationServerConfiguration` uses xref:configuration-model.adoc#customizing-the-configuration[`OAuth2AuthorizationServerConfigurer`] to apply the default configuration and registers a `SecurityFilterChain` `@Bean` composed of all the infrastructure components supporting an OAuth2 authorization server.
 
-[TIP]
-`OAuth2AuthorizationServerConfiguration.applyDefaultSecurity(HttpSecurity)` is a convenience (`static`) utility method that applies the default OAuth2 security configuration to `HttpSecurity`.
-
 The OAuth2 authorization server `SecurityFilterChain` `@Bean` is configured with the following default protocol endpoints:
 
 * xref:protocol-endpoints.adoc#oauth2-authorization-endpoint[OAuth2 Authorization endpoint]
@@ -58,11 +55,14 @@ https://openid.net/specs/openid-connect-core-1_0.html[OpenID Connect 1.0] is dis
 ----
 @Bean
 public SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http) throws Exception {
-	OAuth2AuthorizationServerConfiguration.applyDefaultSecurity(http);
-
-	http.getConfigurer(OAuth2AuthorizationServerConfigurer.class)
-		.oidc(Customizer.withDefaults());	// Initialize `OidcConfigurer`
-
+	OAuth2AuthorizationServerConfigurer authorizationServerConfigurer =
+			OAuth2AuthorizationServerConfigurer.authorizationServer();
+	http
+		.securityMatcher(authorizationServerConfigurer.getEndpointsMatcher())
+		.with(authorizationServerConfigurer, (authorizationServer) ->
+			authorizationServer
+				.oidc(Customizer.withDefaults())	// Initialize `OidcConfigurer`
+		);
 	return http.build();
 }
 ----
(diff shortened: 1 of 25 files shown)


jgrandja:
The following sample configuration is the **current way** of configuring a dedicated authorization server `SecurityFilterChain`:

```java
@Bean
@Order(1)
public SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http)
		throws Exception {

	OAuth2AuthorizationServerConfiguration.applyDefaultSecurity(http);
		.oidc(Customizer.withDefaults());	// Enable OpenID Connect 1.0

	http
		// Redirect to the login page when not authenticated from the
		// authorization endpoint
		.exceptionHandling((exceptions) -> exceptions
			.defaultAuthenticationEntryPointFor(
				new LoginUrlAuthenticationEntryPoint("/login"),
				new MediaTypeRequestMatcher(MediaType.TEXT_HTML)
			)
		)
		// Accept access tokens for User Info and/or Client Registration
		.oauth2ResourceServer((resourceServer) -> resourceServer
			.jwt(Customizer.withDefaults()));

	return http.build();
}
```

rwinch:
@jgrandja I'm confused why you added [these](https://github.com/spring-projects/spring-authorization-server/pull/1725#issuecomment-2368101027) [comments](https://github.com/spring-projects/spring-authorization-server/pull/1725#issuecomment-2368104944). In particular, I'd like to make sure that this was not intended to address my [feedback](https://github.com/spring-projects/spring-authorization-server/pull/1725#pullrequestreview-2318582531) since the feedback was requesting the documentation get updated & comments on the this pull request do not resolve that.

jgrandja:
@rwinch The reason for updating the comment in the main issue and providing a side-by-side comparison of the new vs. existing sample configuration is to ensure we are truly simplifying the configuration based on feedback from our community users.

The content updates here was triggered from this [comment](https://github.com/spring-projects/spring-authorization-server/issues/1707#issuecomment-2365139494). And then it made me realize that we should validate this proposed enhancement from the community first before we merge this.

Regardless on whether we merge this PR or not, the one thing that is missing in the reference doc is:

> Is it still recommended to use a multiple SecurityFilterChain's? If so, why (so users can know the tradeoffs of having a single or multiple)? The demonstration of the minimal configuration that is preferred (single/multiple) should be the first config users see in Getting Started.

I will be adding this content in the reference doc this week.

JBraddockm:
I just wanted to share my experience with this new configuration in case someone else lands on this page through their researches. Not sure if this is something specific to my configuration or this is indeed an expected behaviour as part of a wider change in Spring Security, but I had problem with upgrading to the new configuration style.

I have two filter chains where the first one configures the authorization server while the latter deals more specific cases such as allowing certain endpoints, and static files. Unless I put the following in the first security filter, then upgrading to the new configuration style breaks the app. Clients get access denied.

`.authorizeHttpRequests((authorize) ->
			authorize.anyRequest().authenticated()`

Only time that the new configuration style works without the above lines being present in the first security filter chain is by removing the following from the second filter chain while leaving `.anyRequest().authenticated()` in place.

`.requestMatchers(PathRequest.toStaticResources().atCommonLocations())
                    .permitAll()
                    .requestMatchers("/error", "/logout")
                    .permitAll()`

Docs and your examples here indeed include this line in the first security filter, but it could perhaps be missed by some upgrading their old configurations so I just wanted to share my experience here.

Thank you.

JBraddockm:
> I'm not completely following how the new configuration style broke your app. Can you provide your configuration before the upgrade and then after the upgrade so I can better understand?

Thank you for the response. 

The following configuration is working. However, if I remove `http.authorizeHttpRequests((authorize) -> authorize.anyRequest().authenticated());` from the first filter chain, then it is not working. Redirecting to the login page results in 404. There is not a explicit error message or exceptions in the auth-server despite the log level being set to trace. In the client site, I get Access Denied exception. 

```java

  @Bean
  @Order(1)
  SecurityFilterChain authorizationServerSecurityFilterChain(HttpSecurity http) throws Exception {

    OAuth2AuthorizationServerConfigurer authorizationServerConfigurer =
        OAuth2AuthorizationServerConfigurer.authorizationServer();

    http.securityMatcher(authorizationServerConfigurer.getEndpointsMatcher())
        .with(
            authorizationServerConfigurer,
            (authorizationServer) -> authorizationServer.oidc(Customizer.withDefaults()));

    http.authorizeHttpRequests((authorize) -> authorize.anyRequest().authenticated());

    http.exceptionHandling(
            (exceptions) ->
                exceptions.defaultAuthenticationEntryPointFor(
                    new LoginUrlAuthenticationEntryPoint("/login"),
                    new MediaTypeRequestMatcher(MediaType.TEXT_HTML)))
        .oauth2ResourceServer((resourceServer) -> resourceServer.jwt(Customizer.withDefaults()));

    return http.build();
  }

  @Bean
  @Order(2)
  SecurityFilterChain defaultSecurityFilterChain(HttpSecurity http) throws Exception {

    http.authorizeHttpRequests(
            (authorize) ->
                authorize
                    .requestMatchers(PathRequest.toStaticResources().atCommonLocations())
                    .permitAll()
                    .requestMatchers("/error", "/logout")
                    .permitAll()
                    .anyRequest()
                    .authenticated())
        .exceptionHandling(
            exception ->
                exception.defaultAuthenticationEntryPointFor(
                    new HxRefreshHeaderAuthenticationEntryPoint(),
                    new RequestHeaderRequestMatcher("HX-Request")))
        .formLogin(formLogin -> formLogin.loginPage("/login").permitAll())
        .oauth2Login(oauth2 -> oauth2.loginPage("/login").permitAll());
    return http.build();
  }
```
The reason I didn't have `http.authorizeHttpRequests((authorize) -> authorize.anyRequest().authenticated());`  in the first place is the examples I'd followed as I've been learning Spring Authorization Server didn't have it.  For example, [here](https://github.com/spring-projects/spring-authorization-server/blob/4cfe59cd85197fff7732bea8576d519510965eb2/samples/demo-authorizationserver/src/main/java/sample/config/AuthorizationServerConfig.java) or [here](https://github.com/rwinch/spring-enterprise-authorization-server/blob/main/authorization-server/src/main/java/com/example/authorizationserver/SecurityConfig.java). Not sure if it is meant to be there in the first place.

The reason I don't have the `/login` in the `requestMatchers` because it breaks the HxRefreshHeaderAuthenticationEntryPoint, which I don't know why yet, though adding that to the list of permittable endpoints in requestMatchers does not make any difference to the issue.

My apologies if I wasn't clearer.

jgrandja:
@JBraddockm 

> The reason I didn't have `http.authorizeHttpRequests((authorize) -> authorize.anyRequest().authenticated());`  in the first place is the examples I'd followed as I've been learning Spring Authorization Server didn't have it.  For example, [here](https://github.com/spring-projects/spring-authorization-server/blob/4cfe59cd85197fff7732bea8576d519510965eb2/samples/demo-authorizationserver/src/main/java/sample/config/AuthorizationServerConfig.java) or [here](https://github.com/rwinch/spring-enterprise-authorization-server/blob/main/authorization-server/src/main/java/com/example/authorizationserver/SecurityConfig.java). Not sure if it is meant to be there in the first place.

Both of the sample links you provided use the now deprecated `OAuth2AuthorizationServerConfiguration.applyDefaultSecurity(http)`, which applies `authorize.anyRequest().authenticated()` as shown here:

https://github.com/spring-projects/spring-authorization-server/blob/ed0265bf024c98943bbcf28ec1ca635d808a2137/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configuration/OAuth2AuthorizationServerConfiguration.java#L84

As noted in my [previous comment](https://github.com/spring-projects/spring-authorization-server/pull/1725#issuecomment-2507947571):

> `authorize.anyRequest().authenticated()` should always be applied to all filter chains in order to adhere to secure-by-default.

This is a standard best practice for all Spring Security configurations in general. Start off with securing all endpoints and then customize the authorization rules for specific endpoints depending on your security requirements.

JBraddockm:
Thank you.

PR #1728, opened 2024-09-25
Update OAuth2EndpointUtils.java

MicahZheng:
getQueryParameters only get query parameter ，
getFormParameters only get form parameter

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2EndpointUtils.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2EndpointUtils.java
@@ -15,9 +15,9 @@
  */
 package org.springframework.security.oauth2.server.authorization.web.authentication;
 
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.Map;
+import java.net.URLDecoder;
+import java.nio.charset.Charset;
+import java.util.*;
 
 import jakarta.servlet.http.HttpServletRequest;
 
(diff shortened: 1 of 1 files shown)


jgrandja:
@MicahZheng I don't understand why this change is needed?

Did you find an issue? If so, please provide a [minimal sample](https://stackoverflow.com/help/minimal-reproducible-example) or test that reproduces the issue so I can look into it further.

MicahZheng:
When the query parameters and form parameters are the same, the method getFormParameters will not get any parameters. And my partner gave me such parameters when requesting my interface, and the query parameters and form parameters are exactly the same. It does provide enough parameters. Maybe he should only put the form parameters, but I found that the method getQueryParameters and the method getFormParameters are not Just get the corresponding content like his name

MicahZheng:
POST https://example.com/oauth2/token?grant_type=authorization_code&code=the_authorization_code&redirect_uri=https%3A%2F%2Fyourapp.com%2Fcallback&client_id=client_id&client_secret=client_secret&scope=read+write HTTP/1.1
Host: example.com
Content-Type: application/x-www-form-urlencoded

grant_type=authorization_code&code=the_authorization_code&redirect_uri=https%3A%2F%2Fyourapp.com%2Fcallback&client_id=client_id&client_secret=client_secret&scope=read+write

jgrandja:
@MicahZheng I now understand your scenario.

Please review the spec for [Token Endpoint](https://www.ietf.org/archive/id/draft-ietf-oauth-v2-1-07.html#name-token-endpoint):

> The client MUST use the HTTP POST method when making access token requests.

The client should not be sending query parameters AND body parameters as this is an invalid request and should be fixed at the client.

I'm going to close this PR as `getFormParameters()` and `getQueryParameters()` is implemented as expected.

PR #1736, opened 2024-10-11
Replace DelegatingAuthenticationConverter with one in Spring Security

CrazyParanoid:
Replacing DelegatingAuthenticationConverter with a similar one from spring security.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/DelegatingAuthenticationConverter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/DelegatingAuthenticationConverter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2020-2022 the original author or authors.
+ * Copyright 2020-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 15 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/DelegatingAuthenticationConverter.java
```diff
@@ -1,68 +0,0 @@
...
- * <p>
- * Each {@link AuthenticationConverter} is given a chance to
- * {@link AuthenticationConverter#convert(HttpServletRequest)} with the first
- * {@code non-null} {@link Authentication} being returned.
- *
- * @author Joe Grandja
- * @since 0.0.2
- * @see AuthenticationConverter
- */
-public final class DelegatingAuthenticationConverter implements AuthenticationConverter {
```
jgrandja, line 41: This cannot be removed until `2.0` as it would be a breaking change for a minor.

Instead, please add `@Deprecated` at the class level and the `@deprecated` javadoc tag. See `applyDefaultSecurity()` as an example:

https://github.com/spring-projects/spring-authorization-server/blob/3b52e153dc82d837810e7e6c4547bec280deff2f/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configuration/OAuth2AuthorizationServerConfiguration.java#L76

jgrandja (changes requested):
Thanks for the PR @CrazyParanoid.

Please address the review comment and then should be ready to merge.

CrazyParanoid:
Hi @jgrandja, thanks for your feedback. All your comments have been resolved.

jgrandja:
Thanks for the update @CrazyParanoid ! This is now merged 👍

PR #1826, opened 2024-11-16
Use empty map for missing additional parameters

TheMrMilchmann:
This is not only more consistent with the other tokens which use an empty map, but also more in line with the documentation (at least for `OAuth2DeviceVerificationAuthenticationToken`).

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceAuthorizationRequestAuthenticationToken.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceAuthorizationRequestAuthenticationToken.java
@@ -96,7 +96,7 @@ public class OAuth2DeviceAuthorizationRequestAuthenticationToken extends Abstrac
 		this.deviceCode = deviceCode;
 		this.userCode = userCode;
 		this.authorizationUri = null;
-		this.additionalParameters = null;
+		this.additionalParameters = Collections.emptyMap();
 		setAuthenticated(true);
 	}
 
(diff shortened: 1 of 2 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceAuthorizationRequestAuthenticationToken.java
```diff
@@ -96,7 +96,7 @@ public OAuth2DeviceAuthorizationRequestAuthenticationToken(Authentication client
 		this.deviceCode = deviceCode;
 		this.userCode = userCode;
 		this.authorizationUri = null;
-		this.additionalParameters = null;
+		this.additionalParameters = Map.of();
```
jgrandja, line 99: Please use `Collections.emptyMap()` to be consistent with other usages in the framework - `Map.of()` is not currently used.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceVerificationAuthenticationToken.java
```diff
@@ -80,7 +80,7 @@ public OAuth2DeviceVerificationAuthenticationToken(Authentication principal, Str
 		this.principal = principal;
 		this.userCode = userCode;
 		this.clientId = clientId;
-		this.additionalParameters = null;
+		this.additionalParameters = Map.of();
```
jgrandja, line 83: See previous comment

jgrandja (changes requested):
Thanks for the PR @TheMrMilchmann.

Please see review comment and then we can get this merged.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceAuthorizationRequestAuthenticationToken.java
TheMrMilchmann: Done!

jgrandja:
Thanks for the updates @TheMrMilchmann. This is now merged.

PR #1842, opened 2024-11-25
Polish

ngocnhan-tran1996:
This PR includes:

- Use Pattern Matching
- Use `Collections.emptyMap()` instead.

I think `Collections.emptyMap()` return immutable value so we dont need use `Collections#unmodifiableMap` for wrapping purpose


jgrandja:
Thanks for the PR @ngocnhan-tran1996. This is now merged.

PR #1848, opened 2024-12-07
Polish diamond operator

ngocnhan-tran1996:


jgrandja:
Thanks for the polish @ngocnhan-tran1996 ! This is now merged.

Issue #1811, opened 2024-11-11
Support POST for authorization code request flow


sylvain-costanzo:
**Expected Behavior**
A POST /authorize without the `openid` scope should be accepted by the authorization server

**Current Behavior**
A POST /authorize without the openid scope returns a 404 with the following error log :
```
org.springframework.web.servlet.resource.NoResourceFoundException: No static resource oauth/authorize
```
![image](https://github.com/user-attachments/assets/6d4a5523-41ea-462a-b78f-e870d6b15b4f)

**Context**
Our implementation allow to not pass the scope parameter during the first /authorize request, as allowed by the oauth2 specification
https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1 and https://datatracker.ietf.org/doc/html/rfc6749#section-3.3
In this case our server act as if every scopes were requested.

There is no issue with spring authorization server when using a GET, but our implementation replays the /authorize request with a POST after a /login.

The `authorizationEndpointMatcher` of `OAuth2AuthorizationEndpointFilter` only match POST /authorize requests who have the `openid` scope : https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java#L156

I guess this is because the oauth2 spec states that the authorization server MAY support the use of POST https://datatracker.ietf.org/doc/html/rfc6749#section-3.1
while openid spec states that the server MUST support GET and POST https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

Unfortunately we cannot submit this second /authorize with a GET as it would break other parts of our system.
We considered tinkering with the RequestCache to add every scopes of the client in the SavedRequest, but a modification in spring authorization server would be simpler because we can just handle this use case in the token generators.

Is it possible to allow the customization of this matcher, or to remove the `openid` scope matcher for POST in OAuth2AuthorizationEndpointFilter ?

sylvain-costanzo:
Hello @jgrandja  and thank you for your answer

The main issue comes from our custom implementation of the `prompt=login`.
We are using an authentication converter to catch the `/authorize` request if `prompt=login` is detected within a GET request.
Using a POST during our "replay" of the `/authorize` request allow us to not loop indefinitely because of this converter.

We understand that our implementation may be sub-optimal, and that it is the first time someone request this feature.
[...]

PR #1874, opened 2025-01-13
Support POST for authorization code request flow

sylvain-costanzo:
Closes [gh-1811](https://github.com/spring-projects/spring-authorization-server/issues/1811)

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
@@ -39,7 +39,6 @@ import org.springframework.security.oauth2.core.OAuth2AuthenticationException;
 import org.springframework.security.oauth2.core.OAuth2Error;
 import org.springframework.security.oauth2.core.endpoint.OAuth2AuthorizationResponse;
 import org.springframework.security.oauth2.core.endpoint.OAuth2ParameterNames;
-import org.springframework.security.oauth2.core.oidc.OidcScopes;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeRequestAuthenticationException;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeRequestAuthenticationProvider;
 import org.springframework.security.oauth2.server.authorization.authentication.OAuth2AuthorizationCodeRequestAuthenticationToken;
@@ -57,9 +56,7 @@ import org.springframework.security.web.authentication.WebAuthenticationDetailsS
 import org.springframework.security.web.authentication.session.SessionAuthenticationStrategy;
 import org.springframework.security.web.util.RedirectUrlBuilder;
 import org.springframework.security.web.util.UrlUtils;
-import org.springframework.security.web.util.matcher.AndRequestMatcher;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
-import org.springframework.security.web.util.matcher.NegatedRequestMatcher;
 import org.springframework.security.web.util.matcher.OrRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
 import org.springframework.util.Assert;
@@ -151,20 +148,8 @@ public final class OAuth2AuthorizationEndpointFilter extends OncePerRequestFilte
 				HttpMethod.GET.name());
 		RequestMatcher authorizationRequestPostMatcher = new AntPathRequestMatcher(authorizationEndpointUri,
 				HttpMethod.POST.name());
-		RequestMatcher openidScopeMatcher = (request) -> {
-			String scope = request.getParameter(OAuth2ParameterNames.SCOPE);
-			return StringUtils.hasText(scope) && scope.contains(OidcScopes.OPENID);
-		};
-		RequestMatcher responseTypeParameterMatcher = (
-				request) -> request.getParameter(OAuth2ParameterNames.RESPONSE_TYPE) != null;
 
-		RequestMatcher authorizationRequestMatcher = new OrRequestMatcher(authorizationRequestGetMatcher,
-				new AndRequestMatcher(authorizationRequestPostMatcher, responseTypeParameterMatcher,
-						openidScopeMatcher));
-		RequestMatcher authorizationConsentMatcher = new AndRequestMatcher(authorizationRequestPostMatcher,
-				new NegatedRequestMatcher(responseTypeParameterMatcher));
-
-		return new OrRequestMatcher(authorizationRequestMatcher, authorizationConsentMatcher);
+		return new OrRequestMatcher(authorizationRequestGetMatcher, authorizationRequestPostMatcher);
 	}
 
 	@Override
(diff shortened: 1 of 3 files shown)


sylvain-costanzo:
The removal of the oidc scope matcher now allow OAuth2AuthorizationEndpointFilter to send POST /authorize request without the openid scope to the converters.

But I see two issues with this development:

1- The default converter `OAuth2AuthorizationCodeRequestAuthenticationConverter` still need the oidc scope to be used, but removing [this condition](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2AuthorizationCodeRequestAuthenticationConverter.java#L71) will mess things up with the consent request, who, in the default implementation and in the demo-authorizationserver sample, is also a POST without the openid scope (and without the response_type parameter, who make this converter throw an error).

Tackling this issue seems a bit out of scope for this PR so I didn't touch anything, and it can be surely be solved with some custom converters.


2- Another minor issue I see is that the oauth2 specification is saying that a /authorize request (GET or POST) is supposed [to have the parameters in the query string](https://datatracker.ietf.org/doc/html/rfc6749#section-4.1.1),
> The client constructs the request URI by adding the following parameters to the query component of the authorization endpoint URI using the "application/x-www-form-urlencoded" format

contradicting the OIDC spec saying that [POST requests must use the body](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest)
> If using the HTTP POST method, the request parameters are serialized using Form Serialization, per Section 13.2.

So [this way of getting the request parameters](https://github.com/spring-projects/spring-authorization-server/blob/main/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/authentication/OAuth2AuthorizationCodeRequestAuthenticationConverter.java#L76) may have to be updated, but again it can be solved with a custom converter, so it also felt out of scope of this PR.
 

Considering these 2 issues, I am not sure the PR is answering the "Support POST for authorization code request flow", and I would like your opinion on this

sylvain-costanzo:
Rewording my thoughts, I don't think these issues are completely out of scope, but they need significant refactor with more regression risks than the small change here.
The original goal of [gh-1811](https://github.com/spring-projects/spring-authorization-server/issues/1811) was to allow a POST /authorize without scope parameter to be sent to the converters, and this PR allows it.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
```diff
@@ -151,20 +148,8 @@ private static RequestMatcher createDefaultRequestMatcher(String authorizationEn
 				HttpMethod.GET.name());
 		RequestMatcher authorizationRequestPostMatcher = new AntPathRequestMatcher(authorizationEndpointUri,
 				HttpMethod.POST.name());
-		RequestMatcher openidScopeMatcher = (request) -> {
-			String scope = request.getParameter(OAuth2ParameterNames.SCOPE);
-			return StringUtils.hasText(scope) && scope.contains(OidcScopes.OPENID);
-		};
-		RequestMatcher responseTypeParameterMatcher = (
```
jgrandja, line 158: `responseTypeParameterMatcher` should not be removed as it will break the Authorization Consent flow. Only remove `openidScopeMatcher` here as well as in `OAuth2AuthorizationCodeRequestAuthenticationConverter`

sylvain-costanzo: In `OAuth2AuthorizationEndpointFilter`, removing the openid scope and letting the response_type would result of a matcher like this:
```
(GET or (POST and response_type)) or (POST and !response_type)
```

which is the same as (GET or POST) that's why I removed the response_type matcher.

I can add it back if you prefer it that way.

sylvain-costanzo: I removed the openid scope matcher from the converter `OAuth2AuthorizationCodeRequestAuthenticationConverter`

jgrandja: > would result of a matcher like this:
> (GET or (POST and response_type)) or (POST and !response_type)

I believe we still need the `response_type` parameter in the matching because it's the difference between an Authorization Request or Authorization Consent Request.

However, I think we need to adjust the matching as follows:

**Authorization Request**
`(GET or POST) and response_type`

**Authorization Consent Request**
`POST and !response_type`

This should be applied to `OAuth2AuthorizationEndpointFilter` and both `OAuth2AuthorizationCodeRequestAuthenticationConverter` and `OAuth2AuthorizationConsentAuthenticationConverter` to be consistent.

jgrandja:
@sylvain-costanzo 

The Spring team has recently migrated to the [Developer Certificate of Origin](https://en.wikipedia.org/wiki/Developer_Certificate_of_Origin) (DCO) for our contribution process. See [Submitting Pull Requests](https://github.com/spring-projects/spring-authorization-server/blob/main/CONTRIBUTING.adoc#submitting-pull-requests) for additional details on the new process. Please format the commits in this PR as the DCO check did not pass.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/web/OAuth2AuthorizationEndpointFilter.java
sylvain-costanzo: Adding the `response_type` matcher to the GET request in `OAuth2AuthorizationEndpointFilter` will make the server return a 404 if a GET /authorize request is performed without the required `response_type` parameter, instead of a 400.

Specifically it is breaking the test `OAuth2AuthorizationEndpointFilterTests.doFilterWhenAuthorizationRequestMissingResponseTypeThenInvalidRequestError()` so I don't think adding this matcher in `OAuth2AuthorizationEndpointFilter` is a good solution.

For the same reason, adding these matcher rules to `OAuth2AuthorizationCodeRequestAuthenticationConverter` will make the request not enter the converter and prevent the server to return a 400 (it is breaking the same test).

---

This highlight the fact that a POST /authorize request without the `response_type` parameter will be treated as a consent request, so maybe we could add a condition in `OAuth2AuthorizationConsentAuthenticationConverter` to check if the user is authenticated?
A way to solve all this could be to create a new converter after `OAuth2AuthorizationCodeRequestAuthenticationConverter` and `OAuth2AuthorizationConsentAuthenticationConverter` to catch this missing parameter, so we can use the matcher in the converters?

jgrandja: Sorry, the matching rules stated in the [previous comment](https://github.com/spring-projects/spring-authorization-server/pull/1874#discussion_r1922609281) need to be adjusted as follows:

**Authorization Request**
`GET or (POST and response_type)`

**Authorization Consent Request**
`POST and !response_type`

I tried this on my end and all tests pass.

sylvain-costanzo: I added a distinction between the Authorize matcher and the Consent matcher in `OAuth2AuthorizationEndpointFilter`.
A previous commit updated `OAuth2AuthorizationCodeRequestAuthenticationConverter`.
Nothing to do for `OAuth2AuthorizationConsentAuthenticationConverter`
It should be fine now

jgrandja: @sylvain-costanzo I will get to this next week. I'm trying to wrap up a priority task by end of this week.

jgrandja:
Thanks for the updates @sylvain-costanzo. This is now merged.

FYI, I added a polish commit. Please let me know if you have any questions around that.

PR #1887, opened 2025-01-28
Fix broken links in file README

ngocnhan-tran1996:


jgrandja:
Thank you for the update @ngocnhan-tran1996. This is now merged.

Issue #1885, opened 2025-01-23
Device Access Token Response with expired device code returns "authorization_pending" and not "expired_token"


nwholloway:
If you make a valid [_Device Access Token Request_](https://www.rfc-editor.org/rfc/rfc8628.html#section-3.4) and do not grant/deny consent from the provided `verification_uri_complete`, then all polls of the token endpoint will indefinitely return `authorization_pending`.

If you have a client which will continue polling while receiving `authorization_pending` (rather than using `expires_in`), then the client will poll forever.

The reason for this is that `authorization_pending` is always returned if the user code is not invalidated ([Line 150](/spring-projects/spring-authorization-server/blob/740a59e04b1364d9837be9693b1627338fa35f44/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceCodeAuthenticationProvider.java#L150)), irrespective of the age of the device code.

The only operation that will invalidate the user code is either granting/denying consent from the device verification consent page.

I believe that the check for the device code expiring ([Line 173](/spring-projects/spring-authorization-server/blob/740a59e04b1364d9837be9693b1627338fa35f44/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceCodeAuthenticationProvider.java#L173)) should be moved before the check for user code invalidation.

The test [`authenticateWhenDeviceCodeIsExpiredThenThrowOAuth2AuthenticationException`](/spring-projects/spring-authorization-server/blob/740a59e04b1364d9837be9693b1627338fa35f44/oauth2-authorization-server/src/test/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceCodeAuthenticationProviderTests.java#L235) only passes because the user code is created in an invalidated state.

The following update to the test demonstrates the failure.

```diff
        @Test
        public void authenticateWhenDeviceCodeIsExpiredThenThrowOAuth2AuthenticationException() {
                RegisteredClient registeredClient = TestRegisteredClients.registeredClient().build();
                Authentication authentication = createAuthentication(registeredClient);
                OAuth2Authorization authorization = TestOAuth2Authorizations.authorization(registeredClient)
                        .token(createExpiredDeviceCode())
-                       .token(createUserCode(), withInvalidated())
+                       .token(createUserCode())
                        .build();
[...]

PR #1889, opened 2025-01-31
Correct Device Access Token error responses gh-1885

nwholloway:
This fixes two occurrences of the Device Access Token Response returning "authorization_pending" and not the expected response.

The first is when the device token has expired (as raised in #1885).  This now returns the error "expired_token".

The second is when the device token has been invalidated, but the user token has not been invalidated.  This state occurs when a request has been made with the incorrect client, or after device token expiry. This now returns the error "invalid_grant".


jgrandja:
@nwholloway Thank you. I will get to this next week. I'm trying to wrap up a priority task by end of this week.

jgrandja:
Thanks for the PR @nwholloway. This is now merged.

FYI, I added a polish commit. Please let me know if you have any questions for that.

PR #1907, opened 2025-02-15
Use pattern matching

arefbehboudi:


jgrandja:
Thanks for the updates @arefbehboudi. This is now merged.

PR #1914, opened 2025-02-21
Use javadoc-extension

ngocnhan-tran1996:
The main points:

- The new Spring Authorization Server version document will be compatitable with Spring Framework version document
- Reference javadoc will be easier, use `javadoc:org.springframework.security...` instead `{spring-security-api-base-url}...`

### DIFF
--- a/docs/spring-authorization-server-docs.gradle
+++ b/docs/spring-authorization-server-docs.gradle
@@ -30,8 +30,15 @@ tasks.named("generateAntoraYml") {
 
 
 def generateAttributes() {
+	def springBootVersion = project.configurations.named("implementation").get()
+		.dependencies.find {it -> ("spring-boot-dependencies" == it.name) }
+		.version
+	String springSecurityVersion = libs.org.springframework.security.spring.security.bom.get().versionConstraint.displayName
 	return [
-			"spring-authorization-server-version": project.version
+			"spring-authorization-server-version": project.version,
+			"spring-security-reference-base-url": "https://docs.spring.io/spring-security/reference/${generateVersionWithoutPatch(springSecurityVersion)}".toString(),
+			"javadoc-location-org-springframework-security": "https://docs.spring.io/spring-security/site/docs/$springSecurityVersion/api".toString(),
+			"spring-boot-reference-base-url": "https://docs.spring.io/spring-boot/docs/${generateVersionWithoutPatch(springBootVersion)}/reference/html".toString()
 	]
 }
 
@@ -75,4 +82,17 @@ dependencies {
 
 tasks.named("test") {
 	useJUnitPlatform()
+}
+
+static String generateVersionWithoutPatch(String version) {
+
+	def matcher = version =~ /^(\d+.\d+).\d+(-SNAPSHOT|-M\d+)?$/
+	if(matcher) {
+
+		return matcher[0][2] == null || matcher[0][2].startsWith("-M")
+			? matcher[0][1]
+			: matcher[0][1] + matcher[0][2]
+	}
+
+	return version
 }
--- a/docs/spring-authorization-server-docs.gradle
+++ b/docs/spring-authorization-server-docs.gradle
@@ -35,10 +35,10 @@ def generateAttributes() {
 		.version
 	String springSecurityVersion = libs.org.springframework.security.spring.security.bom.get().versionConstraint.displayName
 	return [
-			"spring-authorization-server-version": project.version,
-			"spring-security-reference-base-url": "https://docs.spring.io/spring-security/reference/${generateVersionWithoutPatch(springSecurityVersion)}".toString(),
-			"javadoc-location-org-springframework-security": "https://docs.spring.io/spring-security/site/docs/$springSecurityVersion/api".toString(),
-			"spring-boot-reference-base-url": "https://docs.spring.io/spring-boot/docs/${generateVersionWithoutPatch(springBootVersion)}/reference/html".toString()
+			'spring-authorization-server-version': project.version,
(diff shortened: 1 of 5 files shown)


docs/spring-authorization-server-docs.gradle
```diff
@@ -30,8 +30,15 @@ tasks.named("generateAntoraYml") {
...
 def generateAttributes() {
+	def springBootVersion = project.configurations.named("implementation").get()
+		.dependencies.find {it -> ("spring-boot-dependencies" == it.name) }
+		.version
+	String springSecurityVersion = libs.org.springframework.security.spring.security.bom.get().versionConstraint.displayName
 	return [
-			"spring-authorization-server-version": project.version
+			'spring-authorization-server-version': project.version,
+			'spring-security-reference-base-url': "https://docs.spring.io/spring-security/reference/${generateVersionWithoutPatch(springSecurityVersion)}".toString(),
+			'javadoc-location-org-springframework-security': "https://docs.spring.io/spring-security/site/docs/$springSecurityVersion/api".toString(),
```
jgrandja, line 40: I'm not sure if this is correct?

I reviewed the changes in Spring Security [spring-security#15386](https://github.com/spring-projects/spring-security/issues/15386) and the gradle task [syncAntoraAttachments](https://github.com/spring-projects/spring-security/commit/5bd4db1a13011e50011b58c412dc1b566ae8988b#diff-aab37362d26f57bdb166a26fc4ca122d72faf4af628462ae56b71f0b4486764cR18) was added which copies the javadoc as an asset to the site. I'm wondering if we need to do the same here?

@rwinch Is there a way to copy the Spring Security javadoc as an asset to the Spring Authorization Server site docs and reference it via a site asset? Or is this correct by directly referencing the Spring Security javadoc site?

jgrandja: @rwinch As a side note, it appears [syncAntoraAttachments](https://github.com/spring-projects/spring-security/commit/5bd4db1a13011e50011b58c412dc1b566ae8988b#diff-aab37362d26f57bdb166a26fc4ca122d72faf4af628462ae56b71f0b4486764cR18) is a required gradle task otherwise the antora build will fail saying it cannot find the javadoc resource. For example, after adding `javadoc:org.springframework.security.oauth2.server.authorization.OAuth2AuthorizationCode[]`, the build will fail unless I add the `syncAntoraAttachments` task. However, this is not documented in [Javadoc](https://github.com/spring-io/asciidoctor-extensions?tab=readme-ov-file#javadoc), unless I'm missing something?

jgrandja (changes requested):
Thanks for the PR @ngocnhan-tran1996. Please see review comments.

docs/spring-authorization-server-docs.gradle
ngocnhan-tran1996: If we have the way copy the Spring Security javadoc to to the Spring Authorization Server, please let me know

rwinch: > @rwinch Is there a way to copy the Spring Security javadoc as an asset to the Spring Authorization Server site docs and reference it via a site asset? 

You should copy Spring Authorization Servers javadoc to the Antora site, but you should NOT copy Spring Security's Javadoc.

The javadoc extension is just syntactic sugar for creating a link to Javadoc and can operate in a numerous different ways. It creates a link and does NOT create/copy Javadoc in the location it is pointing to.

## Linking to Spring Authorization Server Javadoc

The default mode is to use `xref:attachment$api/java` which, as an xref to an [attachment](https://docs.antora.org/antora/latest/page/attachments/), will be validated by Antora that the link exists. This means you can get validation of the javadoc links existing for free. You will use this mode for linking to Spring Authorization Server javadoc. Since the javadoc extension is just creating a link, you still need to ensure that the Javadoc is added.

Spring Security [configures](https://github.com/spring-projects/spring-security/blob/64d3397a9c373b6f1c3b28e9e56714fdb2ec1e1e/docs/antora.yml#L9) the [Antora Collector Extension](https://gitlab.com/antora/antora-collector) to invoke the [generateAntoraResources](https://github.com/spring-projects/spring-security/blob/64d3397a9c373b6f1c3b28e9e56714fdb2ec1e1e/docs/spring-security-docs.gradle#L33) task (and in turn [generateAntoraYml and syncAntoraAttachments](https://github.com/spring-projects/spring-security/blob/64d3397a9c373b6f1c3b28e9e56714fdb2ec1e1e/docs/spring-security-docs.gradle#L34)). The `generateAntoraYml` task is how dynamic properties like version are set properly and `synAntoraAttachments` is how the Javadoc is generated/added as attachments to the site. Once it is added to the site, the Javadoc is automatically published as a part of your site.

## Linking to Other Module's Javadoc 

As for linking to other module's Javadoc (e.g. Spring Security), you should NOT copy it. Instead use [Package Specific Locations](https://github.com/spring-io/asciidoctor-extensions?tab=readme-ov-file#package-specific-locations) or [Link Specific Locations](https://github.com/spring-io/asciidoctor-extensions?tab=readme-ov-file#link-specific-locations). You will notice that these demonstrate ways that external sites can be linked to. In this case it is more obvious, but worth pointing out that the javadoc extension does not publish the javadocs. It is just syntactic sugar for creating a link to something that must already exist.

jgrandja:
@ngocnhan-tran1996 

As you may be aware, Spring Authorization Server has moved to Spring Security (gh-2195). The `1.4.x` and `1.5.x` branch will continue to be maintained with bug and security fixes, however, all new features and enhancements will need to be added in Spring Security from `7.0` onwards.

Given that this PR is an enhancement and Spring Security is already using the `javadoc-extension`, I'll go ahead and close this PR.

Issue #1882, opened 2025-01-20
Add ability to disable an endpoint


SledgeHammer01:
**Expected Behavior**

**Current Behavior**
Out of the box experience is that 

**Context**

jgrandja:
@SledgeHammer01 This enhancement makes sense. We'll consider adding it.

AlTurner-MOJ:
We are in an identical situation and following a PEN test have been asked by our AppSec team to disable unused end points. We are only using the `/oauth2/token` and `/oauth2/authorize` end points.

tommyttf:
Hi @jgrandja, I am interested in working on this issue, and have read your comment in #1498 that we better not to have a flag setting for each configurer. How about we have a`removeConfigurer` function and `disableXXXEndpoint` for each configurer

For example in `OAuth2AuthorizationServerConfigurer`
```
	public OAuth2AuthorizationServerConfigurer disableDeviceVerificationEndpoint() {
		removeConfigurer(OAuth2DeviceVerificationEndpointConfigurer.class);
		return this;
	}

...

	@SuppressWarnings("unchecked")
	private void removeConfigurer(Class<?> type) {
		this.configurers.remove(type);
	}
```

Then we can disable the endpoint by
```
		http
			.with(authorizationServerConfigurer, (authorizationServer) ->
				authorizationServer
						.disableDeviceVerificationEndpoint()
			)
```

What do you think?

OrangeDog:
This should follow the same pattern as `AbstractHttpConfigurer`, and add `disable()` to `AbstractOAuth2Configurer`.

I do not like @tommyttf's solution at all, as it's inconsistent with anything else.

dineshgupta630:
Absolutely! Here it is, formatted for a GitHub comment—using Markdown, British spelling, and a casual but clear tone:

---

Hi @jgrandja and @OrangeDog,

Just wanted to suggest an idea for disabling endpoints that fits in with how Spring Security usually does things, without adding loads of new public APIs.

-  Introduce an `AbstractOAuth2Configurer` class, modelled after `AbstractHttpConfigurer`, which provides a standard `disable()` method and a `disabled` flag.

- Each endpoint configurer (e.g. `AuthorizationEndpointConfigurer`, `TokenEndpointConfigurer`, etc.) would extend `AbstractOAuth2Configurer`.

[...]

PR #1924, opened 2025-03-03
Add ability to disable an endpoint

tommyttf:
Add ability to disable endpoint

Related gh-1882

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
@@ -15,23 +15,14 @@
  */
 package org.springframework.security.oauth2.server.authorization.config.annotation.web.configurers;
 
-import java.net.URI;
-import java.util.ArrayList;
-import java.util.LinkedHashMap;
-import java.util.List;
-import java.util.Map;
-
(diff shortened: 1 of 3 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
```diff
@@ -170,6 +166,11 @@ public OAuth2AuthorizationServerConfigurer clientAuthentication(
 		return this;
 	}
 
+	public OAuth2AuthorizationServerConfigurer disableClientAuthentication() {
```
jgrandja, line 169: As mentioned in this [comment](https://github.com/spring-projects/spring-authorization-server/pull/1498#discussion_r1456470201):

> I also think we should come up with a solution that allows other endpoints to be disabled if needed. **I don't think a "flag" setting on all the configurers makes sense.**

I consider `disableClientAuthentication()`, `disableAuthorizationServerMetadataEndpoint()`, `disableAuthorizationEndpoint()`, etc. to be a similar solution as proposed in the previous PR but with an "explicit" flag through the introduction of the `disable**` public methods.

As mentioned, this is still not ideal as it introduces too many new public API's and I would prefer a solution that is more isolated.

I have been thinking about this for a while and may have a solution to address this but unfortunately I don't have the time at the moment as I have a couple of other features I've been working on that are priority for the upcoming release.

I'm going to close this PR but if you have other ideas please comment in gh-1882 before submitting another PR.

PR #1928, opened 2025-03-05
Update OAuth2AuthorizationServerConfigurer.java

george510257:
oauth2ResourceServer should be configured externally, along with JWTs and Opaque Tokens.


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2AuthorizationServerConfigurer.java
```diff
@@ -391,18 +391,18 @@ public void init(HttpSecurity httpSecurity) throws Exception {
...
-			if (oidcConfigurer.getConfigurer(OidcUserInfoEndpointConfigurer.class) != null
-					|| oidcConfigurer.getConfigurer(OidcClientRegistrationEndpointConfigurer.class) != null) {
-				httpSecurity
-					// Accept access tokens for User Info and/or Client Registration
-					.oauth2ResourceServer(
-							(oauth2ResourceServer) -> oauth2ResourceServer.jwt(Customizer.withDefaults()));
-
-			}
-		}
+		// oauth2ResourceServer should be configured externally, along with JWTs and Opaque Tokens.
```
jgrandja, line 394: This was updated in this commit 2c7975485fca39338b1fa4457b84708b0296eca2 and is expected as it simplifies (minimizes) the configuration.

PR #1955, opened 2025-04-01
Prevent NPE

ngocnhan-tran1996:


jgrandja:
Thanks for the fix @ngocnhan-tran1996. This is now merged.

PR #1972, opened 2025-04-16
Replace @MockBean with @MockitoBean

DevDengChao:


jgrandja:
Thanks for the PR @DevDengChao. This is now merged.

Issue #1977, opened 2025-04-17
User Code expiration is not considered


antoinelauzon-bell:
Hi,

**Describe the bug**

`OAuth2DeviceVerificationAuthenticationProvider` only gets the provided user code but does not check its expiration. It means that a user can input an expired user code and still "complete" the flow on its second device. The flow would later fail on the `/oauth2/token` call because the device code is expired.

**To Reproduce**

Use the "DemoAuthorizationServerApplication" sample.

Before starting the sample, reduce the device/user code duration (to not wait too long):

```
.tokenSettings(TokenSettings.builder().deviceCodeTimeToLive(Duration.ofSeconds(30)).build())
```

Start a device code flow with some scope(s). For example, use the request below to start the flow:

```
POST http://localhost:9000/oauth2/device_authorization
scope=message.read&client_id=device-messaging-client
```

Wait for the expiration (30s) before continuing the flow on the second device (i.e. entering `http://localhost:9000/activate?user_code=X` in a browser).

The browser shows that the flow "correctly" completed. But once the tokens are requested (token endpoint), it return the error "expired_token". It seems to me that the error happens too late. The user should be blocked on its second device because the user code is expired.

**Expected behavior**

An error should occur on the second device if the user code is expired.

I can provide additional details, just let me know.

Thank you

jgrandja:
@antoinelauzon-bell Thank you for logging this issue. I agree that the user code should be checked for expiry (or if invalidated) in `OAuth2DeviceVerificationAuthenticationProvider`. Would you be interested in submitting a fix?

Issue #1894, opened 2025-02-05
[Device Code Grant] Submitting the same user code more than once results in internal server error


moscicky:
**Describe the bug**
[Device Code Grant] Submitting the same user code more than once results in internal server error cause by NPE

**To Reproduce**
1. [Run the demo sample](https://github.com/spring-projects/spring-authorization-server/tree/main/samples#demo-sample)
2. Go to [http://127.0.0.1:8080/authorize?grant_type=device_code](http://127.0.0.1:8080/authorize?grant_type=device_code)
3. Click `activate` and copy the user code
4. Go to [http://localhost:9000/activate](http://localhost:9000/activate), paste the user code and submit. You will see success message.
5. Go to [http://localhost:9000/activate](http://localhost:9000/activate), paste the user code again and submit. You will see error message, internal server error is returned.

This is because there is an NPE `OAuth2DeviceVerificationAuthenticationProvider` because the code assumes that `SCOPE` attribute will not be null but it is erased on the first verification. This can be probably fixed by checking whether the user code has been previously invalidated.

```java
java.lang.NullPointerException: Cannot invoke "java.util.Collection.iterator()" because "c" is null
        at java.base/java.util.AbstractCollection.containsAll(AbstractCollection.java:310) ~[na:na]
        at org.springframework.security.oauth2.server.authorization.authentication.OAuth2DeviceVerificationAuthenticationProvider.requiresAuthorizationConsent(OAuth2DeviceVerificationAuthenticationProvider.java:194) ~[spring-security-oauth2-authorization-server-1.4.2-SNAPSHOT.jar:1.4.2-SNAPSHOT]
        at org.springframework.security.oauth2.server.authorization.authentication.OAuth2DeviceVerificationAuthenticationProvider.authenticate(OAuth2DeviceVerificationAuthenticationProvider.java:134) ~[spring-security-oauth2-authorization-server-1.4.2-SNAPSHOT.jar:1.4.2-SNAPSHOT]
        at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:182) ~[spring-security-core-6.3.0.jar:6.3.0]
        at org.springframework.security.oauth2.server.authorization.web.OAuth2DeviceVerificationEndpointFilter.doFilterInternal(OAuth2DeviceVerificationEndpointFilter.java:163) ~[spring-security-oauth2-authorization-server-1.4.2-SNAPSHOT.jar:1.4.2-SNAPSHOT]
[...]

PR #1997, opened 2025-05-05
Check user code expiry and invalidity

antoinelauzon-bell:
It ensures that a user code is neither expired nor invalidated during the verification step. See gh-1894 and gh-1977 for more details.

Notes:

- `invalidat_grant` appears to be the expected error code (see  [RFC 6749, Section 5.2](https://datatracker.ietf.org/doc/html/rfc6749#section-5.2)). It might be useful to distinguish these exceptions though (e.g. by using subclasses of OAuth2AuthenticationException).
- A small gap remains where a user could verify a user code in the final seconds before it expires, leasing to an expired device code on the next poll from the initial device. This scenario would require very unlucky timing.


OrangeDog:
Would it not be better to just check `isActive()`? Not only is it less code, but it also covers cases you have missed.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceVerificationAuthenticationProvider.java
```diff
@@ -109,6 +109,21 @@ public Authentication authenticate(Authentication authentication) throws Authent
...
+				this.logger.trace("User code is invalided");
+			}
+			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
+		}
+
+		if (userCode.isExpired()) {
+			if (this.logger.isTraceEnabled()) {
+				this.logger.trace("User code is expired");
+			}
+			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
```
colin-riddell, line 124: Is it worth providing more context on the error..
```suggestion
						OAuth2Error error = new OAuth2Error(OAuth2ErrorCodes.INVALID_GRANT, "User code is expired.", null);
						throw new OAuth2AuthenticationException(error);
```

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceVerificationAuthenticationProvider.java
```diff
@@ -109,6 +109,21 @@ public Authentication authenticate(Authentication authentication) throws Authent
 			this.logger.trace("Retrieved authorization with user code");
 		}
 
+		OAuth2Authorization.Token<OAuth2UserCode> userCode = authorization.getToken(OAuth2UserCode.class);
+		if (userCode.isInvalidated()) {
+			if (this.logger.isTraceEnabled()) {
+				this.logger.trace("User code is invalided");
+			}
+			throw new OAuth2AuthenticationException(OAuth2ErrorCodes.INVALID_GRANT);
```
colin-riddell, line 117: Is it worth providing more context on the error..
```suggestion
						OAuth2Error error = new OAuth2Error(OAuth2ErrorCodes.INVALID_GRANT, "User code is invalidated.", null);
						throw new OAuth2AuthenticationException(error);
```

jgrandja:
@antoinelauzon-bell Apologies for the late response but I've been busy preparing for the `1.5` release this Tuesday. I will likely review this next week.

jgrandja (changes requested):
Thanks for this fix @antoinelauzon-bell. Please see review comment.

Also, please update copyright year for all files and rebase off of `1.3.x`.

Thanks.

Issue #1950, opened 2025-03-26
Support injecting clock into token generation code


symposion:
`JwtGenerator` currently does this to get the token issued time:

```
Instant issuedAt = Instant.now();
```

Compare this against how the JwtTimestampValidator in the Spring resource-server implementation is code:

```
 Instant expiry = jwt.getExpiresAt();
if (expiry != null && Instant.now(this.clock).minus(this.clockSkew).isAfter(expiry)) {
    OAuth2Error oAuth2Error = this.createOAuth2Error(String.format("Jwt expired at %s", jwt.getExpiresAt()));
    return OAuth2TokenValidatorResult.failure(new OAuth2Error[]{oAuth2Error});
} 
```

Notice that you can override the clock in the validator to change the definition of "now" It would be very useful if the same thing could be done on the generation side as well.

**Context**

We have an extensive testing framework which makes of comparison against known-good outputs. But obviously if these outputs include any kind of time derived from the current time of execution, the test output becomes variable. Rather than having to clutter up our test code with manual timestamp assertions for all of these, we require all of our services to obtain the current time from a special service that supports "fixing" time and then manually adjusting during the course of a test. This allows for stable test output even when timestamps are involved.

It would be a trivial change to support overriding the clock in the JwtGenerator instead of acquiring the "now" value directly from `Instant.now()` I'm happy to submit a PR

DevDengChao:
I’ll vote up for this.

I wrote a lot of tests regarding the concept of `now` and `today`. However, when the test cases ran past 23:59:59, they failed. To address this, I had to inject a `TimestampSupplier` into the logic, allowing me to control the exact `now` during the test execution.

AlessandroMinoccheri:
I am trying to solve it @jgrandja , feedback is welcome if the direction is good or not.
https://github.com/spring-projects/spring-authorization-server/pull/2004

lnhrdt:
Just wanted to chime in that we hit this today too. We were testing our authorization server alongside other services and needed control over time to simulate various flows. We assumed that registering a Clock bean would be enough, like it is in most of the Spring ecosystem. That consistency is usually a strong point with Spring, so this felt like an unexpected gap.

Would definitely appreciate a way to inject a custom clock here.

jgrandja:
@lnhrdt 

> We assumed that registering a Clock bean would be enough, like it is in most of the Spring ecosystem.

Which Spring project supports a `Clock` `@Bean`? Can you point me to the reference?

[...]

PR #2004, opened 2025-05-08
injecting clock when we are generating the token

AlessandroMinoccheri:
In this PR I am trying to solve this issue: https://github.com/spring-projects/spring-authorization-server/issues/1950
By injecting clock into the token generation


jgrandja:
@AlessandroMinoccheri Apologies for the late response but I've been busy preparing for the `1.5` release this Tuesday. I will likely review this next week.

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/token/JwtGenerator.java
```diff
@@ -61,16 +62,19 @@
...
+	private final Clock clock;
 
 	private OAuth2TokenCustomizer<JwtEncodingContext> jwtCustomizer;
 
 	/**
 	 * Constructs a {@code JwtGenerator} using the provided parameters.
 	 * @param jwtEncoder the jwt encoder
 	 */
-	public JwtGenerator(JwtEncoder jwtEncoder) {
+	public JwtGenerator(JwtEncoder jwtEncoder, Clock clock) {
```
jgrandja, line 73: This is a breaking change. Alternatively, we could add an overloaded constructor with `JwtEncoder, Clock`, but we typically favour setters (e.g.  `setClock(Clock)`) when the attribute is not required.

However, even with the setter approach, there are [too many classes](https://github.com/spring-projects/spring-authorization-server/issues/1950#issuecomment-2762147772) that require this public API change and my preference would be to minimize this.

Another possible solution would be to obtain the `Clock` from `AuthorizationServerContext`, which is available in `OAuth2TokenContext`. Thoughts?

jgrandja: > How much is problematic to add a new public method in terms of integrations?

It's not problematic but configuration could be verbose as there are up to [6x OAuth2TokenGenerator](https://github.com/spring-projects/spring-authorization-server/issues/1950#issuecomment-2762147772) that would need to be configured.

> taking things from contexts seems everytime and overhead of resources and it can create side effects.

I'm not sure I understand. Why do you think it would be an overhead in resources and what side effects are you referring to?

AlessandroMinoccheri: Sorry, I will rephrase my sentence with more info.
For me, use context has different cons:
AuthorizationServerContext is request-scoped
- The AuthorizationServerContext is tied to the current request lifecycle.
- If you try to extract the Clock from this context outside of a token generation request (e.g. in background jobs, unit tests, or beans instantiated at application startup), the context won't exist, leading to IllegalStateException or null values.

It violates separation of concerns
- The Clock is a core time abstraction often needed for testing, validation, and system logic.
- Getting it from a request-scoped context couples your logic too tightly to the OAuth2 token issuance flow, making your code harder to test.

- When using Clock from AuthorizationServerContext, you cannot easily inject mock clocks in unit or integration tests.

jgrandja: All very good points @AlessandroMinoccheri.

Ok, let's keep it simple and add `setClock()` to all `OAuth2TokenGenerator`.

AlessandroMinoccheri:
@jgrandja can you give me some feedback for this PR?

docs/modules/ROOT/pages/core-model-components.adoc
```diff
@@ -393,6 +393,7 @@ The following example shows how to register an `OAuth2TokenGenerator` `@Bean`:
 public OAuth2TokenGenerator<?> tokenGenerator() {
 	JwtEncoder jwtEncoder = ...
 	JwtGenerator jwtGenerator = new JwtGenerator(jwtEncoder);
+	jwtGenerator.setClock(Clock.systemUTC());
```
jgrandja, line 396: Please remove this as `Clock.systemUTC()` is the default (see other comment).

oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/config/annotation/web/configurers/OAuth2ConfigurerUtils.java
```diff
@@ -128,6 +129,7 @@ private static JwtGenerator getJwtGenerator(HttpSecurity httpSecurity) {
 			JwtEncoder jwtEncoder = getJwtEncoder(httpSecurity);
 			if (jwtEncoder != null) {
 				jwtGenerator = new JwtGenerator(jwtEncoder);
+				jwtGenerator.setClock(Clock.systemUTC());
```
jgrandja, line 132: Please remove this explicit configuration since `Clock.systemUTC()` is the default. Also, please remove this in all other places including tests.

jgrandja (changes requested):
Thanks for the updates @AlessandroMinoccheri. Please see review comments.

jgrandja:
Closing this in favour of https://github.com/spring-projects/spring-security/commit/469ed096459c6015a38d0849f5a3f53f9d1cf393

Issue #1965, opened 2025-04-09
Setting requireAuthorizationConsent is ignored for device code grant


antoinelauzon-bell:
Hi,

Unsure if it's a feature request or a bug, feel free to adjust the category.

**Expected Behavior**

As [documented](https://github.com/spring-projects/spring-authorization-server/blob/b87bf077ed13d9738c15a840cd9ebc04e3930f8f/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/settings/ClientSettings.java#L140) on the `requireAuthorizationConsent` setting, I would expect its value to be considered during a device code grant. For example, if the value is set to false, the consent page (that includes the user code and the scopes by default) should be skipped.

**Current Behavior**

In the case of the device code grant (`OAuth2DeviceVerificationAuthenticationProvider` class), the check only does a difference between the current scopes and the requested scopes. It means that the first time the flow is executed, the user has to accept the scopes (and confirm the user code). After that, if the scopes remain unchanged, the consent page is skipped.

[OAuth2DeviceVerificationAuthenticationProvider](https://github.com/spring-projects/spring-authorization-server/blob/b87bf077ed13d9738c15a840cd9ebc04e3930f8f/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2DeviceVerificationAuthenticationProvider.java#L134) (sadly not as customizable as the class below)
vs
[OAuth2AuthorizationCodeRequestAuthenticationProvider](https://github.com/spring-projects/spring-authorization-server/blob/b87bf077ed13d9738c15a840cd9ebc04e3930f8f/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2AuthorizationCodeRequestAuthenticationProvider.java#L207)

**Context**

I would like to be able to always skip the consent page. More precisely, the user code validation would be done in a separated page (like in your example). This page would then POST to the device verification endpoint (that would always skip the consent).

Thank you

jgrandja:
@antoinelauzon-bell I've confirmed that there is no check for `RegisteredClient.getClientSettings().isRequireAuthorizationConsent()` to allow for skipping the consent page. This would be a fairly quick update but are you looking for something more flexible like what was implemented in `OAuth2AuthorizationCodeRequestAuthenticationProvider` gh-1541 ?

antoinelauzon-bell:
Hi, thanks for the reply.

[...]

PR #2048, opened 2025-06-16
Add device verification authentication context support

dineshgupta630:
Device-Code Grant: pluggable consent check (Fixes #[1965](https://github.com/spring-projects/spring-authorization-server/issues/1965))

Adds a pluggable predicate to control whether consent is required in the device code grant flow.

Notes: No breaking changes; everything is opt-in.


jgrandja:
@dineshgupta630 I'm good with either option. I'll let you pick. 

> with option 1, do you reckon there's a chance the enhancement might not get the green light when it goes through Spring Security's process

It doesn't make a difference as I'm managing this PR. But we'll need to get this in before `7.0.0-RC1` Oct 20 so if you can update the code based on latest feedback in 2 weeks time then we should be good.

dineshgupta630:
Nice, I can definitely do that. Will do it this weekend. Thanks very much @jgrandja!

dineshgupta630:
Hi @jgrandja , thanks ever so much for your help with this - really appreciate it! I've just pushed through the changes, so when you get a moment, would you mind having a look and giving it a review? 

Do let me know if there's anything else you'd like me to tweak - more than happy to sort it out straightaway. Cheers.

jgrandja:
@dineshgupta630 Thanks for the updates. This is now merged via baa3b287d6e8b906db94ffa885b113a626edfdea

PR #2093, opened 2025-07-11
Make PAR request URI lifetime configurable

edouardhue:
- Replaces the hard-coded lifespan in `OAuth2PushedAuthorizationRequestUri` with a parameter.
- Adds a time-to-live attribute to `OAuth2PushedAuthorizationRequestAuthenticationProvider` to be passed to `OAuth2PushedAuthorizationRequestUri`.
- Defines a matching parameter in `OAuth2PushedAuthorizationRequestEndpointConfigurer`, with a default value of 5 minutes (same as the previous hard-coded value).
- Provides documentation for this new parameter.

The time-to-live must be a strictly positive duration.

The previous default behaviour is retained, no change is required to existing code.

Fixes gh-2042


jgrandja:
@edouardhue As noted in this [comment](https://github.com/spring-projects/spring-authorization-server/issues/2042#issuecomment-2978094937):

> I'll leave this issue open and we'll see if there is demand for this configuration option.

This enhancement is not scheduled for a release. If there is demand (based on upvotes) for this in the future, we will consider it for a future release.

As an FYI, before submitting a PR, please reach out to see if the PR would be accepted so you don't waste your efforts. Thanks.

I'll close this PR.

Issue #2205, opened 2025-09-26
Token Introspection fails if issuer claim is not a URI


MatthiasDrewsCS:
**Describe the bug**
The token introspection endpoint fails, if the passed token contains an issuer claim, which is not a URI. 

Looking at the specification in https://datatracker.ietf.org/doc/html/rfc7519#section-4.1.1, it says that it must be an URI **or** String. In our specific case it is not a URI, because we set the issuer to String as we are following the client assertion specification (https://datatracker.ietf.org/doc/html/rfc7521), so the issuer claim is the clientId.

The stacktrace is:
```
java.lang.IllegalArgumentException: iss must be a valid URL
	at org.springframework.security.oauth2.server.authorization.OAuth2TokenIntrospection$Builder.validateURL(OAuth2TokenIntrospection.java:337)
	at org.springframework.security.oauth2.server.authorization.OAuth2TokenIntrospection$Builder.validate(OAuth2TokenIntrospection.java:307)
	at org.springframework.security.oauth2.server.authorization.OAuth2TokenIntrospection$Builder.build(OAuth2TokenIntrospection.java:278)
	at org.springframework.security.oauth2.server.authorization.authentication.OAuth2TokenIntrospectionAuthenticationProvider.withActiveTokenClaims(OAuth2TokenIntrospectionAuthenticationProvider.java:158)
	at org.springframework.security.oauth2.server.authorization.authentication.OAuth2TokenIntrospectionAuthenticationProvider.authenticate(OAuth2TokenIntrospectionAuthenticationProvider.java:114)
	at org.springframework.security.authentication.ProviderManager.authenticate(ProviderManager.java:182)
...
```

**Expected behavior**
Using the introspection token endpoint should be possible if the issuer claim is not a URI. I don't see why the claims are validated in the introspection endpoint in such a way, that tokens which have been created just before are invalid

jgrandja:
@MatthiasDrewsCS 

[RFC 7519 JSON Web Token (JWT)](https://datatracker.ietf.org/doc/html/rfc7519) is **a generalized specification** that other specifications **MAY** comply to in full or part.

The `iss` claim in the [OAuth 2.0 Token Introspection response ](https://datatracker.ietf.org/doc/html/rfc7662#section-2.2) states:

> String representing the issuer of this token, as defined in JWT [[RFC7519](https://datatracker.ietf.org/doc/html/rfc7519)]

[...]

PR #2206, opened 2025-09-28
Fix OAuth2TokenIntrospection issuer validation and add unit tests

ramesh-nair-dev:
### Summary
This PR updates the OAuth2TokenIntrospection class to correctly handle the 'issuer' (iss) claim. The changes ensure that the issuer can be either a valid URI or a plain string without throwing exceptions during token introspection validation.

### Changes
- Updated `OAuth2TokenIntrospection.Builder` to accept non-URI issuer strings.
- Enhanced `validateIssuer` method to allow plain string issuers while still supporting valid URIs.
- Added unit tests to cover:
  - Non-URI issuer strings
  - Valid URI issuer strings
- Ensured existing claim validations (active, scope, exp, iat, nbf, aud) remain intact.

### Benefits
- Prevents unnecessary exceptions when using non-URI issuer strings.
- Improves robustness and flexibility of OAuth2 token introspection.
- Increases test coverage for issuer-related functionality.

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2TokenIntrospection.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2TokenIntrospection.java
@@ -305,7 +305,7 @@ public final class OAuth2TokenIntrospection implements OAuth2TokenIntrospectionC
 						"aud must be of type List");
 			}
 			if (this.claims.containsKey(OAuth2TokenIntrospectionClaimNames.ISS)) {
-				validateIssuer(this.claims.get(OAuth2TokenIntrospectionClaimNames.ISS), "iss must be a valid URL");
+				validateIssuer(this.claims.get(OAuth2TokenIntrospectionClaimNames.ISS), "iss must not be empty");
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/OAuth2TokenIntrospection.java
```diff
@@ -326,16 +326,24 @@ private void acceptClaimValues(String name, Consumer<List<String>> valuesConsume
...
 
+			String str = url.toString();
+			if (str.isEmpty()) {
+				throw new IllegalArgumentException(errorMessage);
+			}
+
 			try {
-				new URI(url.toString()).toURL();
+				// Try parsing as URI
+				new URI(str);
```
MatthiasDrewsCS, line 341: Is there really a benefit from first trying to parse a URI, and if this fails, still continue successfully? IMHO it would be enough to just check if the string is blank or not. Am I missing something?

ramesh-nair-dev: Thanks for the feedback You’re right — if we allow non-URI strings (like client IDs), then attempting to parse as a URI may not add much value.

My initial thought behind keeping the URI parsing was to still preserve compatibility with tokens where iss is a proper URI, and not reject them unintentionally. But since the validation already ensures the string isn’t blank, simplifying to just a non-blank check would be sufficient and cleaner.

I can update the implementation to only validate that iss is not blank. Would you prefer that direction?

ramesh-nair-dev: I will make the changes

ramesh-nair-dev:
Hey matt can you please review changes

MatthiasDrewsCS:
I like that. It would be even better to also slightly adjust the error message to something like "iss must not be empty"

ramesh-nair-dev:
Thanks @MatthiasDrewsCS for the review and approval 
Could a maintainer please approve the workflows and merge if everything looks good?

jgrandja:
@ramesh-nair-dev See [comment](https://github.com/spring-projects/spring-authorization-server/issues/2205#issuecomment-3380920742)

PR #2320, opened 2026-04-07
Rotated refresh token uses the expiry of initial refresh token

kpur-sbab:
In Spring Authorization Server, when refresh token rotation is enabled, each newly issued refresh token receives a full TTL. As a result, if a refresh token is rotated close to its expiry, the new token effectively resets its lifetime. This behavior appears to violate the https://github.com/spring-projects/spring-authorization-server/issues/297 stated guideline and could introduce a potential security concern.

The fix is to make the changes in OAuth2RefreshTokenGenerator to enable this support by reusing the expiry from the context's authorization model


jgrandja:
Closing as per [comment](https://github.com/spring-projects/spring-authorization-server/issues/2318#issuecomment-4243956372)

Issue #2195, opened 2025-09-12
Move Spring Authorization Server to Spring Security


jgrandja:
See [spring-security#17880](https://github.com/spring-projects/spring-security/issues/17880)

PR #2337, opened 2026-05-27
Add support for BINARY (BYTEA) column type in JdbcOAuth2AuthorizationService

s-chan-o:
## Problem

`JdbcOAuth2AuthorizationService` fails when PostgreSQL (or PostgreSQL-compatible databases such as AWS Aurora DSQL) is used as the backend store.

PostgreSQL represents binary data using the `BYTEA` type, which is reported by the JDBC driver as `Types.BINARY` (`-2`). The current implementation only handles `Types.BLOB` and `Types.CLOB`, so BYTEA columns cause data read/write failures.

Fixes gh-2321

## Changes

Added `Types.BINARY` handling in three locations within `JdbcOAuth2AuthorizationService`:

| Method | Change |
|--------|--------|
| `getLobValue` | Reads bytes and converts to `String`; trims trailing null bytes present in fixed-length `BINARY` columns |
| `doSetValue` | Writes `byte[]` or `String` values via `LobCreator` for `Types.BINARY` parameters |
| `mapToSqlParameter` | Converts `String` values to `byte[]` when the column type is `Types.BINARY` |

## Testing

- Added `custom-oauth2-authorization-schema-binary-data-type.sql` with `binary(10000)` column definitions (HSQL equivalent of PostgreSQL BYTEA, mapping to `Types.BINARY`)
- Added `tableDefinitionWhenBinarySqlTypeThenAuthorizationUpdated()` test following the same pattern as the existing CLOB test


jgrandja:
@s-chan-o 

PR #2345, opened 2026-06-12
Compile with Java 25 and Gradle 9.5

bostko:
I want to run project without downloading new JDK and aim at more present version of gradle.


jgrandja:
@bostko 

PR #2347, opened 2026-06-18
Parse pushed request URI from last delimiter

### LINKED ISSUE #2346: OAuth2PushedAuthorizationRequestUri generates URIs which cannot be parsed due to ___ in the state part
**Describe the bug**
spring-security-oauth2-authorization-server 1.5.2 and 1.5.7 

A percentage of PAR (Pushed Authorization Request) requests fail randomly. 
[OAuth2PushedAuthorizationRequestUri](https://github.com/spring-projects/spring-authorization-server/blob/1.5.8/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2PushedAuthorizationRequestUri.java) generates URIs in this format:  
```
urn:ietf:params:oauth:request_uri:<Base64URL_state>___<epoch_millis>  
Generator: Base64StringKeyGenerator(Base64.getUrlEncoder())  
- Base64URL alphabet: A-Z, a-z, 0-9, -, _ (64 characters)  
- Character '_' (underscore) IS in the output!  
- Delimiter '___' (triple underscore) also uses '_'  
=> CONFLICT!  
``` 
the parse method contains the bug:

```
urn:ietf:params:oauth:request_uri:xXMGJTZwzXIFL8i_DFu_EM8IeWC___frCWjpiF2q-xs=___1781670640281  
^^^ ^^^  
indexOf() matches triple underscore delimiter  
Result:  
- substring = "frCWjpiF2q-xs=___1781670640281"  
- Long.parseLong(...) → NumberFormatException!  
- catch(Exception) → throwError("invalid_request", "request_uri")  
- User receives error, authentication FAILS  
``` 

**To Reproduce**
Failure probability is 1/6,554, the sample project demonstrates this with a unit test.

**Expected behavior**
All created URIs can be parsed.

Either the state generation is changed or the parse method accepts URIs with ___ in the state part.

A suggested fix is to change the parse method to replace `indexOf` with `lastIndexOf` when looking for the delimiter position. 

**Sample**

https://github.com/christopherm-backbase/spring-auth-server-par-request-uri-repro

goutamadwant:
Fixes gh-2346

This updates pushed authorization request URI parsing so the expiration timestamp is read after the last delimiter. The generated state uses Base64 URL characters, so it can contain underscores and may include the same delimiter sequence used before the timestamp.

Tests:
- `./gradlew :spring-security-oauth2-authorization-server:test --tests OAuth2PushedAuthorizationRequestUriTests --tests OAuth2AuthorizationCodeRequestAuthenticationProviderTests --tests OAuth2PushedAuthorizationRequestAuthenticationProviderTests`

### DIFF
--- a/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2PushedAuthorizationRequestUri.java
+++ b/oauth2-authorization-server/src/main/java/org/springframework/security/oauth2/server/authorization/authentication/OAuth2PushedAuthorizationRequestUri.java
@@ -59,7 +59,7 @@ final class OAuth2PushedAuthorizationRequestUri {
 
 	static OAuth2PushedAuthorizationRequestUri parse(String requestUri) {
 		int stateStartIndex = REQUEST_URI_PREFIX.length();
-		int expiresAtStartIndex = requestUri.indexOf(REQUEST_URI_DELIMITER) + REQUEST_URI_DELIMITER.length();
+		int expiresAtStartIndex = requestUri.lastIndexOf(REQUEST_URI_DELIMITER) + REQUEST_URI_DELIMITER.length();
 		OAuth2PushedAuthorizationRequestUri pushedAuthorizationRequestUri = new OAuth2PushedAuthorizationRequestUri();
 		pushedAuthorizationRequestUri.requestUri = requestUri;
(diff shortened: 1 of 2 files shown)


christopherm-backbase (approved):

jgrandja:
Closing as [per comment](https://github.com/spring-projects/spring-authorization-server/issues/2346#issuecomment-5332490275)

PR #2352, opened 2026-06-26
GH-2333 Document clustered OIDC sessions

### LINKED ISSUE #2333: OidcLogoutAuthenticationProvider fails with 400 on sid validation in a clustered Spring Authorization Server setup
### Environment
- Spring Boot: 4.0.2
- Spring Security OAuth2 Authorization Server: 7.0.2
- Spring Session with Redis (namespace: `auth:server`)
- JdbcOAuth2AuthorizationService (PostgreSQL)
- 2 auth server instances behind a load balancer (round-robin)

---

### Describe the bug

When running two instances of Spring Authorization Server behind a round-robin load balancer, the OIDC RP-Initiated Logout flow fails intermittently with HTTP 400. The error occurs inside `OidcLogoutAuthenticationProvider` when validating the `sid` parameter from the `id_token_hint`.

The auth server uses:
- **Redis** for HTTP session storage (shared between instances via Spring Session)
- **JDBC** for OAuth2 authorization storage (shared between instances via `JdbcOAuth2AuthorizationService`)

Despite both stores being shared, the `sid` validation fails when the logout request is routed to a different instance than the one that originally created the login session.

---

### Observed behavior

```
Authentication failed with provider OidcLogoutAuthenticationProvider
since OpenID Connect 1.0 Logout Request Parameter: sid

org.springframework.security.oauth2.core.OAuth2AuthenticationException
```

The request is then forwarded to `/error` with the full `id_token_hint`, `post_logout_redirect_uri`, `client_id`, and `state` parameters, returning HTTP 400.

---

### Expected behavior

Since both Redis (session store) and JDBC (authorization store) are shared across all instances, `OidcLogoutAuthenticationProvider` should be able to validate the `sid` claim regardless of which instance handles the logout request.

---

### Root cause analysis

The `sid` claim in the `id_token_hint` represents the auth server's HTTP session ID for the user's login session. `OidcLogoutAuthenticationProvider` attempts to correlate this `sid` against an active session. However, in a clustered setup with round-robin load balancing:

1. **Instance A** handles the login — creates the session, stores `sid` in the ID token
2. **Instance B** handles the logout — cannot correlate the `sid` even though the session data exists in the shared Redis store

This suggests that the `sid`-to-session correlation mechanism inside `OidcLogoutAuthenticationProvider` relies on **in-memory state** rather than the shared session store, making it fundamentally incompatible with a clustered deployment even when Redis-backed Spring Session is configured.

There is no `OidcSessionRegistry` or equivalent interface in the authorization server jar (`spring-security-oauth2-authorization-server:7.0.2`) that would allow replacing this in-memory correlation with a shared store implementation.

---

### Steps to reproduce

1. Configure Spring Authorization Server with Redis-backed Spring Session
2. Configure `JdbcOAuth2AuthorizationService` for shared token storage
3. Run two instances behind a round-robin load balancer
4. Log in via the authorization code flow
5. Trigger RP-Initiated Logout with a valid `id_token_hint` containing a `sid` claim
6. Observe intermittent 400 errors when the logout request is handled by a different instance than the login

---

goutamadwant:
## Summary

- Document that clustered OpenID Connect logout deployments need a shared `SessionRegistry`.
- Show how to register `SpringSessionBackedSessionRegistry` with a Spring Session indexed repository.
- Clarify in the Redis guide that Redis-backed authorization server data is separate from Redis-backed HTTP sessions.

Fixes gh-2333

## Testing

- `git diff --check`
- `./gradlew :spring-authorization-server-docs:antora --console=plain`

### DIFF
--- a/docs/modules/ROOT/pages/core-model-components.adoc
+++ b/docs/modules/ROOT/pages/core-model-components.adoc
@@ -532,3 +532,18 @@ public HttpSessionEventPublisher httpSessionEventPublisher() {
 	return new HttpSessionEventPublisher();
 }
 ----
+
+In a clustered environment, the `SessionRegistry` must also be shared across application instances.
+For example, when using https://docs.spring.io/spring-session/reference/index.html[Spring Session] to back HTTP sessions, register a `SpringSessionBackedSessionRegistry` instead of `SessionRegistryImpl`:
+
+[source,java]
+----
+@Bean
+public SessionRegistry sessionRegistry(
+		FindByIndexNameSessionRepository<? extends Session> sessionRepository) {
+	return new SpringSessionBackedSessionRegistry<>(sessionRepository);
+}
+----
+
+When backing Spring Session with Redis, make sure an indexed session repository is configured, for example with `@EnableRedisIndexedHttpSession`, so that a `FindByIndexNameSessionRepository` is available.
+This allows the OpenID Connect 1.0 Logout Endpoint to validate the `sid` claim even when the logout request is handled by a different application instance than the original login request.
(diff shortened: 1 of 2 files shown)


jgrandja:
@goutamadwant 

